PR #7, opened 2019-04-03
Fix uselocale to support non-default locales.

sunfishcode:
Musl's locales implementation depends on pthreads. Since WASI libc
doesn't yet have pthreads, it currently has alternate code for
supporting locales. Previously, this code just assumed that it only had
to support the default locale, however libc++ uses uselocale with
non-default locals, so add support for that.

With this, the C++ <iostream>-style hello world now works, with
-fno-exceptions.


sunfishcode:
Taking the liberty of merging this, to unblock C++. This also cleans up an earlier hack, so even if we should make further changes here, this is a better place to start from than before.

PR #8, opened 2019-04-03
Filter out __FLT16_* macros from predefined-macros.txt.

sunfishcode:
Whether these are defined or not depends on the version of clang in use,
and they're not that important at this time, so for now just ignore them
for the purposes of this testing output.

This fixes the issue reported in https://github.com/CraneStation/wasi-sdk/issues/6.


sunfishcode:
This fixes CraneStation/wasi-sdk#6 and is a fairly simple patch.

PR #9, opened 2019-04-03
Add comments explaining changes to upstream source files.

sunfishcode:
Add at least a one-line comment on every
`#ifdef __wasilibc_unmodified_upstream` briefly indicating what its
purpose is.

While here, reenable some of the code that was previously disabled when
getenv wasn't supported, as getenv is now supported.


sunfishcode:
Merging this to keep this branch in sync with https://github.com/WebAssembly/reference-sysroot/pull/11.

PR #10, opened 2019-04-03
Provide a public interface to preopened directory lookups.

sunfishcode:
For users of especially non-C compilers, provide an API for looking up
preopened directories. This is used internally in WASI libc to translate
from eg. `open` to `openat`, but it can now also be used by user code.

### DIFF
--- a/expected/wasm32-wasi/defined-symbols.txt
+++ b/expected/wasm32-wasi/defined-symbols.txt
@@ -242,6 +242,7 @@ __unlist_locked_file
 __uselocale
 __utc
 __wasilibc_fd_renumber
+__wasilibc_find_relpath
 __wasilibc_init_preopen
 __wasilibc_register_preopened_fd
 __wasilibc_rmdirat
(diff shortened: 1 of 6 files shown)


jedisct1:
Oh, yes, please!

alexcrichton:
Thanks @sunfishcode! For posterity this solves a relatively core problem that came up in the development of https://github.com/rust-lang/rust/pull/59619 which was how to best implement absolute-path-taking-APIs outside of libc itself (aka in the Rust standard library), and an interface like this ended up being perfect for our use case!

The only comment I'd have is I wonder if it'd be best to return an `int` error code perhaps? It seems a bit wonky to reach inside the struct to see if it returned an invalid fd as opposed to treating it like most other libc functions which return an error code.

sunfishcode:
@alexcrichton Ok, I adjusted the function a little to return the file descriptor, or -1 on failure like a typical C function. It provides the relative path via an out parameter. Does that make sense?

alexcrichton:
Looks great to me!

alexcrichton:
I'm integrating this into Rust with https://github.com/rust-lang/rust/pull/59727 and some quick local testing shows it working perfectly!

PR #15, opened 2019-04-15
Update to cloudlibc 8835639f27fc42d32096d59d294a0bbb857dc368.

sunfishcode:
There's no functional change here; the only change is the patch I
submitted upstream, so we can remove some local changes.


sbc100:
Do you have scripts for automating these kinds of updates?  if so they sounds very useful.

sunfishcode:
No, right now it's a manual process.

PR #21, opened 2019-04-20
Fix floatscan no long double usage

michaelvoronov:
In the current version of `libc-printscan-no-long-double` long double could still appear by using functions with suffix `l` that receive long double parameters. So in the result binary, functions like __extenddftf2 will be generated.

### DIFF
--- a/libc-top-half/headers/private/printscan.h
+++ b/libc-top-half/headers/private/printscan.h
@@ -37,6 +37,12 @@ typedef double long_double;
 #define LDBL_MAX_10_EXP DBL_MAX_10_EXP
 #undef frexpl
 #define frexpl(x, exp) frexp(x, exp)
+#undef copysignl
+#define copysignl(x, y) copysign(x, y)
+#undef fmodl
+#define fmodl(x, y) fmod(x, y)
+#undef scalbnl
+#define scalbnl(arg, exp) scalbn(arg, exp)
 __attribute__((__cold__, __noreturn__))
 static void long_double_not_supported(void) {
     void abort(void) __attribute__((__noreturn__));
--- a/libc-top-half/musl/src/internal/floatscan.c
+++ b/libc-top-half/musl/src/internal/floatscan.c
@@ -301,8 +301,13 @@ static long double decfloat(FILE *f, int c, int bits, int emin, int sign, int po
 
 	/* Calculate bias term to force rounding, move out lower bits */
 	if (bits < LDBL_MANT_DIG) {
-		bias = copysignl(scalbn(1, 2*LDBL_MANT_DIG-bits-1), y);
-		frac = fmodl(y, scalbn(1, LDBL_MANT_DIG-bits));
+#if defined(__wasilibc_printscan_no_long_double)
+		bias = copysign(scalbn(1, 2*LDBL_MANT_DIG-bits-1), y);
+		frac = fmod(y, scalbn(1, LDBL_MANT_DIG-bits));
+#else
+        bias = copysignl(scalbn(1, 2*LDBL_MANT_DIG-bits-1), y);
+        frac = fmodl(y, scalbn(1, LDBL_MANT_DIG-bits));
+#endif
 		y -= frac;
 		y += bias;
 	}
@@ -320,7 +325,11 @@ static long double decfloat(FILE *f, int c, int bits, int emin, int sign, int po
 			else
 				frac += 0.75*sign;
 		}
-		if (LDBL_MANT_DIG-bits >= 2 && !fmodl(frac, 1))
+#if defined(__wasilibc_printscan_no_long_double)
+		if (LDBL_MANT_DIG-bits >= 2 && !fmod(frac, 1))
+#else
+        if (LDBL_MANT_DIG-bits >= 2 && !fmodl(frac, 1))
+#endif
 			frac++;
 	}
 
@@ -338,7 +347,11 @@ static long double decfloat(FILE *f, int c, int bits, int emin, int sign, int po
 			errno = ERANGE;
 	}
 
-	return scalbnl(y, e2);
+#if defined(__wasilibc_printscan_no_long_double)
+	return scalbn(y, e2);
+#else
+    return scalbnl(y, e2);
+#endif
 }
 
 #if defined(__wasilibc_printscan_no_long_double)
@@ -451,7 +464,11 @@ static long double hexfloat(FILE *f, int bits, int emin, int sign, int pok)
 	}
 
 	if (bits < LDBL_MANT_DIG)
-		bias = copysignl(scalbn(1, 32+LDBL_MANT_DIG-bits-1), sign);
+#if defined(__wasilibc_printscan_no_long_double)
+		bias = copysign(scalbn(1, 32+LDBL_MANT_DIG-bits-1), sign);
+#else
+        bias = copysignl(scalbn(1, 32+LDBL_MANT_DIG-bits-1), sign);
+#endif
 
 	if (bits<32 && y && !(x&1)) x++, y=0;
 
@@ -464,7 +481,11 @@ static long double hexfloat(FILE *f, int bits, int emin, int sign, int pok)
 
 	if (!y) errno = ERANGE;
 
-	return scalbnl(y, e2);
+#if defined(__wasilibc_printscan_no_long_double)
+	return scalbn(y, e2);
+#else
+    return scalbnl(y, e2);
+#endif
 }
 
 #if defined(__wasilibc_printscan_no_long_double)
--- a/libc-top-half/musl/src/internal/floatscan.c
+++ b/libc-top-half/musl/src/internal/floatscan.c
@@ -290,7 +290,7 @@ static long double decfloat(FILE *f, int c, int bits, int emin, int sign, int po
 #if defined(__wasilibc_printscan_no_long_double)
 		y = 1000000000.0 * y + x[a+i & MASK];
 #else
-        y = 1000000000.0L * y + x[a+i & MASK];
+		y = 1000000000.0L * y + x[a+i & MASK];
 #endif
 	}
 
--- a/libc-top-half/musl/src/internal/floatscan.c
+++ b/libc-top-half/musl/src/internal/floatscan.c
@@ -305,8 +305,8 @@ static long double decfloat(FILE *f, int c, int bits, int emin, int sign, int po
 		bias = copysign(scalbn(1, 2*LDBL_MANT_DIG-bits-1), y);
 		frac = fmod(y, scalbn(1, LDBL_MANT_DIG-bits));
 #else
-        bias = copysignl(scalbn(1, 2*LDBL_MANT_DIG-bits-1), y);
-        frac = fmodl(y, scalbn(1, LDBL_MANT_DIG-bits));
+		bias = copysignl(scalbn(1, 2*LDBL_MANT_DIG-bits-1), y);
+		frac = fmodl(y, scalbn(1, LDBL_MANT_DIG-bits));
 #endif
 		y -= frac;
 		y += bias;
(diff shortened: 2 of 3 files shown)


sunfishcode:
Thanks for noticing this!

For copysignl, fmodl, and scalbnl, I'd prefer to add macros to printscan.h to reroute them to double versions, as is done for frexpl:
https://github.com/CraneStation/wasi-sysroot/blob/master/libc-top-half/headers/private/printscan.h#L38
so that we don't need as many local modifications in the floating-point formatting code.

One additional place that would be good to fix is this constant with an `L` suffix:
https://github.com/CraneStation/wasi-sysroot/blob/master/libc-top-half/musl/src/internal/floatscan.c#L290
This one will require a local modification.

As background, these printf and scanf optimizations for long double are a work in progress. While working on them, we noticed an alternative approach, which is to have the compiler automatically select the version of printf needed. The compiler side of this is now implemented:
https://reviews.llvm.org/D57620
although more work remains on the WASI libc side before we can enable this for WASI. If you're interested, I'd be happy to mentor someone on this.

michaelvoronov:
Thank you for comment, It seems a better idea. Fixed it.

For background - sounds interesting, I would gladly help if it is possible:).

And also I have one more suggestion. It concerns macroses `isnan`, `isinf`, `isnormal`, and `isfinite`. In musl it has the next [definitions](https://github.com/CraneStation/wasi-sysroot/blob/320054e84f8f2440def3b1c8700cedb8fd697bf8/libc-top-half/musl/include/math.h#L61). After compiling to wasm these macroses produce 100+ wasm instructions (the example of `isnan` [here](https://gist.github.com/michaelvoronov/5d7a396b1a2e7eb5e2c517555875b771)) with imports of `__extenddftf2` and `__fpclassifyl` from compiler-rt. The thinks gets worse if we consider that it is a macro not a function - every call of these harmless functions will produce 100+ wasm instructions. What do you think of introduction a new macro for that one or maybe it is better to make it on the llvm side?

sunfishcode:
> Thank you for your comment. It seems a better idea. Fixed it.

Thanks!

> For background - sounds interesting, I would gladly help if it is possible:).

Cool. It's a little complex, but I think we can do it in steps.

So in that LLVM patch I linked to, if we enable that optimization for WASI, it'll rewrite `printf` calls to `iprint` if there's no floating point, or `__small_printf` if there's no long double. So as a first step we can make libc provide `iprintf` an alias for `printf`, and `__small_printf` as a version of printf compiled with long double disabled.

That way, a program with no long double will automatically get a printf with no long double support, which is the main thing we want. However, a program with long double support may now get two versions of printf linked in. And, a program with no floating point will still get a printf with full floating point support in it. However, I think it's fine to do these in separate steps.

Does that sound doable? Feel free to ask questions if anything is unclear.

> And also I have one more suggestion. It concerns macroses isnan, isinf, isnormal, and isfinite. In musl it has the next definitions. After compiling to wasm these macroses produce 100+ wasm instructions (the example of isnan here) with imports of __extenddftf2 and __fpclassifyl from compiler-rt. The things get worse if we consider that it is a macro, not a function - every call of these harmless functions will produce 100+ wasm instructions. What do you think about introducing a new macro for that one or maybe it is better to make it on the llvm side?

I'm curious what compiler you're using here; when I compile a simple `isnan` test with clang, even at -O0, I don't see the `__extenddftf2` or `__fpclassifyl` calls.

Enabling optimizations helps a lot, however things are still a little inconsistent. For example at -O1 the `__DOUBLEBITS` function doesn't get inlined. So I'd be in favor of moving `isnan` and friends, including `signbit` and `fpclassify`, over to the compiler-provided builtins -- `__builtin_isnan` and similar, so that we always get the fully optimized form. Would you be interested in doing that?

michaelvoronov:
> Does that sound doable? Feel free to ask questions if anything is unclear.
Thank you for sharing some details, i will investigate it tomorrow.

> I'm curious what compiler you're using here; when I compile a simple isnan test with clang, even at -O0, I don't see the __extenddftf2 or __fpclassifyl calls.

Hmm, maybe it could depend on the context. I am getting this while compiling [Redis](https://github.com/fluencelabs/redis/blob/wasm/src/t_hash.c#L631) without -O option. Anyway, i will investigate it in a couple days and will create a PR.

P.S. Yes, It is possible to compile Redis 5.0.3 with WASI to Webassembly. Great work, guys!

sunfishcode:
Have you had a chance to investigate this?

michaelvoronov:
Hi, yes, i now have enough time to work on it and have some questions: Is `print_no_Lf` from [here](https://github.com/CraneStation/wasi-sysroot/blob/e5f14be38362f1ab83302895a6e74b2ffd0e2302/libc-top-half/musl/src/stdio/printf.c#L16) is just the old name for `__small_printf`? If so, can I make a micro-PR first that renames it?

P.S. Also, maybe you know at once what could be wrong in this [issue](https://github.com/fluencelabs/fluence/issues/649)?

sunfishcode:
Ah, yes, it's the same. `printf_no_Lf` was an earlier name for it, and we changed our minds and went with `__small_printf` in the LLVM patch, but wasi-sysroot isn't updated yet. So yeah, a first PR that just renames that would be a good first step.

sunfishcode:
Also, #649 is quite likely https://bugs.llvm.org/show_bug.cgi?id=41508 which is now [fixed upstream](https://github.com/llvm/llvm-project/commit/5081e41bdae2eb14a3f3eb8810263f9fea8fc7c1). It's not in the 8.0 release, but I've [requested it be included in 8.0.1](https://bugs.llvm.org/show_bug.cgi?id=41602).

michaelvoronov:
Thank you!

Also, I have investigated `isnan` macro and its friends on example of Redis with different optimizations flags. So for `hincrbyCommand` (https://github.com/fluencelabs/redis/blob/wasm/src/t_hash.c#L564) function:
-00 - https://gist.github.com/michaelvoronov/b73648e1886f1f03a3d02f1aaf46d8e1, isnan and isinf begin on row 225.
- O1 - https://gist.github.com/michaelvoronov/8df1b1af94735705a18086ac8bb7623a, isnan and isinf begin on row 72 and look much more optimized.
-O2 - https://gist.github.com/michaelvoronov/f8a6bd8d148acacd63ec9baa0d5a1315, isnan and isinf begin on row 103 and look like the previous one.

https://gist.github.com/michaelvoronov/5d7a396b1a2e7eb5e2c517555875b771 - this one that I was mentioned earlier I got in compilation by two steps (clang -e + wasm-ld) since this deadlock. So generally clang optimize it well on the O2 level.

But, as you already sad there is one little inconsistent moment, let's view on 
https://gist.github.com/michaelvoronov/942c74e494408ff8845f4a5a0c01f14d. It is a subset of resulted Redis wast file compiled on 01 level. And there are seven the same `__DOUBLE_BITS_` functions - it seems that one for every compilation unit (source code file). Another interesting moment is that clang optimize with inlining only several isnan-similar functions call. Like the previous one `hincrbyCommand`, but there is no such optimization for `zsetAdd`, `incrbyfloatCommand` (they are also in this listing), and many other.

But since there are no such __DOUBLE_BITS_ functions O2 optimization at all, do you think it still possible to make a PR with builtins instead of isnan and its friends? If so, I will make it today (already tested it).

sunfishcode:
Yes, a PR to wasi-sysroot that changes isnan and friends to __builtin_isnan and friends would be very welcome. The fact that you found a place at -O1 that isn't optimized means there's a decent chance there are cases where it wouldn't be optimized at -O2 too. Also, from a compiler-writer perspective, union-based type punning is murky, particularly in a function meant to be #included and inlined in many different contexts, and it's much nicer when you can just say what you want to do directly.

Note that the convention in wasi-sysroot for making changes like this is to use a preprocessor conditional like this:

```
#ifdef __wasilibc_unmodified_upstream // comment explaining the change
// original code here
#else
// new code here
#endif
```

michaelvoronov:
Also, I have another little problem while porting Redis - using functions like `qsort`, `srand` without `#include <stdlib.h>`. In this case clang generates warning like this one 
```
2 warnings generated.
./src/geo.c:586:9: warning: implicit declaration of function 'qsort' is invalid in C99 [-Wimplicit-function-declaration]
        qsort(ga->array, result_length, sizeof(geoPoint), sort_gp_asc);
        ^
./src/geo.c:588:9: warning: implicit declaration of function 'qsort' is invalid in C99 [-Wimplicit-function-declaration]
        qsort(ga->array, result_length, sizeof(geoPoint), sort_gp_desc);
```
and linker also generates warning like this one:
```
wasm-ld: warning: function signature mismatch: qsort
>>> defined as (i32, i32, i32, i32) -> i32 in geo.o
>>> defined as (i32, i32, i32, i32) -> void in /opt/wasi-sdk/share/sysroot/lib/wasm32-wasi/libc.a(qsort.o)

wasm-ld: warning: function signature mismatch: srand
>>> defined as (i32) -> i32 in server.o
>>> defined as (i32) -> void in /opt/wasi-sdk/share/sysroot/lib/wasm32-wasi/libc.a(rand.o)
```

In case of x86/amd64 on all calling conventions this mistake is harmless, but in wasm32-* targets it lead compiler to generate additional `drop` instruction after each call of such functions. E.g., `wasm2wat` from `wabt` ends with such error:
```
redis.wasm:00289ab: error: type mismatch in drop, expected [any] but got []
redis.wasm:00289cd: error: type mismatch in drop, expected [any] but got []
```

What do you think, is it good idea to force somehow (maybe on clang/llvm side?) this warning to error, since the result binary will be broken in such scheme (all validator will consider it as incorrect)?

(sorry, maybe it isn't a good idea to write about this case in this thread).

sunfishcode:
(I don't mind answering here, but also feel free to file new issues for questions like this.)

Yeah, it seems like this is something wasm-ld may want to issue errors about. I've now filed https://bugs.llvm.org/show_bug.cgi?id=41717 to propose the idea.

michaelvoronov:
Cool, thank you!

So, about floats I have some unclear moments:
1. As I understand correctly, the final goal on the last step is to have only one compiled to-half-libc with these three printf versions? So, there will one cons - a little size overhead for programs that used different printf simultaneously.
2. If so, It seems that it is possible to implement `iprintf` as a separate .c file that includes `printf.c` and call to the original `printf` with `__wasilibc_printscan_no_floating_point` defined and `__small_printf` as  `printf.c` with `__wasilibc_printscan_no_long_double` defined. It will avoid code duplication.
3. Maybe it is also good to make the same for `scanf` family functions? E.g., at now scanf performs `%g` processing that implies linking with heavy floatscan. It could be especially useful for some projects implied deterministic computation since many of them refuse any floating point operations since its non-determinism.

sunfishcode:
Re: 1: That sounds right, with one modification. We can reduce the size overhead by using weak symbols:
 - In the version that defines `iprintf`, make `iprintf` weak  -- with `__attribute__((weak))`.
 - In the version that defines `__small_printf`, make `__small_printf` weak, and define a strong alias named `iprint` to it -- with `int iprintf(const char *, ...) __attribute__((alias("__small_printf")));`.

This should eliminate the size overhead in the case where you have floating-point but no long double. And it means that if you do use long double, you only get two printf implementations, rather than one.

(Ideally, we'd eliminate the duplication even in the long double case too, however the second `iprintf` needs to be strong to override the first one, and then since it's strong a third one can't override it, so we can't use the same trick twice. I think there are ways to solve this, but as far as I can tell they're much more involved.)

Re: 2: Yes, avoiding code duplication is good. That can be accomplished either by having multiple files which #include the same implementation, or having Makefile rules which compile the same file with different preprocessor defines.

Re: 3: Yes, all of the above can be done for scanf as well as printf. It is a lower priority, because scanf is much less frequently used than printf, however the same considerations apply otherwise.

PR #22, opened 2019-04-21
Fixes #20

### LINKED ISSUE #20: localtime is defined in the header, but has no implementation
the `localtime` function is [present in wasi-sysroot's headers](https://github.com/CraneStation/wasi-sysroot/blob/e5f14be38362f1ab83302895a6e74b2ffd0e2302/libc-top-half/musl/include/time.h#L67), but not in the actual libc library, so programs calling localtime fails during linking:

```
#include <time.h>

int main() {
	time_t a;
	localtime(&a);
}
```

```
zhuowei@zhuowei-laptop:~/wasi-sdk/bin$ ./clang-8 localtime.c 
wasm-ld: error: /tmp/localtime-230b9c.o: undefined symbol: localtime
clang-8: error: lld command failed with exit code 1 (use -v to see invocation)
```

Either the localtime function should be removed from the header, or an implementation should be added to the libc.

kubkon:
Adds missing localtime.c source file to Makefile and to expected
defined symbols list.


sunfishcode:
Looks good, thanks!

PR #27, opened 2019-04-23
Add missing asctime symbol

kubkon:
Let me know if you'd prefer one larger PR with more missing symbols included, or whether small PRs like this one or #22 are OK.


sunfishcode:
Small PRs like this are fine.

PR #29, opened 2019-04-23
Implement reallocarray.

sunfishcode:
reallocarray is non-standard but present in GLIBC and BSDs.


jedisct1:
Yes! This is a very useful function to avoid arithmetic overflows, and I totally support its inclusion.

PR #30, opened 2019-04-23
Make calloc set ENOMEM when failing due to overflow.

sunfishcode:
This fixes a bug in upstream dlmalloc, where it doesn't set errno to
ENOMEM in overflow or footprint overrun cases.


sunfishcode:
It appears to be a minor oversight in the upstream code. I've sent a report to the upstream author.

PR #39, opened 2019-05-01
Rename printf_no_Lf to __small_printf

michaelvoronov:
Step 1 of better no long double optimizations:
- rename `printf_no_Lf` to `__small_printf` according to https://reviews.llvm.org/D57620.


sunfishcode:
Looks good!

PR #40, opened 2019-05-01
Force clang to optimize isnan and friends

michaelvoronov:
At now `isnan` and friends from `math.h` could be non-optimized by the compiler. E.g., I am getting the following [result](https://gist.github.com/michaelvoronov/942c74e494408ff8845f4a5a0c01f14d) for Redis with O1 optimization level - there are several __DOUBLE_BITS_ separate functions (each one for translation unit where it is used), and only half of them was inlined by clang. And there is a decent chance that these cases wouldn't be optimized at -O2 too (thanks @sunfishcode for noticing about it). Also, `cmath` from many libc realizations extensively uses builtins for different math operations (`isnan` and friends are in this list).


michaelvoronov:
Sorry, there were some troubles with git branches, fixed all of them.

sunfishcode:
Looks good!

PR #43, opened 2019-05-04
Fix possible integer overflow in mmap

michaelvoronov:
Fix possible integer overflow in mmap call and some other little stuff.

### DIFF
--- a/libc-bottom-half/mman/mman.c
+++ b/libc-bottom-half/mman/mman.c
@@ -58,7 +58,7 @@ void *mmap(void *addr, size_t length, int prot, int flags,
         return MAP_FAILED;
     }
 
-    // Check for overflow
+    // Check for integer overflow.
     if(sizeof(struct map) + length < sizeof(struct map)) {
         errno = EINVAL;
         return MAP_FAILED;
--- a/libc-bottom-half/mman/mman.c
+++ b/libc-bottom-half/mman/mman.c
@@ -58,10 +58,16 @@ void *mmap(void *addr, size_t length, int prot, int flags,
         return MAP_FAILED;
     }
 
+    //  To be consistent with POSIX.
+    if (length == 0) {
+        errno = EINVAL;
+        return MAP_FAILED;
+    }
+
     // Check for integer overflow.
     size_t buf_len = 0;
-    if(__builtin_add_overflow(length, sizeof(struct map), &buf_len)) {
-        errno = EINVAL;
+    if (__builtin_add_overflow(length, sizeof(struct map), &buf_len)) {
+        errno = ENOMEM;
         return MAP_FAILED;
     }
 
--- a/libc-bottom-half/mman/mman.c
+++ b/libc-bottom-half/mman/mman.c
@@ -58,6 +58,12 @@ void *mmap(void *addr, size_t length, int prot, int flags,
         return MAP_FAILED;
     }
 
+    // Check for overflow
+    if(sizeof(struct map) + length < sizeof(struct map)) {
+        errno = EINVAL;
+        return MAP_FAILED;
+    }
+
     // Allocate the memory.
     struct map *map = malloc(sizeof(struct map) + length);
     if (!map) {
@@ -76,7 +82,7 @@ void *mmap(void *addr, size_t length, int prot, int flags,
     if ((flags & MAP_ANON) == 0) {
         char *body = map->body;
         while (length > 0) {
-            ssize_t nread = pread(fd, body, length, offset);
+            const ssize_t nread = pread(fd, body, length, offset);
             if (nread < 0) {
                 if (errno == EINTR)
                     continue;
@@ -97,9 +103,6 @@ void *mmap(void *addr, size_t length, int prot, int flags,
 
 int munmap(void *addr, size_t length) {
     struct map *map = (struct map *)addr - 1;
-    off_t offset = map->offset;
-    int flags = map->flags;
-    int prot = map->prot;
     
     // We don't support partial munmapping.
     if (map->length != length) {
--- a/libc-bottom-half/mman/mman.c
+++ b/libc-bottom-half/mman/mman.c
@@ -58,14 +58,16 @@ void *mmap(void *addr, size_t length, int prot, int flags,
         return MAP_FAILED;
     }
 
+    const size_t buf_len = sizeof(struct map) + length;
+
     // Check for integer overflow.
-    if(sizeof(struct map) + length < sizeof(struct map)) {
+    if(buf_len < length) {
         errno = EINVAL;
         return MAP_FAILED;
     }
 
     // Allocate the memory.
-    struct map *map = malloc(sizeof(struct map) + length);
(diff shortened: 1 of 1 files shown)


libc-bottom-half/mman/mman.c
```diff
@@ -58,8 +58,16 @@ void *mmap(void *addr, size_t length, int prot, int flags,
         return MAP_FAILED;
     }
 
+    const size_t buf_len = sizeof(struct map) + length;
+
+    // Check for integer overflow.
+    if(buf_len < length) {
+        errno = EINVAL;
```
jedisct1: We should probably still return `ENOMEM` here, not `EINVAL`.

jedisct1: Also, you can use the `__builtin_add_overflow` builtin.

libc-bottom-half/mman/mman.c
```diff
@@ -58,10 +58,9 @@ void *mmap(void *addr, size_t length, int prot, int flags,
         return MAP_FAILED;
     }
 
-    const size_t buf_len = sizeof(struct map) + length;
-
     // Check for integer overflow.
-    if(buf_len < length) {
+    size_t buf_len = 0;
+    if(__builtin_add_overflow(length, sizeof(struct map), &buf_len)) {
         errno = EINVAL;
```
jedisct1: Thanks! Can we return `ENOMEM`? `mmap(2)` returns `EINVAL` when the base address is not page-aligned or the flags are invalid, which doesn't seem to apply here.

michaelvoronov: (i was in process of comment writing t the previous comment, so move it here:))

`__builtin_add_overflow` is good idea, but I think that `EINVAL` is more precise error type here because of:
1. According to the mmap [man](http://man7.org/linux/man-pages/man2/mmap.2.html), EINVAL - We don't like addr, length, or offset (e.g., they are **too large**, or not aligned on a page boundary).
2. Let's extrapolate (like the common way in math to find out function extremum and its properties) this situation on the case of 4 Gb user-space address space size and sizeof map == 2 Gb. And then let's imagine a user that have enough RAM and trying to allocate > 2 Gb of memory.  In this case errno would be `ENOMEM`, and, from my point of view, common user would think about that he is still not have enough memory instead of checking the length parameter.

Also there is `EOVERFLOW` (On 32-bit architecture together with the large file extension (i.e., using 64-bit off_t): the number of pages used for length plus number of pages used for offset would overflow unsigned long (32 bits).) It seems that we also could use it here :).

P.S. I have just found that `EINVAL (since Linux 2.6.12) length was 0.`. What do you think about adding checking of length == 0?

michaelvoronov: I think again about it and it seems that `EOVERFLOW` is the most nested error class suitable here and using it will lead user to more faster understanding what was going wrong. E.g., it used [here](https://github.com/CraneStation/wasi-sysroot/blob/320054e84f8f2440def3b1c8700cedb8fd697bf8/libc-top-half/musl/src/time/localtime_r.c).

jedisct1: `EOVERFLOW` may be okay. But this value is never returned on BSD and is not documented in the Linux man page, so applications may not be checking for it.

The man pages lists very specific cases where `EINVAL` is returned. All these cases are triggered by application bugs, not by hitting system limits, unlike `ENOMEM` which is a runtime error.

Although it wouldn't be an issue for our emulation, returning `EINVAL` for a zero length sounds like a good idea, if only to be consistent with other major platforms.

michaelvoronov: Sorry, but what man pages do you mean? Man from [mmap(2)](https://linux.die.net/man/2/mmap) contains definition for `EOVERFLOW`. Also, this man pages contains this definition `EINVAL - We don't like addr, length, or offset (e.g., they are too large, or not aligned on a page boundary).` that i have already mentioned. From here it clear that `EINVAL` should be used for too large length. Maybe you used BSD-like man pages?

And, yes, we can look on this problem from the side of runtime/compile time problem. `ENOMEM` is runtime error, but we cannot say that overflow belong to only one of these error class - it depends only on current length parameter and compile-time size of map struct and length could be known in compile time or depends on some user input as an example. But I believe that it is wrong to belong in only to runtime class in context of no memory situation - because really it isn't depend anyhow on current memory amount in the OS.

jedisct1: OpenBSD [doesn't have `EOVERFLOW` listed](http://man.openbsd.org/mmap) as a possible return value.

The Linux man page (up-to-date Arch Linux) starts the description of that return value with "[The file is a regular file](https://pubs.opengroup.org/onlinepubs/7908799/xsh/mmap.html)". 

Same restriction [on FreeBSD](https://www.freebsd.org/cgi/man.cgi?query=mmap&sektion=2&apropos=0). 

But the file doesn't have to be a regular file in our case.

Returning anything would be fine, to be honest. But `ENOMEM` has a higher probability to be properly handled by applications than others.

sunfishcode: That's the Linux man page. The [POSIX documentation](http://pubs.opengroup.org/onlinepubs/9699919799/functions/mmap.html) seems to prefer ENOMEM when "there is insufficient room in the address space to effect the mapping" which covers this case. And since the add here is an implementations detail, EINVAL wouldn't make it clear what the actual problem is. So I agree with @jedisct1 that ENOMEM seems best here.

libc-bottom-half/mman/mman.c
```diff
@@ -58,8 +58,15 @@ void *mmap(void *addr, size_t length, int prot, int flags,
         return MAP_FAILED;
     }
 
+    // Check for integer overflow.
+    size_t buf_len = 0;
+    if(__builtin_add_overflow(length, sizeof(struct map), &buf_len)) {
```
sunfishcode: Style nit: add a space between the `if` and the `(`.

libc-bottom-half/mman/mman.c
michaelvoronov: I think that It some kind of philosophical question, because `ENOMEM` from POSIX isn't express this error as accurate as possible. Maybe that why Linux man pages define new cases for `EINVAL` and describe EOVERFLOW another way (linux kernel [returns EOVERFLOW]((https://github.com/torvalds/linux/blob/master/mm/mmap.c#L1422)) in this case [whatever file is](https://github.com/torvalds/linux/blob/master/mm/mmap.c#L1462)).
Anyway, If it needs to be strictly compatible with POSIX, It seems the best way is to use `ENOMEM` here (and also I agree with @sunfishcode as a owner of this repo).

michaelvoronov:
Fixed comments and also added failing if length == 0 to be compatible with POSIX (`If len is zero, mmap() shall fail and no mapping shall be established` from [here](http://pubs.opengroup.org/onlinepubs/9699919799/functions/mmap.html)).

sunfishcode:
Good call on handling `length == 0` too.

PR #50, opened 2019-05-07
Don't declare lockfile and timezone functions.

sunfishcode:
These aren't defined, so don't declare them either.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -384,7 +384,7 @@ include_dirs:
 	      "$(SYSROOT_INC)/pty.h" \
 	      "$(SYSROOT_INC)/dlfcn.h" \
 	      "$(SYSROOT_INC)/ulimit.h" \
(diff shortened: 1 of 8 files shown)


jedisct1:
Would it make sense to add a syscall for file locking?

sunfishcode:
Oops, I got the comment mixed up here. `flockfile` etc. are for locking a `FILE` in multi-threaded programs, not POSIX file locking. We'll enable them when we support threads.

(File locking would also be useful, the reason WASI Core doens't include them is that POSIX file locks are [per-process](https://gavv.github.io/articles/file-locks/#posix-record-locks-fcntl), which makes them awkward to fit into WASI, and which is a significant limitation.)

sunfishcode:
This PR now removes several more unimplemented functions and headers, and is ready for review!

PR #54, opened 2019-05-07
AT_FDCWD support

### LINKED ISSUE #42: AT_FDCWD undefined?
I'm getting this issue building gnulib areadlink.c. The source is available here:

https://github.com/coreutils/gnulib/blob/master/lib/areadlink.c

The error is this:

```
wasm32-unknown-wasi-clang -DHAVE_CONFIG_H -DEXEEXT=\"\" -I. -I..  -I../intl -I../intl -DDEPENDS_ON_LIBICONV=1 -DDEPENDS_ON_LIBINTL=1    -g -O2 -c -o areadlink.o areadlink.c
areadlink.c:41:13: error: use of undeclared identifier 'AT_FDCWD'
  if (fd != AT_FDCWD)
            ^
areadlink.c:55:24: error: use of undeclared identifier 'AT_FDCWD'
  return careadlinkat (AT_FDCWD, filename, NULL, 0, NULL, careadlinkatcwd);
                       ^
2 errors generated.
make[6]: *** [Makefile:1496: areadlink.o] Error 1
```

AT_FDCWD should be available in wasi, correct? I can't find any direct references outside of Musl though. If it's not provided, what alternatives are available?

You can repro this with nix:

```
$ nix build -f channel:nixpkgs-unstable pkgsCross.wasi32.gettext --arg config '{ allowUnsupportedSystem = true; }'
```

/cc @ericson2314

sunfishcode:
This implements `AT_FDCWD`, as well as absolute paths in `*at` functions, following the scheme described [here](https://github.com/CraneStation/wasi-sysroot/issues/42#issuecomment-489182928).

This is a relatively elaborate change, and is not at all tested yet, so I'm inclined to take some time here to think this through.

Fixes #42.


Ericson2314:
Looking good!

sunfishcode:
Ok, I think this is in pretty good shape now.

sunfishcode:
This will need to be revised in terms of #214.

sunfishcode:
An updated version of this patch is now at #230.

PR #62, opened 2019-05-09
Import libc-test and adapt it for use with WASI.

sunfishcode:
This is a preliminary PR importing musl's [libc-test](https://wiki.musl-libc.org/libc-test.html) testsuite and adapting it for WASI. It's already helped find several bugs in WASI libc.


sunfishcode:
This is still work-in-progress, and currently depends on several fixes in separate PRs, but overall it's making good progress, and has helped find numerous bugs already.

sbc100:
Any chance we could import this a git submodule?  This repo could just contain the run script and list of expected pass/fail result?

sunfishcode:
If there's a way to do that, I'm open to it. The trick is, many of these tests aren't fine-grained. For example, the `stdio.c` API test has a lot of tests for the contents of `<stdio.h>`, most of which which we want, however it also tests for `popen`, which POSIX defined in `<stdio.h>`, but which WASI libc doesn't implement. With local modifications, [like this](https://github.com/CraneStation/wasi-sysroot/pull/62/files#diff-cf40c797de5bd4743f0833ff67a3738cR82), we can disable just the parts that aren't supported, and still make use of the rest of the tests.

sunfishcode:
libc-test's math tests include some data files which are under other licenses -- one is a non-commercial "the software may not be redistributed for a fee except as reasonable to cover media costs", and another is GPLv2. So I'm now investigating moving it into a separate repo with a git submodule.

sunfishcode:
I've now started [wasi-libc-test](https://github.com/CraneStation/wasi-libc-test), and populated it with the contents of this PR.

PR #72, opened 2019-05-29
Ensure argv is null terminated.

sbc100:
This matches the behavior of populate __environ.

See https://github.com/WebAssembly/WASI/issues/27


sunfishcode:
Makes sense.

The alternative here is to say that it's always the WASI implementation's job to add the NULL terminator, however doing it in crt1.c here is simpler for non-C-family languages that don't require the NULL terminator, so let's go with that.

PR #74, opened 2019-06-05
Reactor support.

sunfishcode:
This adds support for a new "Reactor" executable model.

The "Commands" and "Reactors" concepts are introduced here:
https://github.com/WebAssembly/WASI/issues/13

A companion Clang patch, which just consists of using the new
crt1-reactor.o and Reactor-specific entry point name, is here:
https://reviews.llvm.org/D62922

Instead of an entrypoint named "_start", which calls "main", which
then scopes the lifetime of the program, Reactors have a
"_initialize" function. When "_initialize" exits, the intention is that the program should
persist and be available for calling.

At present, the main anticipated use for this is in environments like
Node, where WASI-using modules can be imported and don't necessarily
want the semantics of a "main" function.

[edit: update to reflect design changes, and drop "experimental"]


Makefile
```diff
@@ -418,7 +418,8 @@ startup_files: include_dirs
 	#
 	@mkdir -p "$(OBJDIR)"
 	cd "$(OBJDIR)" && \
-	"$(WASM_CC)" $(WASM_CFLAGS) -c $(CRT_SOURCES) -MD -MP && \
+	"$(WASM_CC)" $(WASM_CFLAGS) -c $(CRT1_SOURCE) -MD -MP && \
```
sbc100: Can you add `-o crt1.o` to this line for clarity?

Also, can you avoid the "cd" and put the OBJDIR in the `-o` flag to avoid the line continuation stuff?

basics/crt/crt1.c
```diff
@@ -1,12 +1,25 @@
 extern void __wasm_call_ctors(void);
+void _Exit(int) __attribute__((noreturn));
+
+#ifdef REACTOR_RUNTIME
+extern void reactor_setup(int, char *[]);
```
sbc100: Should we make this weak?  So people don't have to define a setup function unless they want to?

Also, I'm not sure about this name. I feel like it should have wasi in the name, and also that it probably shouldn't have argv at all.    I feel like argv is a very commandliney thing and reators probably shouldn't have that at all.   But I understand this an early prototype ..

Makefile
sbc100: In this long run I think it will be nicer to move the common code here into libc.a and to have these two be separate source files (crt1.c + crt1-reactor.c) that contain very little code.

sunfishcode: `-o crt1.o` and the line continuation stuff is now cleaned up.

I've also started refactoring this code, but there's more to do.

basics/crt/crt1.c
sunfishcode: Making it weak is a good idea, that's done.

I'm also not sure about the name. Or passing in argv. These are still things to think about.

Makefile
```diff
@@ -32,8 +32,9 @@ endif
 # These variables describe the locations of various files and
 # directories in the source tree.
 BASICS_DIR = $(CURDIR)/basics
-BASICS_INC = $(BASICS_DIR)/include
-BASICS_CRT_SOURCES = $(wildcard $(BASICS_DIR)/crt/*.c)
+BASICS_HEADERS_PUBLIC = $(BASICS_DIR)/headers/public
+BASICS_HEADERS_PRIVATE = $(BASICS_DIR)/headers/private
```
sbc100: This header file rename seems logically separate no?   Why not continue to using `include` ?  `include/public` +  `include/private`?

basics/crt/crt1.c
sbc100: sgtm

sunfishcode:
I've now rewritten this patch, which is much easier now that a bunch of other pieces of startup have been factored out. The patch is now *very* simple.

See also the corresponding WASI PR: https://github.com/WebAssembly/WASI/pull/256

libc-bottom-half/crt/reactor-crt1.c
```diff
@@ -0,0 +1,6 @@
+extern void __wasm_call_ctors(void);
+
+void _activate(void) {
```
sbc100, line 3: Should we add `attribute(export-name("_activate"))` so that its is exported from the wasm module without additional linker flags?

sbc100: I guess that `--entry=_activate` is an alternative to this.    If we use the attribute we could pass `--no-entry` for reactors.  Not sure that is preferable...but I do like that fact the using that attributing means that simply including this version crt1.o will cause the export to exist.

sunfishcode: Ooh. I'd been expecting to submit clang and wasm-ld patches to make them do the right thing when -mexec-model=reactor is given, however using `export_name` is an interesting idea that potentially simplifies things, and more precisely describes what's going on. I've now added a patch which does that.

sunfishcode:
This is now updated to use `_initialize`, and to use "crt1-reactor.o" as suggested a while ago in https://reviews.llvm.org/D62922.

And while here, I noticed that wasi-libc was exporting a `_fini` function which wasn't actually used, so I added a fix for that too.

sunfishcode:
Rebased and ready for review!

libc-bottom-half/crt/crt1.c
```diff
@@ -3,6 +3,7 @@ extern void __wasm_call_ctors(void);
 extern int __original_main(void);
 extern void __prepare_for_exit(void);
 
+__attribute__((export_name("_start")))
 void _start(void) {
```
sbc100, line 7: Remind my why this file is duplicated here and in basics/crt/crt1.c?

sbc100:
The PR description seems out-of-date.      

I was kind of expecting to see some change to Makefile to build the new version of crt1?

libc-bottom-half/crt/crt1.c
sunfishcode: Good point; it's an artifact of the "reference sysroot" idea that isn't important anymore. I'll submit a separate PR to remove it.

sunfishcode:
The Makefile is set up to compile all .c files in the crt directory into .o files in the sysroot, so just adding a new .c file here is sufficient.

sunfishcode:
Title and PR description updated!

PR #75, opened 2019-06-24
Update expected defines after upstream llvm change

sbc100:
As of https://reviews.llvm.org/D63030 upstream clang includes its
builtin include paths before any of the sysroot include paths.

This means that several header files previously coming from libc are
now coming llvm's compiler-builtin include path.

This does make this test more fragile.  I'm open to suggestions on
how to elevate this problem.  Perhaps we should trim more of the
output before comparing?

### DIFF
--- a/expected/wasm32-wasi/predefined-macros.txt
+++ b/expected/wasm32-wasi/predefined-macros.txt
@@ -92,9 +92,9 @@
 #define CFLUSH CDISCARD
 #define CHARBITS (sizeof(char) * 8)
 #define CHARCLASS_NAME_MAX 14
-#define CHAR_BIT 8
-#define CHAR_MAX 127
-#define CHAR_MIN (-128)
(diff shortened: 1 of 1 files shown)


sunfishcode:
As a temporary workaround, what if we added an explicit `-isystem $(SYSROOT_INC)` to [the compile command](https://github.com/CraneStation/wasi-libc/blob/master/Makefile#L472) which generates predefined_macros.txt, to force it to always use the libc headers first?

Longer term, we probably do want some more intelligent filtering here. predefined_macros.txt is really useful for development; it puts a spotlight on changes to what macros are defined, which makes reviewing patches easier and often catches simple mistakes, so I'd like to keep it it we can, but as the compiler landscape gets more diverse, it may get more complex to maintain.

sbc100:
Closing in favor of #84

PR #78, opened 2019-06-25
Add Azure Pipelines definition

tschneidereit:
This sets up Azure Pipelines CI for Linux, Mac, and Windows. To make in particular the Windows build work, a number of Makefile fixes are required—those are included in this PR, too.


azure-pipelines.yml
```diff
@@ -0,0 +1,52 @@
...
+        rustup update stable
+        rustup default stable
+        rustup component add llvm-tools-preview
+        echo "##vso[task.setvariable variable=WASM_NM;]$(rustc --print sysroot|sed 's|C:|/c|'|sed 's|\\|/|g')/lib/rustlib/x86_64-pc-windows-msvc/bin/llvm-nm.exe"
+      displayName: Install llvm-nm (Windows)
+      condition: and(succeeded(), eq( variables['Agent.OS'], 'Windows_NT' ))
+    - bash: |
+        set -e
+        make -j4
+      displayName: Build
```
sbc100: Missing newline at EOF

azure-pipelines.yml
```diff
@@ -0,0 +1,52 @@
...
+        curl -f http://releases.llvm.org/8.0.0/clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz | tar xJf -
+        export CLANG_DIR=`pwd`/clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-16.04/bin
+        echo "##vso[task.prependpath]$CLANG_DIR"
+      displayName: Install clang (Linux)
+      condition: and(succeeded(), eq(variables['Agent.OS'], 'Linux'))
+    - bash: |
+        set -e
+        rustup update stable
+        rustup default stable
+        rustup component add llvm-tools-preview
```
sbc100: I don't understand why rust is has anything to do with installing `llvm-nm` . Are there not official llvm releases we can use?

azure-pipelines.yml
```diff
@@ -0,0 +1,52 @@
...
+        echo "##vso[task.setvariable variable=WASM_NM;]$(rustc --print sysroot|sed 's|C:|/c|'|sed 's|\\|/|g')/lib/rustlib/x86_64-pc-windows-msvc/bin/llvm-nm.exe"
+      displayName: Install llvm-nm (Windows)
+      condition: and(succeeded(), eq( variables['Agent.OS'], 'Windows_NT' ))
+    - bash: |
+        set -e
+        make -j4
```
sbc100: Since this is just a one liner for now maybe just `- bash: make -j4`.  Or is there a way to not even run bash?

azure-pipelines.yml
```diff
@@ -0,0 +1,52 @@
+trigger:
```
sbc100: Is there a filename format that starts with a period?  like `.travis.yml`?

Makefile
```diff
@@ -245,7 +245,10 @@ $(SYSROOT_LIB)/libwasi-emulated-mman.a: $(LIBWASI_EMULATED_MMAN_OBJS)
 
 %.a:
 	@mkdir -p "$(@D)"
-	$(WASM_AR) crs $@ $^
+	$(WASM_AR) crs $@ $(wordlist 1, 200, $(patsubst $(CURDIR)/%.o,%.o,$^))
+	$(WASM_AR) crs $@ $(wordlist 200, 400, $(patsubst $(CURDIR)/%.o,%.o,$^))
+	$(WASM_AR) crs $@ $(wordlist 400, 600, $(patsubst $(CURDIR)/%.o,%.o,$^))
+	$(WASM_AR) crs $@ $(wordlist 600, 1000, $(patsubst $(CURDIR)/%.o,%.o,$^))
```
sbc100: Ooo.  This is pretty ugly.  Is this because of the command line limit?  I wonder if we can find some other way?  Perhaps a response file?    We should at least add a comment.  Also the final increment is bigger than the others, is that deliberate?

Makefile
```diff
@@ -267,7 +270,7 @@ $(OBJDIR)/%.o: $(CURDIR)/%.c include_dirs
 	@mkdir -p "$(@D)"
 	"$(WASM_CC)" $(WASM_CFLAGS) -MD -MP -o $@ -c $<
 
-include $(shell find $(OBJDIR) -name \*.d) /dev/null
+-include $(shell find $(OBJDIR) -name \*.d)
```
sbc100: I assume the `/dev/null` was there to prevent some kind of warning when no .d files are found?     Is there such a warning now?

pchickey:
Agree with @sbc100 - it would be great to package releases for the various platforms here. However, even the platforms that bring their own llvm will probably need the compiler-rt, libcxx, and libcxxabi that is built in cranestation/wasi-sdk. Is the pipelineization of that repo up next?

tschneidereit:
Thank you for all the feedback! I'll add some comments to the Makefile and the .yml file to explain some of the more absurd contortions required to make Windows happy ...

As for creating releases—I do plan to do that, yes. I however also agree with @pchickey: a bundle that also contains compiler-rt, libcxx, and libcxxabi will be more helpful. That might well deserve the name `wasi-sysroot`, too :D

sbc100:
Ah, ok, I guess the sysroot should probably be build and released from wasi-sdk then.    We should have wasm-sdk binaryies which include both llvm and the sysroot, and wasi-sysroot releases which don't have any host tools.

Makefile
```diff
@@ -1,13 +1,13 @@
...
+WASM_CC ?= clang
+WASM_NM ?= $(patsubst %clang,%llvm-nm,$(WASM_CC))
+WASM_AR ?= $(patsubst %clang,%llvm-ar,$(WASM_CC))
+WASM_CFLAGS ?= -O2
 # The directory where we build the sysroot.
-SYSROOT = $(CURDIR)/sysroot
+SYSROOT ?= $(CURDIR)/sysroot
 # A directory to install to for "make install".
-INSTALL_DIR = /usr/local
+INSTALL_DIR ?= /usr/local
```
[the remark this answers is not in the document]
tschneidereit: The problem is that we need to pipe this through multiple make invocations in the case of wasi-sdk. I guess we could do that explicitly, but I'd have a preference for not infecting the Makefiles with essentially CI-specific bits and pieces.

I added `?=` to all of these to make them consistent, but I'm more than happy to do whatever you prefer here, as long as it makes the CI work :)

Makefile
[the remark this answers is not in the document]
tschneidereit: Correct: the command line was something like ~65k characters long, and Windows has a limit of 32k chars. I'll add a comment explaining this, and fix the two issues you point out. Excellent call on the bad failure mode!

Makefile
```diff
@@ -493,7 +496,7 @@ finish: startup_files libc
 
 check: finish
 	# Check that the computed metadata matches the expected metadata.
-	diff -ur "$(CURDIR)/expected/$(MULTIARCH_TRIPLE)" "$(SYSROOT_SHARE)"
+	diff -wur "$(CURDIR)/expected/$(MULTIARCH_TRIPLE)" "$(SYSROOT_SHARE)"
```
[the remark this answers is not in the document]
tschneidereit: Will do!

azure-pipelines.yml
tschneidereit: Yeah, this certainly requires some explanation :)

The reason this is needed is that, weirdly, the LLVM release for Windows doesn't include `llvm-nm`. Rust's `llvm-tools-preview` component includes `llvm-nm`, and Azure Pipelines machines have Rust preinstalled, so was the easiest way to make `llvm-nm` available that I could come up with.

I'll add a comment explaining this.

azure-pipelines.yml
tschneidereit: This script used to contain more steps, but you're right—it can be one line by now. And it also doesn't need to run bash, given that the env var is set using `task.setvariable`. Will fix.

azure-pipelines.yml
tschneidereit: I renamed the file to start with a `.`

Makefile
tschneidereit: `llvm-ar` doesn't take an input file, so I'm not sure there really is an alternative. I do certainly agree it's ugly :)

I added a comment explaining this, and also simplified and fixed things up a bit based on @sunfishcode's feedback.

Makefile
tschneidereit: Probably not :) Unfortunately Windows bails on `/dev/null`, so something has to change about this. I'm open to suggestions for how to keep the warning.

tschneidereit:
> Is the pipelineization of that repo up next?

@pchickey I actually started with that repo, but kept running into `wasi-libc`-related issues, so decided to fix those in isolation first :)

.azure-pipelines.yml
```diff
@@ -0,0 +1,53 @@
...
+      condition: and(succeeded(), eq(variables['Agent.OS'], 'Darwin'))
+    - bash: |
+        set -euo pipefail
+        curl -f http://releases.llvm.org/8.0.0/clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz | tar xJf -
+        export CLANG_DIR=`pwd`/clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-16.04/bin
+        echo "##vso[task.prependpath]$CLANG_DIR"
+      displayName: Install clang (Linux)
+      condition: and(succeeded(), eq(variables['Agent.OS'], 'Linux'))
+    - bash: |
+        # Windows releases of LLVM don't include the llvm-nm tool, which is needed for building
```
sbc100: That seems rather odd.  Should we open an llvm bug to get it added?   Thanks for the explanation.

Makefile
sbc100: I meant that we *don't* want a warning there are no .d files yet generated.   At least thats the only reason I can see for adding this?  But I don't really know why it was added.   Dan do you remember why you added this originally?

Makefile
sbc100: It looks like `llmv-ar` does take a response file but `--help` didn't document it.  I made an llvm change to add to the `--help` message: https://reviews.llvm.org/D63836.

I'm not totally sure it helps here but perhaps its an option.  Not a blocker for this change landing.

Makefile
sunfishcode: I don't think I had anything subtle in mind for this one; I just didn't think of `-include`.

Makefile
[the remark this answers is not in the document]
sunfishcode: Ok, `?=` works for me.

sbc100:
Anything still blocking this?

tschneidereit:
> Anything still blocking this?

Nothing was blocking it, no—this just fell off my radar :(

PR #84, opened 2019-07-12
Fix predefined macros check after recent clang change

sbc100:
This is an alternative to #75.


sbc100:
Ping

sunfishcode:
Cool! We may eventually get to the point where it's too awkward to support both clang 8.0 and 9.0 in the same tree, but this lets us keep doing so for now :-).

PR #88, opened 2019-07-19
Update to musl 1.1.23.

sunfishcode:
See the WHATSNEW file for details; among other things, this contains new
math library implementation for log/exp/pow.


sunfishcode:
This isn't urgent at all, and can wait until we have a CI setup.

tschneidereit:
The main question for CI for this repo is how to run tests. I guess we could set things up such that prebuilt binaries of the SDK are downloaded and their tests run against new libc builds? That seems awfully circular however :/

sbc100:
happy for this to land if you've done some local testing.

pchickey:
I have kicked off a build to run local tests against lucet (via wasi-sdk), I don't think that it will give particularly good coverage but its better than nothing.

pchickey:
Passes the Lucet test suite (31368eea) locally

sunfishcode:
I've now completed a [libc-test](https://github.com/CraneStation/wasi-libc/pull/62) run on this branch. There was only one unexpected failure:

src/math/special/erfc.h:6: RN erfc(0x1.5db559fe5c0bap+0) want 0x1.b53cf571d328fp-5 got 0x1.b53cf571d328cp-5 ulperr -2.609 = -0x1.8p+1 + 0x1.900982p-2

This is a minor rounding difference, and the most likely cause of this is the [new exp implementations](http://git.musl-libc.org/cgit/musl/commit/src/math/exp.c?id=e16f7b3c02e17d0ace779a11f0d53a9c05fdd434), as erfc uses exp, though I haven't investigated to confirm this yet.

sunfishcode:
I've now confirmed that the new exp implementation is what causes the failure, and reproduced the failure on x86-64. I've now also reported this upstream: https://www.openwall.com/lists/musl/2019/08/08/1

The failure is a very minor rounding difference, so I don't think it's important for us here to block on.

sunfishcode:
For completeness, the failure turned out to be due to the test, so the code in libc is ok.

PR #89, opened 2019-07-19
Set the MALLOC_ALIGNMENT to 16?

sunfishcode:
This isn't strictly required, as wasm SIMD loads and stores work on
unaligned memory. However, it may provide better performance. That said,
this isn't currently studied by any benchmarking.

It's not clear what the best approach is here. I'm opening this PR to start a discussion about what the value should be.

### DIFF
--- a/dlmalloc/src/dlmalloc.c
+++ b/dlmalloc/src/dlmalloc.c
@@ -29,6 +29,9 @@
 #define NO_MALLINFO 1
 #define NO_MALLOC_STATS 1
 
+/* Align malloc regions to 16, to avoid unaligned SIMD accesses. */
+#define MALLOC_ALIGNMENT 16
+
 /*
  * Declare errno values used by dlmalloc. We define them like this to avoid
  * putting specific errno values in the ABI.


jedisct1:
16 byte alignment is not a bad idea. Besides improving performance, this is consistent with what most allocators are doing.

By the way, I've been experimenting with Microsoft's [mimalloc](https://github.com/microsoft/mimalloc) as a replacement for dlmalloc. The `dev` branch supports WebAssembly/WASI out of the box.

Mimalloc has several advantages over the current allocator. Besides dealing very well with fragmentation, it has interesting security features (encrypted free lists, randomization even without `mmap()`) that are worth having in an environment that only exposes linear, unprotected memory.

I have a branch of `wasi-libc` replacing `dlmalloc` with a stripped-down version of `mimalloc`. The only issue at the moment is the fact that it has quite a lot of internal symbols that visibility attributes don't hide in a static library.

Oh, and it uses 16 bytes alignment.

sunfishcode:
Mimalloc sounds cool, thanks for bringing that up! I didn't put a lot of thought into the initial choice of dlmalloc, so I'm interested in hearing how it compares to alternatives.

> Mimalloc has several advantages over the current allocator. Besides dealing very well with fragmentation, it has interesting security features (encrypted free lists, randomization even without `mmap()`) that are worth having in an environment that only exposes linear, unprotected memory.

I'd be interested to see how it compares on code size too.

Depending on how things work out, I could see WASI libc offering multiple malloc implementations, allowing users to select which one they want, perhaps at link time. Some users won't want the randomization, and there will always be tradeoffs around performance and other factors.

> I have a branch of `wasi-libc` replacing `dlmalloc` with a stripped-down version of `mimalloc`. The only issue at the moment is the fact that it has quite a lot of internal symbols that visibility attributes don't hide in a static library.

Ah. Would you be interested in opening a draft PR for this?

sunfishcode:
Any objections to landing this?

jedisct1:
No objections.

PR #107, opened 2019-10-12
adjust S_IFIFO to avoid collisions

vapier:
S_IFIFO currently has the same value as S_IFSOCK which breaks code
that does things like "switch (mode & S_IFMT)" as we end up with
duplicate labels.  Move S_IFIFO to a unique unused value.

### DIFF
--- a/expected/wasm32-wasi/predefined-macros.txt
+++ b/expected/wasm32-wasi/predefined-macros.txt
@@ -1717,7 +1717,7 @@
 #define S_IFBLK (0x6000)
 #define S_IFCHR (0x2000)
 #define S_IFDIR (0x4000)
-#define S_IFIFO (0xc000)
+#define S_IFIFO (0x1000)
 #define S_IFLNK (0xa000)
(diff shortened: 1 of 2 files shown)


vapier:
i guess more importantly, the current defines make it impossible to differentiate between fifos & sockets :)

vapier:
all of that sounds fine.  i'll try to tackle it when i get a chance.  thx!

vapier:
i posted https://github.com/WebAssembly/WASI/pull/189 to add FIFO filetype

vapier:
my WASI PR is merged now. was there something else needed here before merging?

sunfishcode:
Before we merge changes into wasi-libc which WASI implementations will need to support, we'll need the WASI Subgroup to publish a new [snapshot](https://github.com/WebAssembly/WASI/tree/master/phases/). There isn't a specific schedule for that yet, but it hopefully won't be too long.

vapier:
i think a new snapshot has been posted by now ?  so can we merge ?

sunfishcode:
Unfortunately no, we have not published a new snapshot yet. Snapshot 2 is set to include modularization, which requires updates to the witx tooling, which is not finished yet. Hopefully once we're past this milestone, subsequent snapshots can be published more frequently.

PR #108, opened 2019-10-13
fix the typo

michaelvoronov:


libc-bottom-half/sources/sbrk.c
```diff
@@ -15,7 +15,7 @@ void *sbrk(intptr_t increment) {
         abort();
     }
 
-    uintptr_t old = __builtin_wasm_memory_grow(0, (uintptr_t)increment / PAGESIZE);
+    const uintptr_t old = __builtin_wasm_memory_grow(0, (uintptr_t)increment / PAGESIZE);
```
sbc100: Is there any value in adding a const here?

michaelvoronov: Just to follow good coding style (in case if function will be extended).

PR #109, opened 2019-10-16
Call `populate_environ` only if we actually need environment variables.

sunfishcode:
This avoids a couple of mallocs and calls to __wasi_environ_sizes_get
and __wasi_environ_get in programs that don't use environment variables.

In theory this should work, however it doesn't currently work as
intended, and seems to be blocked by [LLVM PR43696].

[LLVM PR43696]: https://bugs.llvm.org/show_bug.cgi?id=43696

### DIFF
--- a/libc-bottom-half/sources/__environ.c
+++ b/libc-bottom-half/sources/__environ.c
@@ -21,12 +21,15 @@ __wasi_errno_t __wasilibc_populate_environ(void) {
         return err;
     }
 
-    // Allocate memory for the array of pointers, adding null terminator.
-    __environ = malloc(sizeof(char *) * (environ_count + 1));
-
     // Allocate memory for storing the environment chars.
     char *environ_buf = malloc(sizeof(char) * environ_buf_size);
-    if (__environ == NULL || environ_buf == NULL) {
+    if (environ_buf == NULL) {
+        return __WASI_ENOMEM;
+    }
+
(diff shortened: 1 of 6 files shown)


sunfishcode:
Actually, I figured out an even better approach. The PR no longer *links* in populate_args if the application doesn't call environment variable routines.

libc-bottom-half/sources/__environ.c
```diff
@@ -0,0 +1,37 @@
...
+    /* Get the sizes of the arrays we'll have to create to copy in the environment. */
+    size_t environ_count;
+    size_t environ_buf_size;
+    err = __wasi_environ_sizes_get(&environ_count, &environ_buf_size);
+    if (err != __WASI_ESUCCESS) {
+        return err;
+    }
+
+    /* Allocate memory for the array of pointers, adding null terminator. */
+    __environ = malloc(sizeof(char *) * (environ_count + 1));
```
michaelvoronov: I think it would be good to add int overflow checks here, because a compiled binary could be launched on different execution environments with different lifecycle. E.g. on some sandboxed application this could potentially be a vulnerability.

sunfishcode: Good catch, thanks! Looking at this more closely, there were also cases where memory could be leaked or `__environ` could be left pointing to uninitialized memory if a failure happened partway through. I've now fixed these.

PR #110, opened 2019-10-16
Reorganize and simplify the libpreopen code

sunfishcode:
wasi-libc's copy of libpreopen has evolved so many local changes that it's no longer worth keeping the upstream code structure and marking changes with `__wasilibc_unmodified_upstream`.

This PR merges the source files into a single file, removes all `__wasilibc_unmodified_upstream` code, eliminates the ability to allocate multiple preopen lists, eliminates the need for `__wasilibc_init_preopen`, eliminates the non-standard `eaccess`, and makes several other cleanups. It also enables `NDEBUG` so that internal assertions are disabled in release builds.


sunfishcode:
I've now simplified this PR, by removing two followup optimizations (which I'll submit separately), to make it easier to review.

sunfishcode:
It still is derived from libpreopen code, so it's still useful to keep it in its own directory since it has a different license from the code in `sources`.

PR #111, opened 2019-10-16
Update to musl 1.2.0.

sunfishcode:
See the WHATSNEW file for details; this doesn't have any major changes
for wasi-libc; in particular, the new catgets and GLOB_TILDE features
are disabled.


sunfishcode:
Now updated for musl 1.2.0.

See the WHATSNEW file for details. The biggest change in musl is the
switch to 64-bit time_t for 32-bit targets, however WASI libc was already
using 64-bit time_t. The main change affecting WASI is an update to
Unicode 12.1.0.

sunfishcode:
Thanks! Yes, the wasi-sdk and wasi-libc-test tests pass.

PR #114, opened 2019-10-19
Use __heap_base by dlmalloc

michaelvoronov:
Force dlmalloc to use free memory between __heap_base and initial. Some details described [here](https://github.com/CraneStation/wasi-libc/issues/94#issuecomment-539045750).


dlmalloc/src/malloc.c
```diff
@@ -5197,6 +5209,43 @@ static void internal_inspect_all(mstate m,
 }
 #endif /* MALLOC_INSPECT_ALL */
 
+#ifdef __wasilibc_unmodified_upstream // Define a function that initializes the initial state of dlmalloc
+#else
```
bjorn3: You could use `#ifndef`.

michaelvoronov: Hi!

Yes, but It seems that current repo has pattern of `#ifdef` {empty} + `#else` as a coding style requirement. E.g. you can find a lot of such patterns in wasi-libc and no patterns with `#ifndef`.

I think that pattern with `#else` in some situations provides more easy way to update from upstream.

@sunfishcode is it true?

sunfishcode: Yeah, that is what we do elsewhere in wasi-libc. And yes, I sometimes write scripts when working with upstream changes, which are slightly simpler if all the conditionals run the same way. That said, it's not hugely important, so we could change the convention if people find it odd.

dlmalloc/src/malloc.c
```diff
@@ -4588,6 +4593,13 @@ void* dlmalloc(size_t bytes) {
   ensure_initialization(); /* initialize in sys_alloc if not using locks */
 #endif
 
+#if __wasilibc_unmodified_upstream // Try to initialize the allocator.
+#else
+  if (!is_initialized(gm)) {
+    try_init_allocator();
+  }
+#endif
```
sunfishcode: Would it work to call this from within `init_mparams` instead? That way it'd get called for all entrypoints, not just `malloc`.

In that case, it shouldn't do `ensure_initialization` itself, but just do its work after `init_mparams` has done its work.

dlmalloc/src/malloc.c
```diff
@@ -5197,6 +5209,44 @@ static void internal_inspect_all(mstate m,
...
+#else
+/* ------------------ Exported try_init_allocator -------------------- */
+
+/* Symbol marking the end of data, bss and explicit stack, provided by wasm-ld. */
+extern unsigned char __heap_base;
+
+/* Initialize the initial state of dlmalloc to be able to use free memory between __heap_base and initial. */
+void try_init_allocator(void) {
+  /* Check that it is a first-time initialization. */
+  if (is_initialized(gm)) {
```
sunfishcode: Since we only call `try_init_allocator` when `is_initialized(gm)` is true, this can be an `assert` rather than a failure condition.

michaelvoronov: Good point

dlmalloc/src/malloc.c
michaelvoronov: Hmmm, i've though about it, but there are some clues:
1. `ensure_initialization` is used for mspace (mspace is like Arena in malloc for different continuous space of allocations). At now, mspace is disabled, but if someone want to enable it, there will some unobvious problems. And it is logically incorrect.
2. Currently, `ensure_initialization` is disabled in `dlmalloc`, because it is inside `#if USE_LOCKS` that is also disabled. And it is called into `sys_alloc`. But for our purpose it is important to have it initialized before main logic of chunk choosing in `dlmalloc` being called. => in case of  `ensure_initialization` we will need two augmentation (into `ensure_initialization` and `dlmalloc`) instead of one in current case.
3. On the other side this initialization into `dlmalloc` seems correct, because other entry point (`realloc`, `calloc`) in all important CFG paths will call it before any chunk manipulation. The only exception here is `free`, but it is UB to call `free` without any previous `malloc`.

So, yes, it possible to move top chunk initialization into `insure_initialization`, but it seems that it is better leave it as now.

sunfishcode: Thanks for looking into that! I think what you have right now looks like a reasonable approach then.

sunfishcode:
Merging; thanks for implementing this!

michaelvoronov:
mimalloc is on the way :)

PR #118, opened 2019-10-22
Fix corner cases and style in the argv/argc code.

sunfishcode:
This fixes some potential memory leaks in pathological situations, and
updates the code to use `//`-style comments.

This is a subset of https://github.com/CraneStation/wasi-libc/pull/112, extracted for easier reviewing.

### DIFF
--- a/libc-bottom-half/crt/crt1.c
+++ b/libc-bottom-half/crt/crt1.c
@@ -13,29 +13,47 @@ void _Exit(int) __attribute__((noreturn));
 static __wasi_errno_t populate_args(size_t *argc, char ***argv) {
     __wasi_errno_t err;
 
-    /* Get the sizes of the arrays we'll have to create to copy in the args. */
+    // Get the sizes of the arrays we'll have to create to copy in the args.
     size_t argv_buf_size;
-    err = __wasi_args_sizes_get(argc, &argv_buf_size);
+    size_t new_argc;
+    err = __wasi_args_sizes_get(&new_argc, &argv_buf_size);
     if (err != __WASI_ESUCCESS) {
         return err;
     }
-    if (*argc == 0) {
+    if (new_argc == 0) {
         return __WASI_ESUCCESS;
     }
 
-    /* Allocate memory for the array of pointers, adding null terminator. */
-    *argv = malloc(sizeof(char *) * (*argc + 1));
-    /* Allocate memory for storing the argument chars. */
-    char *argv_buf = malloc(sizeof(char) * argv_buf_size);
-    if (*argv == NULL || argv_buf == NULL) {
+    // Add 1 for the NULL pointer to mark the end, and check for overflow.
+    size_t num_ptrs = new_argc + 1;
+    if (num_ptrs == 0) {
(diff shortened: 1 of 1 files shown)


sunfishcode:
@tschneidereit There's a Windows CI failure affecting this and other PRs which seems unrelated to the actual PRs; do you know what might be causing this?

The error message is:

```
error: could not remove 'setup' file: 'C:\Users\VssAdministrator\.cargo\bin/rustup-init.exe'
info: caused by: Access is denied. (os error 5)
```

alexcrichton:
@sunfishcode this is unfortunately a [known bug](https://github.com/microsoft/azure-pipelines-image-generation/issues/1224), but you can fix it by passing `--no-self-update` to `rustup update`

sunfishcode:
Thanks! The workaround is now in, and the CI is passing.

libc-bottom-half/crt/crt1.c
```diff
@@ -13,29 +13,47 @@ void _Exit(int) __attribute__((noreturn));
...
     }
 
-    /* Allocate memory for the array of pointers, adding null terminator. */
-    *argv = malloc(sizeof(char *) * (*argc + 1));
-    /* Allocate memory for storing the argument chars. */
-    char *argv_buf = malloc(sizeof(char) * argv_buf_size);
-    if (*argv == NULL || argv_buf == NULL) {
+    // Add 1 for the NULL pointer to mark the end, and check for overflow.
+    size_t num_ptrs = new_argc + 1;
+    if (num_ptrs == 0) {
```
sbc100: I don't see how this can be zero  since you just checked that new_argc is not zero.

libc-bottom-half/crt/crt1.c
```diff
@@ -13,29 +13,47 @@ void _Exit(int) __attribute__((noreturn));
...
         return __WASI_ENOMEM;
     }
 
-    /* Make sure the last pointer in the array is NULL. */
-    *argv[*argc] = NULL;
+    // Allocate memory for storing the argument chars.
+    char *argv_buf = malloc(argv_buf_size);
```
sbc100: In emscripten I ended up inlining this function into _start itself so that these could become `alloca` and simple programs could avoid depending malloc + free.  Not related to this change but worth mentioning for later maybe?

libc-bottom-half/crt/crt1.c
sunfishcode: It would be zero if the `+ 1` overflows (per the comment). Would it be more clear if this code used `__builtin_add_overflow` instead?

sbc100: Oh yes, forgot about the pesky people passing 2^32 command line arguments :)

libc-bottom-half/crt/crt1.c
sunfishcode: That's a good idea. Once the libpreopen reorg PR lands, and the next PR to make it so that the libpreopen code doesn't get linked in if you don't use it, then this will be the only `malloc` left, so we should remove it. I'll submit a separate PR for that.

PR #121, opened 2019-10-25
Allocate the argv buffers with alloca instead of malloc.

sunfishcode:
This way, if an application doesn't otherwise use malloc, they don't
need to link in malloc.

Idea from https://github.com/CraneStation/wasi-libc/pull/118#discussion_r337596386!

One thing I'm somewhat concerned by is the lack of error reporting in `alloca`. Perhaps that's a broader problem that we'll want to address anyway?


sunfishcode:
I am considering waiting until `__stack_limit` is available ([patch here](https://reviews.llvm.org/D70515)), followed by a way to check for stack overflow in clang, before enabling this.

PR #123, opened 2019-10-25
Remove more unsupported headers.

sunfishcode:
utime.h, sysmacros.h, and libintl.h are all currently unsupported.
Removing them helps programs that autodetect features based on the
existence of headers.


sunfishcode:
The CI failure is because this turns out to depend on https://github.com/CraneStation/wasi-libc/pull/122.

PR #124, opened 2019-10-25
Avoid using user identifiers in function declarations.

sunfishcode:
MultiSource/Benchmarks/McCat in llvm-test-suite has a macro named
`n`, so rename function parameters to avoid colliding.


sbc100:
`#define n ..`   this is fine...

PR #126, opened 2019-10-25
Avoid varargs conventions when calling open

sunfishcode:
Varargs calls take more code space, and for `open`, aren't strictly necessary. Marking as a draft, as this depends on https://github.com/CraneStation/wasi-libc/pull/110.


sunfishcode:
#110 has now landed, and this is now rebased, so this is ready for review!

PR #127, opened 2019-10-25
Don't link in libpreopen initialization code when it isn't needed.

sunfishcode:
This uses similar weak-symbol tricks are the environment-variable initialization code. Marking as a draft, as this depends on https://github.com/CraneStation/wasi-libc/pull/126.

With this and previous PRs, a simple `int main(void)` function that just returns a value only imports `proc_exit`, as environment variables, command-line arguments, and libpreopen are now only pulled in when they're actually needed.

### DIFF
--- a/libc-bottom-half/crt/crt1.c
+++ b/libc-bottom-half/crt/crt1.c
@@ -9,50 +9,14 @@ extern void __wasm_call_ctors(void);
 extern int __original_main(void);
 extern void __prepare_for_exit(void);
 void _Exit(int) __attribute__((noreturn));
-
-static __wasi_errno_t populate_libpreopen(void) {
-    // Skip stdin, stdout, and stderr, and count up until we reach an invalid
-    // file descriptor.
-    for (__wasi_fd_t fd = 3; fd != 0; ++fd) {
-        __wasi_prestat_t prestat;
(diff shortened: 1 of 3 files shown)


sunfishcode:
#126 is now merged, so this is ready for review!

sunfishcode:
This is a very simple patch which makes a big impact on code size for small programs. Anyone want to review?

libc-bottom-half/crt/crt1.c
```diff
@@ -9,50 +9,14 @@ extern void __wasm_call_ctors(void);
...
-    return __WASI_ESUCCESS;
-}
+__wasi_errno_t __wasilibc_populate_libpreopen(void) __attribute__((weak));
 
 void _start(void) {
-    // Record the preopened resources.
-    if (populate_libpreopen() != __WASI_ESUCCESS) {
-        _Exit(EX_OSERR);
+    // Record the preopened resources, if needed.
+    if (&__wasilibc_populate_libpreopen != NULL) {
```
michaelvoronov: Maybe one if? As for me, it looks better if there indentation level as less as possible (within reasonable limits of cause :) ).

sunfishcode: Done!

PR #128, opened 2019-11-05
Avoid a `strdup` call in `__wasilibc_populate_libpreopen`.

sunfishcode:
Optimize `__wasilibc_populate_libpreopen` to avoid calling `strdup` in
the common case where it's called from `__wasilibc_populate_libpreopen`.

This PR depends on #127.

### DIFF
--- a/libc-bottom-half/libpreopen/libpreopen.c
+++ b/libc-bottom-half/libpreopen/libpreopen.c
@@ -498,11 +498,7 @@ int
 __wasilibc_register_preopened_fd(int fd, const char *path)
 {
     const char *name = strdup(path);
-    if (name == NULL) {
-        return -1;
-    }
-
-    return __wasilibc_register_preopened_fd(fd, name);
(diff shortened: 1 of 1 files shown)


sunfishcode:
#127 is now merged, so now this patch is ready for review!

libc-bottom-half/libpreopen/libpreopen.c
```diff
@@ -494,6 +491,20 @@ __wasilibc_register_preopened_fd(int fd, const char *path)
...
 }
 
+/// Register the given pre-opened file descriptor under the given path.
+///
+/// This function does not take ownership of `path`.
+int
+__wasilibc_register_preopened_fd(int fd, const char *path)
+{
+    const char *name = strdup(path);
+    if (name == NULL) {
```
michaelvoronov: `return name == NULL ? -1 : __wasilibc_register_preopened_fd(fd, name);` if it convenient with the coding style :).

sunfishcode: Done!

PR #129, opened 2019-11-05
 Ignore rights in libpreopen

sunfishcode:
Don't ignore paths which don't have the required rights. This means
that if the lookup finds a path that doesn't have the required
rights, it'll just proceed to the actual operation which will fail
with `ENOTCAPABLE`.

Intuitively, use cases which would depend on having multiple
overlapping matching paths for a given lookup and intelligently
picking the one with the required rights seems like they should
be uncommon.

This is simpler overall, and requires less code.

This depends on #128.

### DIFF
--- a/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
+++ b/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
@@ -8,15 +8,11 @@ extern "C" {
 /**
  * Look up the given path in the preopened directory map. If a suitable
  * entry is found, return its directory file descriptor, and store the
- * computed relative path in *relative_path. Ignore preopened directories
- * which don't provide the specified rights.
+ * computed relative path in *relative_path.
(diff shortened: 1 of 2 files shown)


sunfishcode:
#128 is now merged, so this is now ready for review. The code change is straight-forward, it's mostly just deleting stuff. The interesting question is, is ignoring rights the right thing to do when determining which path to look up?

I think it is. If we do consult the rights, that would mean that one path could mean two different things depending only on one kind of operation it's used in, which seems surprising.

Assuming that's reasonable, this is a win-win-win -- simpler code, smaller code size, and simpler path resolution logic.

sunfishcode:
Anyone interesting in reviewing this? The code change is simple at least -- it's just deleting code.

If anyone has any questions about what this means for rights, feel free to ask!

sunfishcode:
Thanks!

PR #134, opened 2019-11-09
Make __wasi_linkcount_t a uint64_t

cjihrig:
Refs: https://github.com/WebAssembly/WASI/pull/127

### DIFF
--- a/libc-bottom-half/headers/public/wasi/core.h
+++ b/libc-bottom-half/headers/public/wasi/core.h
@@ -170,7 +170,7 @@ typedef uint16_t __wasi_fstflags_t;
 
 typedef uint64_t __wasi_inode_t;
 
-typedef uint32_t __wasi_linkcount_t;
+typedef uint64_t __wasi_linkcount_t;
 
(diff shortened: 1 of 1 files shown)


cjihrig:
Ping @sunfishcode?

sunfishcode:
We need to figure out the right way to coordinate this with the overall switch to `wasi_snapshot_preview0`. Would it make sense to have a branch that we can make all the changes on, and then merge them all into master at once?

cjihrig:
That sounds like a good approach. That would include this PR, https://github.com/CraneStation/wasi-libc/pull/106, and something for https://github.com/WebAssembly/WASI/pull/125 (I was going to PR that as well, but wanted to see how this PR went first). I *think* those are all of the changes for the current snapshot.

sunfishcode:
Ok, I've now created https://github.com/CraneStation/wasi-libc/tree/wasi_snapshot_preview1 so we can do PRs there.

cjihrig:
Thanks. I've updated this PR to target that branch.

PR #136, opened 2019-11-19
Generate the WASI interface from witx.

sunfishcode:
This replaces the hand-maintained <wasi/core.h> header with a
<wasi/api.h> generated from witx.

Most of the churn here is caused by upstream WASI renamings; hopefully
in the future ABI updates will be less noisy.

@cjihrig This obviates https://github.com/CraneStation/wasi-libc/pull/135, and I hope it'll be a more robust approach as the API continues to evolve.

@pchickey


alexcrichton:
Would it make sense to include a submodule and/or reference to the wasi repo, and then check that the generated code matches the reference on CI? 

(to ensure that if the reference is changed and/or the generator the generated code is updated as well)

libc-bottom-half/headers/public/wasi/api.h
```diff
@@ -0,0 +1,2569 @@
+/**
+ * THIS FILE IS AUTO-GENERATED!
```
sbc100: By whom?  Where I go if I want to modify it?

sunfishcode: I've now added a patch to add more information here.

sunfishcode:
> I'm not totally convinced the tool for generating the header should live in this repo. Although I don't have any great ideas for a better place for it. At least maybe we can discuses that separately? How would you feel about splitting this change such that tool is a separate PR that we can discuses. The actual libc change LGTM as is so no need to block that.

I agree, but I also don't have great ideas for a better place for it in its current form. It's specific to wasi-libc, and still pretty experimental. What would you think of leaving it here for now, but re-evaluating it after it's settled into its role a little?

sbc100:
Isn't it C-specific rather then wasi-libc specific?  i.e. wasi-libc is just one way to build something useful on C on top of this C header file?   But I guess this is will at least the primary consumer. 

I can maybe see this living in `WebAssembly/wasi` .. on can imagine generators for many different languages in there, as we already have a president of rust tooling in that repo.

sunfishcode:
The <wasi/api.h> header is hard-wired into the tool right now, plus `__WASI_` names, and some workarounds for WASI-specific issues. Witx will also actively evolving and will require this tool to change to better support eg. modularity, which will also require coordination with libc code. In the future it'll be a more general tool, but there's work to do before we get there.

pchickey:
I agree that this repo seems like the best place for this tool to live for now. We will want a bit of flexibility about merging changes into the `witx` crate in the `WebAssembly/wasi` repo, versus consuming those changes over here.

sunfishcode:
@alexcrichton @pchickey git submodules are pinned to a specific revision, so I'm unclear on the purpose of setting them up with scripts to check that the generated files are unchanged.

sbc100:
I guess the idea would be that you would roll the wasi submodule and in the same PR see the updated headers files and also ensure the header generator is in sync the the witx parser..  although actually it looks like you are pinning that dependency via crates.io and via submodule.

sbc100:
I'm not gonna block this PR if you guys think this is the right place for it.  We can revisit later if needed.  

I'm not huge fan of having rust tooling spread everywhere like this (mostly because its not my personal lingua fanca, but also because its also not the lingua fanca of C toolchain maintainers in general).  But I'm not offering any alternatives right now.

alexcrichton:
@sunfishcode I was thinking of something like https://github.com/bytecodealliance/wasi/tree/master/crates/generate-raw where with a submodule here we'd get benefits like:

* If you're only changing the generator, a PR shows that
* If all you're doing is updating the submodule, a PR shows that
* It's always clear where the header file is being generated from
* CI can be used to ensure that the generated code is, well, always generated

This isn't really critical, though, just something I've seen that's a good way to head off future headaches. Not really a blocker at all though. I haven't though too much how this fits into the general versioning story for wasi in general (and its evolving interfaces over time).

sunfishcode:
@alexcrichton That's a good idea. I've now added a WASI submodule, which isn't needed unless you want to re-generate the header, and running `cargo run` auto-regenerates the header`.

alexcrichton:
👍 

The only other thing I'd do here is to hook this up to CI to fail CI if the generated code in the repository is different than what the generator generates, but that's pretty minor.

tools/wasi-headers/Cargo.toml
```diff
@@ -7,6 +7,6 @@ edition = "2018"
 publish = false
 
 [dependencies]
-clap = "2"
 heck = "0.3.1"
 witx = "0.5.0"
```
pchickey: A path dependency to witx is appropriate now that we're tracking WASI as a submodule, imo.

sunfishcode: Makes sense; done.

sunfishcode:
The CI is now passing, at the cost of making git checkout text files with unix-style line endings on Windows. It would also be possible to make the diff ignore line-ending differences, but with text=lf, we don't need that, we avoid trouble with people running the generator tool on Windows and getting crlf/lf diffs, and we eliminate some line-ending differences in the generated sysroot too.

PR #140, opened 2019-11-22
Wasi snapshot preview1

sunfishcode:
`wasi_snapshot_preview1` is now ready to go! 

Note that in this update, `<wasi/core.h>` is renamed to `<wasi/api.h>`, and many of the constants have different names as well, as they're now auto-generated from witx.


sunfishcode:
Merging, as there are no new changes here; this is just moving the changes from a branch to master.

PR #142, opened 2019-11-25
Use constructor functions for optional init routines.

sunfishcode:
Instead of using weak symbols, use constructor function attributes for the
environment and preopen initialization routines. This is simpler, uses
less code, and is more LTO-friendly.


kripken:
Why is this more LTO friendly?

sunfishcode:
It means we don't have to worry about presenting the objects to the linker in a specific order in order to get it to make the correct choice of which definition to use. With constructor functions, there's no choice to make.

libc-bottom-half/sources/__environ.c
```diff
@@ -1,59 +1,63 @@
...
 extern __typeof(__environ) _environ __attribute__((weak, alias("__environ")));
 extern __typeof(__environ) environ __attribute__((weak, alias("__environ")));
 
-// This function is referenced by a weak symbol in crt1.c, and we define
-// it here in the same source file as __environ, so that this function is
-// linked in iff environment variable support is used.
-__wasi_errno_t __wasilibc_populate_environ(void) {
+// We define this function here in the same source file as __environ, so that
+// this function is called in iff environment variable support is used.
+__attribute__((constructor(0)))
```
sbc100: I'm not sure about the 0 here.  Presumably we want to reserve some number range for even lower level stuff?  (such as stuff generated by the compiler.. although I guess the linker can put any self-generated stuff before 0 if it chooses).

sunfishcode:
We can discuss whether crt1.o should be a bitcode file when I submit the PR for that ;-).

libc-bottom-half/sources/__environ.c
sunfishcode: Makes sense. Offhand it's hard to predict what we'd want. The reserved range is 0-100, so I'll set it to 50 to leave room on both sides.

PR #149, opened 2019-12-19
gen-headers: Generate assertions of layout from witx

pchickey:
Serves as evidence that https://github.com/WebAssembly/WASI/pull/181 is correct

### DIFF
--- a/libc-bottom-half/headers/public/wasi/api.h
+++ b/libc-bottom-half/headers/public/wasi/api.h
@@ -500,7 +500,7 @@ typedef uint64_t __wasi_rights_t;
 /**
  * The right to invoke `fd_datasync`.
  * If `path_open` is set, includes the right to invoke
- * `path_open` with `fdflag::dsync`.
+ * `path_open` with `fdflags::dsync`.
  */
 #define __WASI_RIGHTS_FD_DATASYNC (UINT64_C(1))
 
@@ -523,13 +523,13 @@ typedef uint64_t __wasi_rights_t;
 /**
  * The right to invoke `fd_sync`.
  * If `path_open` is set, includes the right to invoke
- * `path_open` with `fdflag::rsync` and `fdflag::dsync`.
+ * `path_open` with `fdflags::rsync` and `fdflags::dsync`.
  */
 #define __WASI_RIGHTS_FD_SYNC (UINT64_C(16))
 
 /**
  * The right to invoke `fd_seek` in such a way that the file offset
- * remains unaltered (i.e., `WHENCE_CUR` with offset zero), or to
+ * remains unaltered (i.e., `whence::cur` with offset zero), or to
  * invoke `fd_tell`.
  */
 #define __WASI_RIGHTS_FD_TELL (UINT64_C(32))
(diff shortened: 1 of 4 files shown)


pchickey:
The CI doesn't actually build the libc. I tested locally with wasi-sdk 7:

`/opt/wasi-sdk/bin/clang -c libc-bottom-half/headers/public/wasi/api.h`

libc-bottom-half/headers/public/wasi/api.h
```diff
@@ -40,16 +41,22 @@ extern "C" {
 #define __WASI_DIRCOOKIE_START (UINT64_C(0))
 typedef __SIZE_TYPE__ __wasi_size_t;
 
+_Static_assert(sizeof(__wasi_size_t) == 4, "witx calculated size");
+_Static_assert(_Alignof(__wasi_size_t) == 4, "witx calculated align");
```
sunfishcode: Purely for tidiness reasons, it'd be nice to have a newline between these asserts and the subsequent comment.

tools/wasi-headers/src/c_header.rs
```diff
@@ -38,6 +38,7 @@ _Static_assert(_Alignof(int32_t) == 4, "non-wasi data layout");
 _Static_assert(_Alignof(uint32_t) == 4, "non-wasi data layout");
 _Static_assert(_Alignof(int64_t) == 8, "non-wasi data layout");
 _Static_assert(_Alignof(uint64_t) == 8, "non-wasi data layout");
+_Static_assert(_Alignof(void*) == 4, "non-wasi data layout");
```
sunfishcode: It's worth adding a comment somewhere around here mentioning that some of these asserts will need to be generalized when wasm64 happens.

pchickey: Agreed, all of the layout calculations will need to be generalized when that happens, but I'll add a comment so nobody misses it if they're consuming this file.

libc-bottom-half/headers/public/wasi/api.h
pchickey: Fixed

pchickey: Fixed in 475a62

libc-bottom-half/headers/public/wasi/api.h
```diff
@@ -31,6 +31,9 @@ _Static_assert(_Alignof(int32_t) == 4, "non-wasi data layout");
 _Static_assert(_Alignof(uint32_t) == 4, "non-wasi data layout");
 _Static_assert(_Alignof(int64_t) == 8, "non-wasi data layout");
 _Static_assert(_Alignof(uint64_t) == 8, "non-wasi data layout");
+// This header is only valid for wasm32: witx currently calcualtes memory
+// layout for the wasm32 abi only.
```
sbc100: No need to block this PR on this but maybe this is better of at the very top of the header and can be even more explicit. Alongside the `ifndef __wasi__` you could add `ifndef __wasm32__.. this header is only meant for wasm32`

sbc100: Then no comment needed just a `#error`

sunfishcode: For maximal clarity, I suggest phrasing this as "doesn't yet support wasm64" so that it's clear that this is just the present status, not a fundamental limitation.

pchickey: Implemented these suggestions in af28fd5

pchickey:
Well, apparently I forgot to merge this right after returning for the new year, and it got out-of-date. I just rebased on master, pointed at the current wasi master, and regenerated header.

PR #150, opened 2019-12-20
Support whole-program LTO

sunfishcode:
Build and install an LTO version of libc.a and the startup files,
alongside the non-LTO versions.

This also adds support for the proposed `__main_argc_argv` convention,
supporting compilers both with and without that change.

This doesn't depend on https://reviews.llvm.org/D70700, but contains code that anticipates it.

This does depend on https://github.com/CraneStation/wasi-sdk/pull/89.


sbc100:
BTW, I'm working an wasm-ld fix that will allow compiler-to be part of LTO too: https://reviews.llvm.org/D71738  :)

sunfishcode:
> Is it possible to split the renaming of the math sources into a separate PR.

Done, in #151.

> Also, it looks like a lot of the Makefile refactoring could be split out? I can't tell which parts of that relate to this change directly.

I've now split out the main code changes in #15 

Assuming those land, when I rebase this PR over them this PR will be just a Makefile change.

sunfishcode:
The dependencies for this PR all landed, and this PR is now rebased, so it's now just changes to the top-level Makefile to support LTO, and ready for review!

Makefile
```diff
@@ -9,13 +9,17 @@ SYSROOT ?= $(CURDIR)/sysroot
 # A directory to install to for "make install".
 INSTALL_DIR ?= /usr/local
 # single or posix
-THREAD_MODEL = single
```
sbc100: There still seem to be a lot of non-LTO related changes here?   Some kind of re-base issue maybe?

sunfishcode: I've now filed #186 to pull out several refactorings.

yamt:
@sunfishcode are you still interested in this? or may i (or someone) take it over?

TerrorJack:
To avoid duplicate effort: you can apply https://gitlab.haskell.org/ghc/wasi-libc/-/commit/eecc0942af95a90447e6bad8c58d8c133ec82fa2 to `wasi-libc`, then disable the predefined macro/symbol check, after which you can enable either thin/full LTO for wasi-libc as well as compiler-rt/libc++.

Distributing different sysroots both with/without LTO is harder though.

sunfishcode:
@yamt I'm not actively working on this, so you are welcome to take this over!

Makefile
```diff
@@ -158,53 +153,36 @@ LIBC_TOP_HALF_MUSL_SOURCES = \
...
+override CRT_SOURCES = $(LIBC_BOTTOM_HALF_CRT_SOURCES)
 endif
-ifeq ($(THREAD_MODEL), posix)
-override WASM_CFLAGS += -mthread-model posix -pthread
+ifeq ($(LTO),yes)
+# The following files define functions which are called by LLVM CodeGen,
+# which runs after LLVM LTO, so LTO libraries don't satisfy them. They're
+# also relatively uninteresting to LTO, as LLVM recognizes most of them
+# even without seeing their definitions.
+override LIBC_NONLTO_SOURCES = \
```
yamt, line 167: @sunfishcode 
@sbc100 
this exclusion is no longer necessary after https://reviews.llvm.org/D71738, right?

yamt:
> To avoid duplicate effort: you can apply https://gitlab.haskell.org/ghc/wasi-libc/-/commit/eecc0942af95a90447e6bad8c58d8c133ec82fa2 to `wasi-libc`, then disable the predefined macro/symbol check, after which you can enable either thin/full LTO for wasi-libc as well as compiler-rt/libc++.
> 
> Distributing different sysroots both with/without LTO is harder though.

thank you. i guess i will look at full lto first though.

as thin lto seems to have more obstacles.
eg.
https://github.com/llvm/llvm-project/issues/91711
https://github.com/llvm/llvm-project/issues/91700

PR #151, opened 2019-12-20
Move math source files.

sunfishcode:
This is split out from the LTO change. It's needed by the LTO build so
that we can easily build these files in non-LTO mode, since they satisfy
calls emitted by LLVM codegen after LTO.


sbc100:
This should be unnecessary real soon now BTW: https://reviews.llvm.org/D71738

sunfishcode:
That is super cool. I'm inclined to keep the `math` directory here as it's a nice little bit of organization, but we can simplify some things in the LTO build once that's available.

PR #152, opened 2019-12-20
Add support for `__main_argc_argv`.

sunfishcode:
This adds support for the `__main_argc_argv` change, while preserving
compatibility with `__original_main`. This is needed by the LTO build
because the `__original_main` hack works in LLVM codegen, which is after
LTO. The `__main_argc_argv` change is implemented in clang, which makes
it properly visible to LTO.


sunfishcode:
With https://github.com/WebAssembly/tool-conventions/pull/134 and https://reviews.llvm.org/D70700 merged, it's time to merge this too! 🏄

Issue #116, opened 2019-10-21
Style guide


michaelvoronov:
It would good to have a explicit style guide for contributors to this repo (and maybe other in Cranelift). Because there are some divergences between different files (like comment style [here](https://github.com/CraneStation/wasi-libc/blob/master/libc-bottom-half/crt/crt1.c#L108) and [here](https://github.com/CraneStation/wasi-libc/blob/master/libc-bottom-half/mman/mman.c#L61)) and sometimes it hard to make code follow the style guide.

sunfishcode:
wasi-libc includes codes from several external sources, with differing style conventions, so when adding new code, we generally just follow the style of the nearest code.

That said, I agree we could be more consistent in our own code. I've lately been moving wasi-libc's own .c files from `/*` to `//` comments (https://github.com/CraneStation/wasi-libc/pull/112, https://github.com/CraneStation/wasi-libc/pull/110, https://github.com/CraneStation/wasi-libc/pull/109); once those land we could do a follow-up to convert the few remaining places.

PR #153, opened 2019-12-20
Convert more wasi-libc code to `//`-style comments.

sunfishcode:
This is purely a style change, in accordance with #116.


sunfishcode:
@michaelvoronov This is another followup to #116; it's still not a style guide, but it's moving more things to having a consistent style.

PR #154, opened 2019-12-21
build: add cmake based build infrastructure

compnerd:
Rather than rely on make, use CMake to generate the rules in the build
system of choice.  This allows building with `make` or `ninja`, and
enables building the WASI libc on Windows.

```
cmake -G Ninja -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_C_COMPILER=clang -DCMAKE_SYSTEM_NAME=wasi -DCMAKE_MODULE_PATH=cmake/Modules
ninja -C build install
```


compnerd:
Moved from WebAssembly/wasi-libc#12

compnerd:
@sunfishcode I've moved the PR here as per your comments.  I also opted to just dump a first cut attempt to enumerate the sources and add support for the new content.  This should be able to generate the additional libraries added here.

This abstracts the build system, allowing building with ninja or make.

CMakeLists.txt
```diff
@@ -0,0 +1,1132 @@
...
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__typedef_nfds_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__typedef_in_port_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__typedef_in_addr_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__typedef_fd_set.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__typedef_DIR.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__typedef_clockid_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/time.h ${CMAKE_BINARY_DIR}/sysroot/include/
```
compnerd: This collides with basics/include/time.h

CMakeLists.txt
```diff
@@ -0,0 +1,1132 @@
...
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__struct_pollfd.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__struct_msghdr.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__struct_iovec.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__struct_in_addr.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__struct_in6_addr.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__struct_dirent.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/string.h ${CMAKE_BINARY_DIR}/sysroot/include/
```
compnerd: This collides with basics/headers/string.h

CMakeLists.txt
```diff
@@ -0,0 +1,1132 @@
...
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/stdlib.h ${CMAKE_BINARY_DIR}/sysroot/include/
```
compnerd: This collides with basics/include/stdlib.h

CMakeLists.txt
```diff
@@ -0,0 +1,1132 @@
...
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/string.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/stdlib.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__macro_PAGESIZE.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/inttypes.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__header_inttypes.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__functions_memcpy.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__functions_malloc.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/errno.h ${CMAKE_BINARY_DIR}/sysroot/include/
```
compnerd: This collides with libc-bottom-half/headers/public/errno.h

CMakeLists.txt
```diff
@@ -0,0 +1,1132 @@
...
+
+option(ENABLE_THREADS "enable threads" OFF)
+option(BUILD_DLMALLOC "build dlmalloc" YES)
+cmake_dependent_option(BUILD_LIBC_BOTTOM_HALF "build libc bottom half" YES
+  "BUILD_DLMALLOC" YES)
+cmake_dependent_option(BUILD_LIBC_TOP_HALF "build libc top half" YES
+  "BUILD_LIBC_BOTTOM_HALF" YES)
+
+if(CMAKE_C_COMPILER_ID MATCHES Clang)
+  add_compile_options(-target wasm32-unknown-wasi)
```
sunfishcode: We generally abbreviate the target as "wasm32-wasi".

cmake/Modules/Platform/wasi.cmake
```diff
@@ -0,0 +1,3 @@
+set(WASI 1)
+set(UNIX 1)
```
sunfishcode: WASI isn't "Unix" -- we don't predefine `__unix__` and so forth. For code in libc, it's better to just teach the code how to handle WASI than to pretend to be Unix.

CMakeLists.txt
```diff
@@ -0,0 +1,1132 @@
...
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__header_bits_signal.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__function___isatty.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__fd_set.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/fcntl.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__errno_values.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/errno.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/dirent.h ${CMAKE_BINARY_DIR}/sysroot/include/
```
compnerd: Collides with libc-top-half/include/dirent.h

CMakeLists.txt
```diff
@@ -0,0 +1,1132 @@
...
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__header_dirent.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__header_bits_signal.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__function___isatty.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__fd_set.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/fcntl.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__errno_values.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/errno.h ${CMAKE_BINARY_DIR}/sysroot/include/
```
compnerd: Collides with libc-top-half/include/errno.h

CMakeLists.txt
```diff
@@ -0,0 +1,1132 @@
...
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__header_netinet_in.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__header_fcntl.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__header_dirent.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__header_bits_signal.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__function___isatty.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__fd_set.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/fcntl.h ${CMAKE_BINARY_DIR}/sysroot/include/
```
compnerd: Collides with libc-top-half/include/fcntl.h

CMakeLists.txt
```diff
@@ -0,0 +1,1132 @@
...
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__struct_in_addr.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__struct_in6_addr.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__struct_dirent.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/string.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/stdlib.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/__seek.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/libc-bottom-half/headers/public/poll.h ${CMAKE_BINARY_DIR}/sysroot/include/
```
compnerd: Collides with libc-top-half/include/poll.h

cmake/Modules/Platform/wasi.cmake
compnerd: This is not for defining `__unix__`, this it to correct the handling of library generation inside of CMake itself (it makes assumptions about library prefixes, naming, etc based on this variable).

sunfishcode: Ok, makes sense. Would you mind adding this in a comment in the code?

compnerd: Absolutely.

CMakeLists.txt
compnerd: Ah, internally, LLVM will actually normalize the triple to this.  Is there a reason to not just use the normalized form?  If its just to maintain uniformity with Make (which I think longer term could be replaced with the metabuild), that is reasonable.

compnerd:
I'll take a look at #150, thanks for the heads up.  I think most of the metabuild systems do the same.  But, yes, it should give proper dependency tracking/incremental builds and better cross-platform support.  I'm not 100% tied to CMake, if you prefer gn, I could swap to that.

I think that with CMake, you tend to just build in a different tree with a different set of flags rather than the different variants in one.  gn does the same as well, with the difference that you just alter the configuration flags in the same tree.

cmake/Modules/Platform/wasi.cmake
```diff
@@ -0,0 +1,9 @@
+set(WASI 1)
+# NOTE: this is being defined *not* because WASI is unix-esque, which it is not.
```
sbc100: Is this file needed/used?

sbc100:
Can you update the .github actions to use cmake and delete the old Makefile so that we can see how this works in action?

CMakeLists.txt
```diff
@@ -0,0 +1,1109 @@
...
+  LANGUAGES ASM C)
+
+include(CMakeDependentOption)
+
+option(ENABLE_THREADS "enable threads" OFF)
+option(BUILD_DLMALLOC "build dlmalloc" YES)
+cmake_dependent_option(BUILD_LIBC_BOTTOM_HALF "build libc bottom half" YES
+  "BUILD_DLMALLOC" YES)
+cmake_dependent_option(BUILD_LIBC_TOP_HALF "build libc top half" YES
+  "BUILD_LIBC_BOTTOM_HALF" YES)
```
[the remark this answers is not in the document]
compnerd: I honestly do not know.  I replicated the options because they exist in the make system.  If they are no longer useful, I would really prefer to remove the options.  The fewer options there are, the better it is.

CMakeLists.txt
```diff
@@ -0,0 +1,1109 @@
...
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__typedef_ino_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__typedef_mode_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__typedef_nlink_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__typedef_off_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__typedef_ssize_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__typedef_suseconds_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__typedef_time_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different ${PROJECT_SOURCE_DIR}/basics/include/__typedef_uid_t.h ${CMAKE_BINARY_DIR}/sysroot/include/
```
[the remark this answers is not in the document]
compnerd: You need to enumerate the files, yes, I can create a list of the files and use a loop, but, it will make things a bit more complicated due to the dependency tracking.

cmake/Modules/Platform/wasi.cmake
compnerd: Yes, it is needed, and used.  Without this, this controls the behaviour of the build.  See the example command in the commit message.

compnerd:
I'm not very familiar with the GitHub actions, I'll see if I can figure out how to do that.  I think that it may be useful to do that in two stages.

pchickey:
This doesn't appear to be working on my machine. I had to install the latest cmake release, but that doesn't appear to be sufficient. I do think we need to get it working on CI before merging.

To add the cmake and ninja (or make) steps to the Github CI, add each command as a `- run` directive below this line:

https://github.com/CraneStation/wasi-libc/blob/2f6ce572c6a1af48b7392bffddb81b94411c4478/.github/workflows/main.yml#L28

Here is the build failure on my machine:

```sh
[pat@finknottle:src/wasi-libc]% /usr/local/bin/cmake -G Ninja -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_C_COMPILER=clang -DCMAKE_SYSTEM_NAME=wasi -DCMAKE_MODULE_PATH=cmake/Modules  
-- The ASM compiler identification is Clang
-- Found assembler: /usr/bin/clang
-- The C compiler identification is Clang 7.1.0
-- Check for working C compiler: /usr/bin/clang
System is unknown to cmake, create:
Platform/wasi to use this system, please post your config file on discourse.cmake.org so it can be added to cmake
-- Check for working C compiler: /usr/bin/clang -- works
-- Detecting C compiler ABI info
System is unknown to cmake, create:
Platform/wasi to use this system, please post your config file on discourse.cmake.org so it can be added to cmake
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /home/pat/src/wasi-sdk/src/wasi-libc/build
[pat@finknottle:src/wasi-libc]% ninja -C build
ninja: Entering directory `build'
[6/765] Building C object CMakeFiles/libc.dir/basics/sources/complex-builtins.c.o
FAILED: /usr/bin/clang   -O2 -g -DNDEBUG   -target wasm32-unknown-wasi --sysroot=/home/pat/src/wasi-sdk/src/wasi-libc/build/sysroot -fno-trapping-math -mthread-model single -MD -MT CMakeFiles/libc.dir/basics/sources/complex-builtins.c.o -MF CMakeFiles/libc.dir/basics/sources/complex-builtins.c.o.d -o CMakeFiles/libc.dir/basics/sources/complex-builtins.c.o   -c ../basics/sources/complex-builtins.c
../basics/sources/complex-builtins.c:5:10: fatal error: 'complex.h' file not found
#include <complex.h>
         ^~~~~~~~~~~
1 error generated.
[6/765] Building C object CMakeFiles/libc.dir/basics/sources/reallocarray.c.o
FAILED: /usr/bin/clang   -O2 -g -DNDEBUG   -target wasm32-unknown-wasi --sysroot=/home/pat/src/wasi-sdk/src/wasi-libc/build/sysroot -fno-trapping-math -mthread-model single -MD -MT CMakeFiles/libc.dir/basics/sources/reallocarray.c.o -MF CMakeFiles/libc.dir/basics/sources/reallocarray.c.o.d -o CMakeFiles/libc.dir/basics/sources/reallocarray.c.o   -c ../basics/sources/reallocarray.c
../basics/sources/reallocarray.c:1:10: fatal error: 'stdlib.h' file not found
#include <stdlib.h>
         ^~~~~~~~~~
1 error generated.
[6/765] Building C object CMakeFiles/libc.dir/basics/sources/abort.c.o
FAILED: /usr/bin/clang   -O2 -g -DNDEBUG   -target wasm32-unknown-wasi --sysroot=/home/pat/src/wasi-sdk/src/wasi-libc/build/sysroot -fno-trapping-math -mthread-model single -MD -MT CMakeFiles/libc.dir/basics/sources/abort.c.o -MF CMakeFiles/libc.dir/basics/sources/abort.c.o.d -o CMakeFiles/libc.dir/basics/sources/abort.c.o   -c ../basics/sources/abort.c
../basics/sources/abort.c:1:10: fatal error: 'stdlib.h' file not found
#include <stdlib.h>
         ^~~~~~~~~~
1 error generated.
[6/765] Building C object CMakeFiles/libc.dir/libc-top-half/musl/src/misc/getsubopt.c.o
FAILED: /usr/bin/clang   -O2 -g -DNDEBUG   -target wasm32-unknown-wasi --sysroot=/home/pat/src/wasi-sdk/src/wasi-libc/build/sysroot -fno-trapping-math -mthread-model single -MD -MT CMakeFiles/libc.dir/libc-top-half/musl/src/misc/getsubopt.c.o -MF CMakeFiles/libc.dir/libc-top-half/musl/src/misc/getsubopt.c.o.d -o CMakeFiles/libc.dir/libc-top-half/musl/src/misc/getsubopt.c.o   -c ../libc-top-half/musl/src/misc/getsubopt.c
../libc-top-half/musl/src/misc/getsubopt.c:1:10: fatal error: 'stdlib.h' file not found
#include <stdlib.h>
         ^~~~~~~~~~
1 error generated.
[6/765] Building C object CMakeFiles/libc.dir/basics/sources/math/fmin-fmax.c.o
FAILED: /usr/bin/clang   -O2 -g -DNDEBUG   -target wasm32-unknown-wasi --sysroot=/home/pat/src/wasi-sdk/src/wasi-libc/build/sysroot -fno-trapping-math -mthread-model single -MD -MT CMakeFiles/libc.dir/basics/sources/math/fmin-fmax.c.o -MF CMakeFiles/libc.dir/basics/sources/math/fmin-fmax.c.o.d -o CMakeFiles/libc.dir/basics/sources/math/fmin-fmax.c.o   -c ../basics/sources/math/fmin-fmax.c
../basics/sources/math/fmin-fmax.c:10:10: fatal error: 'math.h' file not found
#include <math.h>
         ^~~~~~~~
1 error generated.
[6/765] Building C object CMakeFiles/dlmalloc.dir/dlmalloc/src/dlmalloc.c.o
FAILED: /usr/bin/clang  -I../dlmalloc/include -O2 -g -DNDEBUG   -target wasm32-unknown-wasi --sysroot=/home/pat/src/wasi-sdk/src/wasi-libc/build/sysroot -fno-trapping-math -mthread-model single -MD -MT CMakeFiles/dlmalloc.dir/dlmalloc/src/dlmalloc.c.o -MF CMakeFiles/dlmalloc.dir/dlmalloc/src/dlmalloc.c.o.d -o CMakeFiles/dlmalloc.dir/dlmalloc/src/dlmalloc.c.o   -c ../dlmalloc/src/dlmalloc.c
../dlmalloc/src/dlmalloc.c:7:10: fatal error: 'malloc.h' file not found
#include <malloc.h>
         ^~~~~~~~~~
1 error generated.
ninja: build stopped: subcommand failed.
```

compnerd:
I’m pretty sure that complex.h is provided by the compiler resource directory.  Is the clang install complete?

Thanks for the hints on the GitHub actions.  I’m going to try to see if I can drop the minimum version to 3.12 as the current azure image seems to have that.  Odd since I use the 3.15 on the Swift builds.

cmake/Modules/Platform/wasi.cmake
compnerd: Actually, it seems that with a bit more tweaking, I can get rid of it.

compnerd:
@pchickey would you mind trying again with this current version?  it builds locally and on azure now.

@sbc100 the .github/workflows doesn't seem to be building things, the azure pipelines is.  That too has been updated now.  I've refrained from deleting the Makefile for the time being, it would be best to actually get this merged before we do that.

CMakeLists.txt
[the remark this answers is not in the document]
sunfishcode: I added those options fairly early on with the idea that people may want to replace the "top half" with a different implementation, but reuse the "bottom half" which is where most of the WASI-specific code is, but I haven't heard of anyone actually interested in doing this, so I'd be ok removing the option.

CMakeLists.txt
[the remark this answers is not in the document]
sunfishcode: Is there a way we can factor out some of the redundancy, with a macro or so? That would make this code easier to manage.

CMakeLists.txt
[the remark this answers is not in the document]
compnerd: Okay, that sounds good to me.  I think that fewer options make things easier to maintain, Ill make a change on top of this to remove the option.

CMakeLists.txt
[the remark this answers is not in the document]
compnerd: I'll try to do something with a macro to help with this.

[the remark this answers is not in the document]
compnerd: Collapsed it into an invocation per directory.

CMakeLists.txt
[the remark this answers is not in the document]
compnerd: Removed `BUILD_DLMALLOC`, `BUILD_LIBC_BOTTOM_HALF` and `BUILD_LIBC_TOP_HALF`.

compnerd:
@sunfishcode anything else to get this moving?

CMakeLists.txt
```diff
@@ -0,0 +1,1131 @@
...
+
+add_custom_command(TARGET c POST_BUILD
+  COMMAND ${CMAKE_AR} crs ${CMAKE_BINARY_DIR}/sysroot/lib/libm.a
+  COMMAND ${CMAKE_AR} crs ${CMAKE_BINARY_DIR}/sysroot/lib/librt.a
+  COMMAND ${CMAKE_AR} crs ${CMAKE_BINARY_DIR}/sysroot/lib/libpthread.a
+  COMMAND ${CMAKE_AR} crs ${CMAKE_BINARY_DIR}/sysroot/lib/libcrypt.a
+  COMMAND ${CMAKE_AR} crs ${CMAKE_BINARY_DIR}/sysroot/lib/libutil.a
+  COMMAND ${CMAKE_AR} crs ${CMAKE_BINARY_DIR}/sysroot/lib/libxnet.a
+  COMMAND ${CMAKE_AR} crs ${CMAKE_BINARY_DIR}/sysroot/lib/libresolv.a
+  COMMAND ${CMAKE_AR} crs ${CMAKE_BINARY_DIR}/sysroot/lib/libdl.a)
```
sunfishcode: With the Makefile-based system we hit a problem where we had too many files for a Windows commandline, and there's a manual workaround that splits up the `ar` command into multiple invocations. Is that no longer a problem?

compnerd:
I didn't know that the `libc.imports` was required.  I can try to add that to the build as well.  What is the purpose of the defined symbols text files?

The LTO support in CMake roughly amounts to adding `-DINTERPROCEDURAL_OPTIMIZATION=YES`.  The idea is that rather than having all the variants be built as a single build, you build all the variants individually.  I think that the LTO would just amount to another build configuration in CI that builds with this flag.

CMakeLists.txt
compnerd: Right, that is an issue.  However, CMake is supposed to take responsibility for building up the command line appropriately.  There isnt supposed to be anything in the CMakeLists.txt to support that.  If that breaks, thats a bug in CMake, and we can deal with that separately, but it seems to build currently on Windows (as this changes the CI over to the CMake build).

sbc100: What is the advantage of supporting both at once?   It seems like a pain to support two built system without much benifit, and there is risk it will cause confusion.  I think I'd rather see this new build system replace the existing one atomically if/when it proves itself to be superior.

If you we do decide the land it alongside, then lets make it very clear that its experimental, and not make it part of the CI (to avoid blocking forward progress on maintaining dual build systems).

sunfishcode:
Yes, the libc.imports is required by wasm-ld to know which functions are expected to be left undefined. 

The other generated files, defined-symbols.txt, predefined-macros.txt, and so on, are all just sanity checks, though they're quire useful sanity checks for a libc implementation -- with all the configuration macros and symbol aliases and weak symbols floating around, it's useful to keep close track of things like the set of defined symbols visible to the outside.

tools/gen-imports.py
```diff
@@ -0,0 +1,34 @@
...
+                     '--undefined-only',
+                     '--extern-only',
+                     '--just-symbol-name',
+                     *unparsed],
+                    stdout = subprocess.PIPE)
+    output, error = process.communicate()
+
+    prefix = re.compile(args.prefix)
+    symbols = filter(lambda i: not i.endswith(':'),
+                     re.split('\r?\n', output.decode('utf-8')))
```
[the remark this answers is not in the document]
compnerd: Ah, I was trying to remember `splitlines`.  Thank you!

CMakeLists.txt
```diff
@@ -0,0 +1,1141 @@
+cmake_minimum_required(VERSION 3.12.1)
+
+project(wasi-libc
+  LANGUAGES C)
```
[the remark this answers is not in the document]
compnerd: This is to maintain consistency with all the other parameters being passed on a line on their own.  If you feel strongly, I can make it a single line.

CMakeLists.txt
```diff
@@ -0,0 +1,1141 @@
...
+    COMMAND ${CMAKE_AR} crs ${CMAKE_BINARY_DIR}/sysroot/lib/lib${stub}.a)
+  add_custom_target(${stub} ALL
+    DEPENDS ${CMAKE_BINARY_DIR}/sysroot/lib/lib${stub}.a)
+endforeach()
+
+add_custom_command(OUTPUT ${CMAKE_BINARY_DIR}/sysroot/lib/libc.imports
+  COMMAND ${Python_EXECUTABLE} ${PROJECT_SOURCE_DIR}/tools/gen-imports.py --nm ${CMAKE_NM} $<TARGET_FILE:c> $<TARGET_OBJECTS:crt1> > ${CMAKE_BINARY_DIR}/sysroot/lib/libc.imports
+  DEPENDS c crt1)
+add_custom_target(libc.imports ALL
+  DEPENDS ${CMAKE_BINARY_DIR}/sysroot/lib/libc.imports)
```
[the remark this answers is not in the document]
compnerd: Okay, Ill try to take a look at that.

tools/gen-imports.py
```diff
@@ -0,0 +1,34 @@
...
+
+    (args, unparsed) = parser.parse_known_args()
+
+    process = Popen([args.nm,
+                     '--undefined-only',
+                     '--extern-only',
+                     '--just-symbol-name',
+                     *unparsed],
+                    stdout = subprocess.PIPE)
+    output, error = process.communicate()
```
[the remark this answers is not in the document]
compnerd: Hmm, would that not assign the tuple to the output?

compnerd:
I think that this is complete now.  The only thing that it doesn't generate are two intermediate files (which could be added to `gen-imports.py` if desired).  There is a `test` target that diffs the expected macros and syntax checks the headers.  The `Makefile` has been removed as that is no longer needed for anything - everything is covered by the CMake based build.  Updated the CI configuration to build with CMake as well.

CMakeLists.txt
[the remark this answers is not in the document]
compnerd: Fair enough.  In general, it is pretty normal for `find_package` to always be on a single line.  The `project` declaration can end up having additional parameters, and it becomes easier visually to identify what has changed.  Additionally, once you have version information in there, it is nicer to have it be split up that way for the purposes of seeing when the diff just changes the version number.

compnerd:
I don't really see this is being any more complex, but rather less complex because I can immediately see what files are going where.  With the glob based make, it was very unclear and the order of the commands were particularly important but invisible as they copied files over one another.  It also makes null-builds simpler (see #156).  This not only brings that transparency, but also enables building with the GNU tools, building with Ninja, and the possibility of using other build systems more easily (e.g. gn).  Yes, the change in dependency, IMO, is massively an improvement given that all the dependencies to build this project are now available through a standard installation of Visual Studio on Windows (Python, Ninja, CMake, and clang/LLVM) are distributed through Visual Studio's installer.

This also makes it easier for me to integrate with the project for another larger scoped project of trying to get WASM/WASI support in Swift .

CMakeLists.txt
```diff
@@ -0,0 +1,1178 @@
...
+  -Wno-dangling-else
+  -Wno-unknown-pragmas)
+add_dependencies(libc-top-half sysroot-headers)
+
+target_link_libraries(c PRIVATE
+  libc-bottom-half
+  libc-top-half
+  dlmalloc)
+
+foreach(stub m rt pthread crypt util xnet resolve dl)
```
sunfishcode: Could you preserve the original comment "Create empty placeholder libraries" here?

CMakeLists.txt
```diff
@@ -0,0 +1,1178 @@
...
+    $<TARGET_FILE:c>
+    ${CMAKE_BINARY_DIR}/sysroot/lib/crt1.o
+  DEPENDS
+    ${PROJECT_SOURCE_DIR}/tools/gen-imports.py)
+add_custom_target(libc.imports ALL
+  DEPENDS ${CMAKE_BINARY_DIR}/sysroot/lib/libc.imports)
+add_dependencies(libc.imports c crt1.o)
+
+add_test(NAME check-metadata
+  COMMAND ${CMAKE_COMMAND} -E compare_files --ignore-eol ${PROJECT_SOURCE_DIR}/expected/wasm32-wasi/predefined-macros.txt ${CMAKE_BINARY_DIR}/sysroot/share/predefined-macros.txt)
```
sunfishcode: This is less friendly than running `diff` because it doesn't show what the differences were. Could you add a rule which runs an actual `diff` command?

compnerd:
`-DNDEBUG` is implicit in Release mode, and undefined in debug mode.  Do you want that always defined?

I believe that there are only two variants being built in the CMake build.  The third variant didn't have anyone building it, it's easy to add if desired.

Yes, running `ninja test` / `cmake --test out` will run the metadata tests.

pchickey:
@compnerd do you know how to resolve the build failure here https://github.com/CraneStation/wasi-sdk/pull/94/checks?check_run_id=394413947 - we're using the new CMake build for the sdk, but the libc is not installed yet because this is the step to build it.

compnerd:
@pchickey weird, I wonder how I didn't hit that, however, its just a benign check for the libc.  You can simply add a `-DCMAKE_C_COMPILER_WORKS=YES` when configuring wasi-libc to bootstrap the SDK.

abrown:
Can we close this? #330 is attempting the same kind of thing but is more recent.

abrown:
Ok, I'm closing this; feel free to reopen if this is a better approach.

abrown:
@compnerd, looking through your comments on #330, did you feel this was the better PR to push this idea forward? I can reopen it if that is the case! (I was just trying to clean up old PRs and issues).

PR #159, opened 2020-01-14
Fix environment variable init to exit successfully.

sunfishcode:
This fixes https://github.com/bytecodealliance/wasmtime/issues/783.


sbc100:
Woah..   did we ship this is version wasi-sdk already?  I guess that means that basic test we have there don't cover this ?

sunfishcode:
Yeah, we didn't previously have any tests that set environment variables. I've now opened https://github.com/CraneStation/wasi-sdk/pull/95 to add such a test.

PR #161, opened 2020-01-15
Don't call `free` on paths which are about to call `_Exit`.

sunfishcode:
This is a minor code-size optimization.


sbc100:
Does this kind of optimization break lsan?

sunfishcode:
It will appear as causing a leak, but only in the case where `fd_prestat_dir_name` or `malloc` fail during program startup, which should be very rare and would indicate more serious problems are afoot.

sbc100:
OK lgtm.

PR #167, opened 2020-02-19
Switch CI to fully use github actions

pchickey:
I started by porting the azure pipeline to github actions. I ended up changing how we install llvm on windows because some environment var stuff was behaving differently on windows with github vs azure.

Additionally, uploads the built sysroot as an artifact, so it can be retrieved from the "Checks" pane of each PR.


.github/workflows/main.yml
```diff
@@ -2,8 +2,62 @@ name: CI
...
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+
+    - name: Install clang (Windows)
+      shell: cmd
+      run: |
+        powershell -Command "$ProgressPreference = 'SilentlyContinue'; iwr -outf %TEMP%\LLVM-8.0.0-win64.exe https://rust-lang-ci2.s3.amazonaws.com/rust-ci-mirror/LLVM-8.0.0-win64.exe"
```
sbc100: Can you remember why we can't use releases.llvm.org for windows?  I seem to remember there was a bug?  I wonder if updating to 9.0.0 would fix the issue?

pchickey: idk, im just trying to translate the stuff thats on azure right now. Lets try upgrading it to 9.0.0 after all this stuff works.

currently stuck on an error about llvm-ar not being found (i think?) on windows, but I have to run for the night before fixing that

pchickey: Ended up switching to the llvm release after all, but kept everything at 8.0.0 for now.

.github/workflows/main.yml
```diff
@@ -2,8 +2,67 @@ name: CI
...
+        os: [ubuntu-latest, macos-latest, windows-latest]
+    steps:
+    - uses: actions/checkout@v1
+      with:
+        submodules: true
+
+    - name: Install clang (Windows)
+      shell: bash
+      run: |
+          curl -fsSLO https://releases.llvm.org/8.0.0/LLVM-8.0.0-win64.exe
```
sbc100: So this .exe is secretly a 7zip formatting file?  Strange.

.github/workflows/main.yml
```diff
@@ -2,8 +2,67 @@ name: CI
...
+          7z x LLVM-8.0.0-win64.exe -y -o"llvm"
+          echo "::add-path::$(pwd)/llvm/bin"
+          echo "::set-env name=WASM_AR::$(pwd)/llvm/bin/llvm-ar.exe"
```
sbc100: How do these echo commands to?  Do that perform some action?  Is github reading the stdout of the bash and doing something here?

.github/workflows/main.yml
```diff
@@ -2,8 +2,67 @@ name: CI
...
+      if: matrix.os == 'windows-latest'
+
+    - name: Install llvm-nm (Windows)
+      shell: bash
+      run: |
+        # Add --no-self-update as a workaround for
+        # https://github.com/microsoft/azure-pipelines-image-generation/issues/1224
```
sbc100: Looks like this issue is closed.. can we drop this step completely?

.github/workflows/main.yml
```diff
@@ -2,8 +2,67 @@ name: CI
...
+
+    - name: Build libc
+      shell: bash
+      run: make -j4
+
+    - uses: actions/upload-artifact@v1
+      with:
+        # Upload the sysroot folder. Give it a name according to the OS it was built for.
+        name: ${{ format( 'sysroot-{0}.tgz', matrix.os) }}
+        path: sysroot
```
sbc100: Hmm, do we need this?

Firstly, it should be platform-independent so at least I would upload only from the linux builder.  

Secondly I wonder if we should call it something other than sysroot (maybe wasi-libc-prebuilt?).

Thirdly, does any actually need/want this artifact.   I'm not sure that very useful on its own without the rest of wasi-sdk (wasi-sdk already uploads its own complete sysroot for those who have their own clang).

.github/workflows/main.yml
pchickey: I have no idea why this works. I copy-pasted it from somewhere on the internet, and appears to do just fine. I don't have a windows machine on which to even try it, so I only know that it works here :)

.github/workflows/main.yml
pchickey: Yes, these echo commands tell GH actions to append to the path or set an environment variable. Azure had something that worked similarly but looked a lot more like line noise. https://help.github.com/en/actions/reference/development-tools-for-github-actions#set-an-environment-variable-set-env

.github/workflows/main.yml
pchickey: I think this is useful only for troubleshooting builds. It doesn't cost anything (well, it probably costs Microsoft something in storage, but they have the money) or publish the artifacts anywhere prominent, but authors and reviewers can use it to verify that the sysroot built what they expected on platforms they don't have access to.

Ideally we'd add a separate step that downloads all three artifacts and verifies that they're the same, since we expect the build to be the same on all operating systems.

We do the same thing in wasi-sdk PRs as well, for the same reasons.

sbc100: I don't mind uploading for CI/debugging purposes.  I was just thinking this might not make sense in the github releases.. is that a separate thing?

.github/workflows/main.yml
sbc100: Oh wow.. thats gross.

.github/workflows/main.yml
pchickey: fixed d0ea5ee

.github/workflows/main.yml
pchickey: yeah, this is a separate thing from github releases. Agree that its confusing, I had to figure it out the hard way too

PR #169, opened 2020-02-19
CI: upgrade to llvm 9.0.0

pchickey:


.github/workflows/main.yml
```diff
@@ -34,17 +34,17 @@ jobs:
...
         echo "::add-path::$CLANG_DIR"
         echo "::set-env name=WASM_CC::$CLANG_DIR/clang"
       if: matrix.os == 'macos-latest'
 
     - name: Install clang (Linux)
       shell: bash
       run: |
-        curl -sSf http://releases.llvm.org/8.0.0/clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz | tar xJf -
-        export CLANG_DIR=`pwd`/clang+llvm-8.0.0-x86_64-linux-gnu-ubuntu-16.04/bin
+        curl -sSf https://releases.llvm.org/9.0.0/clang+llvm-9.0.0-x86_64-linux-gnu-ubuntu-16.04.tar.xz | tar xJf -
```
sbc100: http!  nice catch.

sbc100:
Do we want the next version of the SDK be based on llvm 9.0?

It would mean that anyone who wants to bring thier own llvm or use system llvm also need to be on version 9?   I think that requirement is fine, given that its the current stable version.  What do you think @sunfishcode ?

pchickey:
We've historically tracked the latest stable, so I think its a good idea.

sunfishcode:
Tracking stable makes sense to me.

@pchickey How hard would it be to add CI rules that test that the SDK still builds using LLVM 8.0? As far as I'm aware, that should still work; we just don't have any tests for it right now.

.github/workflows/main.yml
pchickey: the link URL from the llvm release website is http, not https. Someone over there should please update that!

pchickey:
I could add a build matrix dimension for llvm version, we'd have to duplicate some steps but it probably wouldn't be too big of a deal.

sbc100:
I'm not sure its worth the effort TBH.  I think its ok to say "this is the version we support and test against, other versions YMMV".

sbc100:
Should we cut a release of the SDK based on llvm 9.0.0 to keep it all in sync?

pchickey:
Its about time, yeah. @sunfishcode how much of your current stuff do you want to land before an SDK release?

PR #183, opened 2020-03-10
Emulate <signal.h>

sunfishcode:
wasm doesn't support signals, so therefore wasi-libc doesn't support signals.

But we can emulate them, at least for the synchronous `raise` case. This PR now implements basic signal emulation.

[edited to reflect the new content of the PR]


sunfishcode:
I ended up rewriting this PR to implement an emulation layer instead of just removing everything, so this PR now supports `raise(SIGABRT)`, and even `signal(SIGABRT, SIG_IGN)`. The emulation layer is disabled by default, and enabled with `-D_WASI_EMULATED_SIGNAL`, similar to `-D_WASI_EMULATED_MMAN` which provides mmap support.

I have some tests that I'll submit to wasi-sdk's testsuite separately.

PR #184, opened 2020-03-10
Lazy-initialize the environment variables.

### LINKED ISSUE #180: Environment variables don't work if `main` isn't used
We discovered an issue in the Rust `wasm32-wasi` target today where a wasm binary uses `std::env` but doesn't end up importing the requisite wasi items to actually get env vars working.

The broken example looks like this:

```rust
#[no_mangle]
pub extern fn foo() {
    println!("{:?}", std::env::var("FOO"));
}
```

And when compiled:

```
$ rustc +nightly foo.rs --target wasm32-wasi -O --crate-type cdylib
$ wasm2wat foo.wasm | grep import
  (import "wasi_snapshot_preview1" "fd_write" (func $_ZN4wasi13lib_generated22wasi_snapshot_preview18fd_write17h8cb541fbb70a6a5bE (type 7)))
```

This is showing that `environ_get` isn't imported at all!

When compiled as a binary though it does indeed work:

```rust
fn main() {
    println!("{:?}", std::env::var("FOO"));
}
```

```
$ rustc +nightly foo.rs --target wasm32-wasi -O
$ wasm2wat foo.wasm | grep import
  (import "wasi_snapshot_preview1" "proc_exit" (func $__wasi_proc_exit (type 1)))
  (import "wasi_snapshot_preview1" "fd_write" (func $_ZN4wasi13lib_generated22wasi_snapshot_preview18fd_write17h8cb541fbb70a6a5bE (type 9)))
  (import "wasi_snapshot_preview1" "fd_prestat_get" (func $__wasi_fd_prestat_get (type 3)))
  (import "wasi_snapshot_preview1" "fd_prestat_dir_name" (func $__wasi_fd_prestat_dir_name (type 8)))
  (import "wasi_snapshot_preview1" "environ_sizes_get" (func $__wasi_environ_sizes_get (type 3)))
  (import "wasi_snapshot_preview1" "environ_get" (func $__wasi_environ_get (type 3)))
```

I'm told that @sunfishcode is working on a fix!

sunfishcode:
This is the first in a series of PRs to make it easier to use WASI libc
in Wasm modules that don't have a `main` function. By initializing the
environment on demand, we avoid depending on having `__wasm_call_ctors`
run.

This uses weak symbols strategically to ensure that if `environ` is
used, it is initialized eagerly, but if only `getenv` and friends
are used, the environment is initialized lazily.

Eventually, I expect we'll have a convention for wasm modules without
main functions which will allow the `__wasm_call_ctors` function to be
called automatically, but this helps in simple cases for now.

Fixes #180. cc @alexcrichton

### DIFF
--- a/libc-top-half/musl/src/env/clearenv.c
+++ b/libc-top-half/musl/src/env/clearenv.c
@@ -9,6 +9,9 @@ int clearenv()
 {
 #ifdef __wasilibc_unmodified_upstream // Lazy environment variable init.
 #else
+// This specialized header is included within the function body to arranges for
+// the environment variables to be lazily initialized. It redefined `__environ`,
+// so don't remove or reorder it with respect to other code.
 #include "wasi/libc-environ-compat.h"
 #endif
 	char **e = __environ;
--- a/libc-top-half/musl/src/env/clearenv.c
+++ b/libc-top-half/musl/src/env/clearenv.c
@@ -7,6 +7,10 @@ weak_alias(dummy, __env_rm_add);
 
 int clearenv()
 {
+#ifdef __wasilibc_unmodified_upstream // Lazy environment variable init.
+#else
+#include "wasi/libc-environ-compat.h"
+#endif
 	char **e = __environ;
 	__environ = 0;
 	if (e) while (*e) __env_rm_add(*e++, 0);
--- a/expected/wasm32-wasi/defined-symbols.txt
+++ b/expected/wasm32-wasi/defined-symbols.txt
@@ -39,7 +39,6 @@ __env_rm_add
 __env_rm_add
 __env_rm_add
 __env_rm_add
-__environ
 __exp2f_data
 __exp_data
 __expo2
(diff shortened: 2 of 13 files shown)


libc-top-half/musl/src/env/clearenv.c
```diff
@@ -7,6 +7,10 @@ weak_alias(dummy, __env_rm_add);
 
 int clearenv()
 {
+#ifdef __wasilibc_unmodified_upstream // Lazy environment variable init.
+#else
+#include "wasi/libc-environ-compat.h"
```
pchickey: I'd add a comment here (& in the other uses) indicating that this header is intended to be used inside a func body, and it redefines the __environ symbol. The latter seems like it may cause unintended consequences if someone performed otherwise innocent-looking code motion without understanding the details of that header.

sunfishcode: Comments added. The most common cases of code motion are guarded against, because if you use `__environ` before it's defined, you'll get errors, and if you use the redefined `__environ` in a function other than the one the header was included in, `__wasilibc_environ` won't be declared and you'll get errors. But I agree, this is particularly tricky code, so comments are good.

PR #185, opened 2020-03-10
Update WASI Libc's README.md.

sunfishcode:
Update the description, point users to wasi-sdk as a simpler place to
get started using this library, and remove old text about being a
"reference" implementation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,9 @@ To use the sysroot, use the `--sysroot=` option:
 to run the compiler using the newly built sysroot.
 
 Note that Clang packages typically don't include cross-compiled builds of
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -48,3 +36,10 @@ To use the sysroot, use the `--sysroot=` option:
 ```
 
 to run the compiler using the newly built sysroot.
+
+Note that Clang packages typically don't include cross-compiled builds of
+compiler-rt, `libclang_rt.builtins-wasm32.a`, so they may not be usable without
```
sbc100: How about generalizing: `Clang packages typically don't include cross-compiled builds of standard libraries such as compiler-rt or libc++abi` since its not just compiler-it

sunfishcode: Done!

PR #186, opened 2020-03-18
Makefile refactoring; NFC

sunfishcode:
This PR contains several Makefile refactorings which were previously folded into #150, now pulled out to address https://github.com/WebAssembly/wasi-libc/pull/150#discussion_r393383090. NFC.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -443,11 +443,13 @@ finish: startup_files libc
 	#
 	mkdir -p "$(SYSROOT_SHARE)"
 
+	#
 	# Collect symbol information.
-	# TODO: Use llvm-nm --extern-only instead of grep. This is blocked on
-	# LLVM PR40497, which is fixed in 9.0, but not in 8.0.
-	# Ignore certain llvm builtin symbols such as those starting with __mul
-	# since these dependencies can vary between llvm versions.
+	#
+	@# TODO: Use llvm-nm --extern-only instead of grep. This is blocked on
+	@# LLVM PR40497, which is fixed in 9.0, but not in 8.0.
+	@# Ignore certain llvm builtin symbols such as those starting with __mul
+	@# since these dependencies can vary between llvm versions.
 	"$(WASM_NM)" --defined-only "$(SYSROOT_LIB)"/libc.a "$(SYSROOT_LIB)"/*.o \
 	    |grep ' [[:upper:]] ' |sed 's/.* [[:upper:]] //' |LC_ALL=C sort > "$(SYSROOT_SHARE)/defined-symbols.txt"
 	for undef_sym in $$("$(WASM_NM)" --undefined-only "$(SYSROOT_LIB)"/*.a "$(SYSROOT_LIB)"/*.o \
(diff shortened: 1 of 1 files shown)


sunfishcode:
There are several changes here, but they're split into commits which can be reviewed independently.

Makefile
```diff
@@ -204,7 +204,7 @@ endif
 # Set the sysroot.
 override WASM_CFLAGS += --sysroot="$(SYSROOT)"
 
-objs = $(patsubst $(CURDIR)/%.c,$(OBJDIR)/%.o,$(1))
+override objs = $(patsubst $(CURDIR)/%.c,$(OBJDIR)/%.o,$(1))
```
sbc100: Can we skip this patch?  Having override everywhere makes all the lines longer and, for me, makes it harder to read.

I guess the existing overrides are there so that variables can be set on the command line and then appended to?   How about we instread remove *all* the overrides, and not try to support variables both set on the command line and appended to?   Or if there are such variables split them into e.g. CFLAGS and EXTRA_CFLAGS.

Makefile
```diff
@@ -443,11 +443,13 @@ finish: startup_files libc
...
 	mkdir -p "$(SYSROOT_SHARE)"
 
+	#
 	# Collect symbol information.
-	# TODO: Use llvm-nm --extern-only instead of grep. This is blocked on
-	# LLVM PR40497, which is fixed in 9.0, but not in 8.0.
-	# Ignore certain llvm builtin symbols such as those starting with __mul
-	# since these dependencies can vary between llvm versions.
+	#
+	@# TODO: Use llvm-nm --extern-only instead of grep. This is blocked on
```
sbc100: huh.. TIL that make echoes comments like this to the screen.

Makefile
```diff
@@ -60,72 +60,40 @@ override LIBC_TOP_HALF_MUSL_DIR = $(LIBC_TOP_HALF_DIR)/musl
...
-    $(LIBC_TOP_HALF_MUSL_SRC_DIR)/dirent/versionsort.c \
-    $(LIBC_TOP_HALF_MUSL_SRC_DIR)/env/clearenv.c \
-    $(LIBC_TOP_HALF_MUSL_SRC_DIR)/env/getenv.c \
-    $(LIBC_TOP_HALF_MUSL_SRC_DIR)/env/putenv.c \
-    $(LIBC_TOP_HALF_MUSL_SRC_DIR)/env/setenv.c \
-    $(LIBC_TOP_HALF_MUSL_SRC_DIR)/env/unsetenv.c \
-    $(LIBC_TOP_HALF_MUSL_SRC_DIR)/unistd/posix_close.c \
+    $(addprefix $(LIBC_TOP_HALF_MUSL_SRC_DIR)/, \
+        misc/a64l.c \
+        misc/basename.c misc/dirname.c \
```
sbc100: What rule have use used to join these file onto the same line?

Any reason not to stick with one-line-per-file?  (I normally like to do this to make diff in particular more readable)

Makefile
sunfishcode: `override` means the thing *can't* be overridden from the command-line. The default is that everything can be overridden from the command-line. The idea is to try to clearly identify which things people can override and which things they can't. I agree that the syntax is ugly, and admit that it's probably overzealous to rigorously defend against rogue command-line arguments though.

Makefile
sunfishcode: It's just ad-hoc groupings of things that are related. I find reducing the number of lines makes a file easier to read.

Makefile
sbc100: My preference would be drop them all then.  I'm not worried about someone who accidentally passes `BASICS_OBJS=x` on the command line.  I'll leave it up to you those.  Not a blocker.

sunfishcode: I've now removed them.

Makefile
sunfishcode: But I don't feel strongly about it. I've now submitted a change to put each file on its own line.

sunfishcode: But, I don't feel strongly about it, so I've now changed it to put each file on its own line.

Makefile
```diff
@@ -497,17 +513,16 @@ finish: startup_files libc
...
+	#
+	# The build succeeded! The generated sysroot is in $(SYSROOT).
+	#
+
 install: finish
 	mkdir -p "$(INSTALL_DIR)"
 	cp -r "$(SYSROOT)/lib" "$(SYSROOT)/share" "$(SYSROOT)/include" "$(INSTALL_DIR)"
 
-.PHONY: default startup_files libc finish check install include_dirs
+.PHONY: default startup_files libc finish install include_dirs
```
sbc100: Don't you feel like there is value is having a separate check rule?   I guess we always build from clean anyway, so its not like you could build and then check later with the current setup?

sunfishcode: The check code is very quick -- it's just a diff of a few files -- so we can just always do the check as part of the regular build.

PR #187, opened 2020-03-18
Fix -std=gnu17 when generating predefined_macros.txt

sbc100:
The default recently changed in upstream clang:
https://reviews.llvm.org/D75383

We want to be immune to such things when generating this list so that
we can build wasi-libc with any recent clang version.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -480,7 +480,7 @@ finish: startup_files libc
 	# TODO: Filter out __FLT16_* for now, as not all versions of clang have these.
 	"$(WASM_CC)" $(WASM_CFLAGS) "$(SYSROOT_SHARE)/include-all.c" \
 	    -isystem $(SYSROOT_INC) \
-	    -std=gnu11 \
+	    -std=gnu17 \
 	    -E -dM -Wno-\#warnings \
 	    -D_ALL_SOURCE \
 	    -U__llvm__ \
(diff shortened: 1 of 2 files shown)


sbc100:
It doesn't matter which we pick I guess.   But if I pick 17 then this change would also include a diff to the predefined macros files like:

```
diff -wur "/b/s/w/ir/cache/builder/src/src/work/wasi-libc/expected/wasm32-wasi" "/b/s/w/ir/cache/builder/src/src/work/wasi-libc-out/share/wasm32-wasi"
diff -wur /b/s/w/ir/cache/builder/src/src/work/wasi-libc/expected/wasm32-wasi/predefined-macros.txt /b/s/w/ir/cache/builder/src/src/work/wasi-libc-out/share/wasm32-wasi/predefined-macros.txt
--- /b/s/w/ir/cache/builder/src/src/work/wasi-libc/expected/wasm32-wasi/predefined-macros.txt	2020-03-17 13:35:30.539327488 -0700
+++ /b/s/w/ir/cache/builder/src/src/work/wasi-libc-out/share/wasm32-wasi/predefined-macros.txt	2020-03-17 19:59:23.777684838 -0700
@@ -2774,7 +2774,7 @@
 #define __STDC_ISO_10646__ 201206L
 #define __STDC_UTF_16__ 1
 #define __STDC_UTF_32__ 1
-#define __STDC_VERSION__ 201112L
+#define __STDC_VERSION__ 201710L
 #define __STDC__ 1
 #define __STDDEF_H 
```

Would you prefer that?

sunfishcode:
Ok, good to know that's the only change. So it doesn't matter much right now, but my general preference is to go with the newer version, in case any of our headers gain features which are only enabled in the newer version.

sbc100:
OK done.

PR #188, opened 2020-03-27
Added utime.h

ritave:


expected/wasm32-wasi/predefined-macros.txt
```diff
@@ -2026,18 +2026,18 @@
 #define YESSTR 0x50002
 #define YXDOMAIN ns_r_yxdomain
 #define YXRRSET ns_r_yxrrset
-#define _ALLOCA_H
```
sbc100: Why idea whats happening with all these whitespace changes?

ritave: Sure - the line endings were inconsistent, intermixing Linux (LF) and Windows (CR LF) newlines. Additionally a few lines had spaces at the end of lines.
I've decided to leave it in, but I can revert if you prefer :)

sbc100: Would you mind landing the no-op update of this file separately?

sunfishcode: This file is entirely auto-generated, and should remain so, so that when people do builds and generate new versions of this file, we can compare with with `diff`. The easiest way to keep it up to date when making changes is to just copy your newly generated file into the tree, rather than manually editing it.

sbc100: Indeed, I was assuming that it was in an inconsistent state previously causing a larger-than-expected diff when it was copied.    Hense the suggestion of a no-op re-built and copy prior to landing.  But indeed, if you are hand-editing that is the wrong approach.

libc-bottom-half/libpreopen/libpreopen.c
```diff
@@ -155,6 +156,23 @@ stat(const char *path, struct stat *st)
...
+
+    // If we can't find a preopened directory handle to open this file with,
+    // indicate that the program lacks the capabilities.
+    if (fd == -1) {
+        errno = ENOTCAPABLE;
+        return -1;
+    }
+    return utimensat(fd, relative_path, times ? ((struct timespec [2]){
+		{ .tv_sec = times->actime }, { .tv_sec = times->modtime }})
+        : 0,  AT_SYMLINK_NOFOLLOW);
```
sunfishcode: Is this intended to be `AT_SYMLINK_NOFOLLOW`? I don't have a definitive reference handy, but most other things in POSIX follow symlinks.

ritave: I've followed lead with the above `int stat` function https://github.com/WebAssembly/wasi-libc/blob/36e6c09f998ea233a1f57b1e4322e1ca9ab222ac/libc-bottom-half/libpreopen/libpreopen.c#L156

sunfishcode: Ah, that's an error, likely an accidental copy+pasto from `lstat`. It should be 0.

ritave:
@sbc100 @sunfishcode I've resolved all of your issues and the code is now clean for a re-review

PR #189, opened 2020-04-01
Fix references to musl version

ggreif:
Git revision needs double-checking!

N.B.: `libc-bottom-half` still seems to refer to an older `musl`.

### DIFF
--- a/libc-top-half/musl/README
+++ b/libc-top-half/musl/README
@@ -20,4 +20,4 @@ Information on full musl-targeted compiler toolchains, system
 bootstrapping, and Linux distributions built on musl can be found on
 the project website:
 
-    http://www.musl-libc.org/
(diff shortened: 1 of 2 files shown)


libc-top-half/musl/README
```diff
@@ -20,4 +20,4 @@ Information on full musl-targeted compiler toolchains, system
 bootstrapping, and Linux distributions built on musl can be found on
 the project website:
 
-    http://www.musl-libc.org/
+    http://musl.libc.org/
```
ggreif, line 23: Should we switch to `https://`?
```suggestion
    https://musl.libc.org/
```

pchickey: Yes!

pchickey:
Thanks!

PR #197, opened 2020-05-23
Make mmap() possible to use in a conformant way

whitequark:
wasi-libc defines `MAP_SHARED_VALIDATE` as `MAP_SHARED|MAP_PRIVATE`: https://github.com/WebAssembly/wasi-libc/blob/86550c37ab14962233f73413ea0e0c0e57d056d7/libc-top-half/musl/include/sys/mman.h#L25-L27

POSIX [requires](https://pubs.opengroup.org/onlinepubs/9699919799/functions/mmap.html) either `MAP_SHARED` or `MAP_PRIVATE` to be specified: Screenshot_20200523_144435

Because `MAP_SHARED_VALIDATE` is not an unique bit flag, the current code makes it impossible to use `mmap()` in wasi-libc in a standards-conformant way.

For context, here's the rather sad [history](https://lwn.net/Articles/758594/) behind `MAP_SHARED_VALIDATE`.

### DIFF
--- a/libc-bottom-half/mman/mman.c
+++ b/libc-bottom-half/mman/mman.c
@@ -23,9 +23,10 @@ struct map {
 void *mmap(void *addr, size_t length, int prot, int flags,
            int fd, off_t offset) {
     // Check for unsupported flags.
-    if ((flags & MAP_FIXED) != 0 ||
+    if ((flags & (MAP_PRIVATE | MAP_SHARED)) == 0 ||
+        (flags & MAP_FIXED) != 0 ||
 #ifdef MAP_SHARED_VALIDATE
-        (flags & MAP_SHARED_VALIDATE) != 0 ||
+        (flags & MAP_SHARED_VALIDATE) == MAP_SHARED_VALIDATE ||
 #endif
 #ifdef MAP_NORESERVE
         (flags & MAP_NORESERVE) != 0 ||


pchickey:
This is a change to code we're using from musl, and I haven't heard of this flag or its history before, but just to check my understanding - this is a place where Linux decided to disobey the POSIX standard in a syscall ABI, and since we are not Linux we can switch to a representation that conforms to POSIX without breaking any user code, correct?

sunfishcode:
Thanks for finding and fixing this! It's possible I introduced this bug when I made a pass through the mmap documentation to collect unrecognized flags to check, but didn't retest afterwords. Since I should have done this before, I've now opened https://github.com/WebAssembly/wasi-sdk/pull/137 adding a basic testcase for this.

@pchickey Yes, this issue just concerns values defined by our musl-derived sys/mman.h, which is only used by userspace right now. A PR to change it to avoid using Linux's little trick would be welcome. I think the PR here is also valuable, either way, as it fixes the code to avoid making an implicit assumption about the headers.

whitequark:
> this is a place where Linux decided to disobey the POSIX standard in a syscall ABI

Strictly speaking POSIX leaves the beahavior with `MAP_PRIVATE|MAP_SHARED` undefined, so Linux does not violate the letter of POSIX. But it is very easy to make the case that it violates the spirit of POSIX. I personally think this trick is quite gross so you wouldn't see me defending it, I was just preserving whatever was in the codebase already.

whitequark:
> Thanks for finding and fixing this!

No problem, WASM solves a *lot* of my headache and I'm happy to spend time finding bugs in it if something doesn't work. All I ask in return is a release of `wasi-sdk` that wouldn't be many months off, so I can actually use the fixes :)

sbc100:
So an alternative to this patch would be to remove `MAP_SHARED_VALIDATE`  completely from our libc headers, since it is  linux specific?   Thus forcing downstream code to use only the POSIX flags.

On the other hand one could argue that we doing pure emulation here anyway so perhaps making an effort to emulate linux make more code work under wasi-libc?   Otherwise why emulate mmap at all.

pchickey:
@whitequark Allegedly the build scripting is now all un-bit-rotted so that a `wasi-sdk` release is pretty easy and just requires a little bit of manual work. I never went through the tribulations of yaml programming to make it fully automated on pushing a git tag, but we really should do them more often, so perhaps we can do one at the end of this week?

whitequark:
That would be wonderful.

pchickey:
The only additional behavior required of `MAP_SHARED_VALIDATE` is that we return EOPNOTSUPP if an unknown flag bit is set. We can emulate that easily enough if we choose to, so I guess the whole argument is whether we want to maximize the amount of code that works under wasi-libc, or stick to just POSIX. I lean towards maximizing how much code works, unless someone else has a strong opinion on this?

sunfishcode:
I agree; since mmap is a compatibility feature that users have to use `-D_WASI_EMULATED_MMAN` to opt into, I think it makes sense to support Linux (or any other platform) features, and not just POSIX.

whitequark:
Okay, should I implement it in this PR?

sunfishcode:
If you don't mind, a new PR for the header change would be good here.

whitequark:
Okay. Is there anything blocking this one?

sunfishcode:
Merged!

whitequark:
Thanks!

PR #199, opened 2020-05-29
Merge the basics component into libc-bottom-half.

sunfishcode:
We no longer have a need to maintain basics as a separate component.
Folding it into libc-bottom-half eliminates a fair amount of redundancy.

This follows up on the comment in #74 to remove the redundant crt1.c, and removes a lot of other redundant things too.


libc-bottom-half/cloudlibc/src/libc/stdlib/qsort.c
```diff
@@ -1,15 +0,0 @@
-// Copyright (c) 2015 Nuxi, https://nuxi.nl/
```
sbc100, line 1: Why is not needed anymore?

libc-bottom-half/sources/string.c
```diff
@@ -1,40 +0,0 @@
-#include <string.h>
-#include <stdint.h>
```
sbc100, line 2: Ditti.

libc-bottom-half/cloudlibc/src/libc/stdlib/qsort.c
sunfishcode: We use the "libc top half"'s `qsort, which is to say, musl's `qsort`.

libc-bottom-half/sources/string.c
sunfishcode: Ditti ditti :-).

libc-bottom-half/cloudlibc/src/libc/stdlib/qsort.c
sbc100: Gotcha

PR #203, opened 2020-06-11
New-style command support.

sunfishcode:
This enables [new-style command support]. Instead of calling
`__wasm_call_ctors` and `__wasm_call_dtors` directly, this lets
wasm-ld automatically call them.

And, this comments out a use of "protected" visibility, since
[WebAssembly doesn't support it].

[new-style command support]: https://reviews.llvm.org/D81689
[WebAssembly doesn't support it]: https://reviews.llvm.org/D81688

### DIFF
--- a/expected/wasm32-wasi/defined-symbols.txt
+++ b/expected/wasm32-wasi/defined-symbols.txt
@@ -276,6 +276,7 @@ _exit
 _flushlbf
 _initialize
 _start
+_start
 a64l
 abort
 abs
--- a/libc-bottom-half/cloudlibc/src/include/stdlib.h
+++ b/libc-bottom-half/cloudlibc/src/include/stdlib.h
@@ -139,8 +139,8 @@ void *calloc(size_t, size_t);
 div_t div(int, int) __pure2;
 double drand48(void);
 double erand48(__uint16_t *);
-_Noreturn void exit(int);
 #endif
+_Noreturn void exit(int);
 void free(void *);
 #ifdef __wasilibc_unmodified_upstream
 char *getenv(const char *);
--- a/libc-bottom-half/crt/crt1-command.c
+++ b/libc-bottom-half/crt/crt1-command.c
@@ -1,0 +1,18 @@
+#include <wasi/api.h>
+#include <stdlib.h>
+extern void __wasm_call_ctors(void);
+extern int __original_main(void);
+extern void __wasm_call_dtors(void);
+
+__attribute__((export_name("_start")))
+void _start(void) {
+    // Call `__original_main` which will either be the application's zero-argument
+    // `__original_main` function or a libc routine which calls `__main_void`.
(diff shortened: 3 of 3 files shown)


libc-top-half/musl/src/internal/stdio_impl.h
```diff
@@ -89,9 +89,11 @@ hidden int __towrite(FILE *);
 hidden void __stdio_exit(void);
 hidden void __stdio_exit_needed(void);
 
+#ifdef __wasilibc_unmodified_upstream // wasm has no "protected" visibility
```
sbc100, line 92: separate change?

libc-bottom-half/crt/crt1.c
```diff
@@ -2,23 +2,17 @@
 extern void __wasm_call_ctors(void);
 extern int __original_main(void);
 extern void __wasm_call_dtors(void);
+_Noreturn void exit(int);
```
sbc100, line 5: Since this is public API can you just include unistd instead?

libc-top-half/musl/src/internal/stdio_impl.h
sunfishcode: Separated out into its own PR now: https://github.com/WebAssembly/wasi-libc/pull/204

libc-bottom-half/crt/crt1.c
sunfishcode: Done.

sunfishcode:
This currently depends on wasm-ld having https://reviews.llvm.org/D81689 in it. Since we're just now dropping support for LLVM 8.0, we'll need to think more about how to manage compatibility here.

sbc100:
Should we tag/branch for each major llvm release that we support?

sunfishcode:
Branch/tag sounds difficult if these live a long time and we end up having to backport patches to all the branches.

I'm thinking about moving the crt1.c in this PR to crt1-command.c, and restoring the previous crt1.c, so that one libc can support both old and new styles -- the compiler driver would be able to pick which one it wants. I'll investigate this in more detail after https://reviews.llvm.org/D81689 lands.

sunfishcode:
https://reviews.llvm.org/D81689 has now landed, and I've now made the change to preserve the existing crt1.c and move the new one into crt1-command.c so that one WASI libc build can support both modes, to avoid the need to depend on the LLVM version.

expected/wasm32-wasi/defined-symbols.txt
```diff
@@ -276,6 +276,7 @@ _exit
 _flushlbf
 _initialize
 _start
+_start
```
whitequark, line 279: Why does `_start` appear twice now?

sunfishcode: It's defined in crt1.o and crt1-reactor.o. Programs will only use one or the, but the script which generates defined-symbols.txt doesn't know that.

PR #214, opened 2020-08-04
Add basic emulation of getcwd/chdir

alexcrichton:
This commit adds basic emulation of a current working directory to
wasi-libc. The `getcwd` and `chdir` symbols are now implemented and
available for use. The `getcwd` implementation is pretty simple in that
it just copies out of a new global, `__wasilibc_cwd`, which defaults to
`"/"`. The `chdir` implementation is much more involved and has more
ramification, however.

A new function, `make_absolute`, was added to the preopens object. Paths
stored in the preopen table are now always stored as absolute paths
instead of relative paths, and initial relative paths are interpreted as
being relative to `/`. Looking up a path to preopen now always turns it
into an absolute path, relative to the current working directory, and an
appropriate path is then returned.

The signature of `__wasilibc_find_relpath` has changed as well. It now
returns two path components, one for the absolute part and one for the
relative part. Additionally the relative part is always dynamically
allocated since it may no longer be a substring of the original input
path.

This has been tested lightly against the Rust standard library so far,
but I'm not a regular C developer so there's likely a few things to
improve!

### DIFF
--- a/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
+++ b/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
@@ -10,18 +10,19 @@ extern "C" {
  * map. If a suitable entry is found, then the file descriptor for that entry
  * is returned. Additionally the absolute path of the directory's file
  * descriptor is returned in `abs_prefix` and the relative portion that needs
- * to be opened is stored in `relative_path`.
+ * to be opened is stored in `relative_path`. The size of the `relative_path`
+ * buffer is provided in the `relative_path_len` argument.
  *
- * Returns -1 if no directories were suitable or if an allocation error
- * happens.
+ * Returns -1 on failure. Errno is set to either:
  *
- * On success the `relative_path` points to a malloc'd string, so you'll
- * need to call `free` on it. The `abs_prefix` return does not need to be
- * free'd.
+ *  * ENOMEM - failed to allocate memory for internal routines.
+ *  * ENOENT - the `path` could not be found relative to any preopened dir.
+ *  * ERANGE - the `relative_path` buffer is too small to hold the relative path.
  */
 int __wasilibc_find_relpath(const char *path,
                             const char **__restrict__ abs_prefix,
-                            char **__restrict__ relative_path);
+                            char *relative_path,
+                            size_t relative_path_len);
 
 #ifdef __cplusplus
 }
--- a/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
+++ b/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
@@ -10,8 +10,11 @@ extern "C" {
  * map. If a suitable entry is found, then the file descriptor for that entry
  * is returned. Additionally the absolute path of the directory's file
  * descriptor is returned in `abs_prefix` and the relative portion that needs
- * to be opened is stored in `relative_path`. The size of the `relative_path`
- * buffer is provided in the `relative_path_len` argument.
+ * to be opened is stored in `*relative_path`.
+ *
+ * The `relative_path` argument must be a pointer to a buffer valid  for
+ * `relative_path_len` bytes, and this may be used as storage for the relative
+ * portion of the path being returned through `*relative_path`.
  *
  * Returns -1 on failure. Errno is set to either:
  *
@@ -21,8 +24,25 @@ extern "C" {
  */
 int __wasilibc_find_relpath(const char *path,
                             const char **__restrict__ abs_prefix,
-                            char *relative_path,
+                            char **relative_path,
                             size_t relative_path_len);
+
+/**
+ * Look up the given `path`, which is interpreted as absolute, in the preopened
+ * directory map. If a suitable entry is found, then the file descriptor for
+ * that entry is returned. Additionally the relative portion of the path to
+ * where the fd is opened is returned through `relative_path`, the absolute
+ * prefix which was matched is stored to `abs_prefix`, and `relative_path` may
+ * be an interior pointer to the `abspath` string.
+ *
+ * Returns -1 on failure. Errno is set to either:
+ *
+ *  * ENOMEM - failed to allocate memory for internal routines.
+ *  * ENOENT - the `path` could not be found relative to any preopened dir.
+ */
+int __wasilibc_find_abspath(const char *abspath,
+                            const char **__restrict__ abs_prefix,
+                            const char **__restrict__ relative_path);
 
 #ifdef __cplusplus
 }
--- a/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
+++ b/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
@@ -16,6 +16,9 @@ extern "C" {
  * `relative_path_len` bytes, and this may be used as storage for the relative
  * portion of the path being returned through `*relative_path`.
  *
+ * See documentation on `__wasilibc_find_abspath` for more info about what the
+ * paths look like.
+ *
  * Returns -1 on failure. Errno is set to either:
  *
  *  * ENOMEM - failed to allocate memory for internal routines.
@@ -35,6 +38,9 @@ int __wasilibc_find_relpath(const char *path,
  * prefix which was matched is stored to `abs_prefix`, and `relative_path` may
  * be an interior pointer to the `abspath` string.
  *
+ * The `abs_prefix` returned string will not contain a leading `/`.
+ * Additionally the returned `relative_path` will not contain a leading `/`.
+ *
  * Returns -1 on failure. Errno is set to either:
  *
  *  * ENOMEM - failed to allocate memory for internal routines.
@@ -53,6 +59,8 @@ int __wasilibc_find_abspath(const char *abspath,
  * `__wasilibc_find_relpath`. The weak-nature of this symbols means that if it's
  * not otherwise included in the compilation then `chdir` wasn't used an there's
  * no need for this symbol.
+ *
+ * See documentation on `__wasilibc_find_relpath` for more information.
  */
 int __wasilibc_find_relpath_alloc(
     const char *path,
--- a/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
+++ b/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
@@ -38,8 +38,10 @@ int __wasilibc_find_relpath(const char *path,
  * prefix which was matched is stored to `abs_prefix`, and `relative_path` may
  * be an interior pointer to the `abspath` string.
  *
- * The `abs_prefix` returned string will not contain a leading `/`.
- * Additionally the returned `relative_path` will not contain a leading `/`.
+ * The `abs_prefix` returned string will not contain a leading `/`. Note that
+ * this may be the empty string. Additionally the returned `relative_path` will
+ * not contain a leading `/`. The `relative_path` return will not return an
+ * empty string, it will return `"."` instead if it would otherwise do so.
  *
  * Returns -1 on failure. Errno is set to either:
  *
--- a/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
+++ b/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
@@ -44,6 +44,24 @@ int __wasilibc_find_abspath(const char *abspath,
                             const char **__restrict__ abs_prefix,
                             const char **__restrict__ relative_path);
 
+/**
+ * Same as `__wasilibc_find_relpath`, except that this function will interpret
+ * `relative` as a malloc'd buffer that will be `realloc`'d to the appropriate
+ * size to contain the relative path.
+ *
+ * Note that this is a weak symbol and if it's not defined you can use
+ * `__wasilibc_find_relpath`. The weak-nature of this symbols means that if it's
+ * not otherwise included in the compilation then `chdir` wasn't used an there's
+ * no need for this symbol.
+ */
+int __wasilibc_find_relpath_alloc(
+    const char *path,
+    const char **abs,
+    char **relative,
+    size_t *relative_len,
+    int can_realloc
+) __attribute__((weak));
+
 #ifdef __cplusplus
 }
 #endif
--- a/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
+++ b/libc-bottom-half/headers/public/wasi/libc-find-relpath.h
@@ -6,14 +6,22 @@ extern "C" {
 #endif
 
 /**
- * Look up the given path in the preopened directory map. If a suitable
- * entry is found, return its directory file descriptor, and store the
- * computed relative path in *relative_path.
+ * Look up the given `path`, relative to the cwd, in the preopened directory
+ * map. If a suitable entry is found, then the file descriptor for that entry
+ * is returned. Additionally the absolute path of the directory's file
+ * descriptor is returned in `abs_prefix` and the relative portion that needs
+ * to be opened is stored in `relative_path`.
  *
- * Returns -1 if no directories were suitable.
+ * Returns -1 if no directories were suitable or if an allocation error
+ * happens.
+ *
+ * On success the `relative_path` points to a malloc'd string, so you'll
+ * need to call `free` on it. The `abs_prefix` return does not need to be
+ * free'd.
  */
 int __wasilibc_find_relpath(const char *path,
-                            const char **__restrict__ relative_path);
+                            const char **__restrict__ abs_prefix,
+                            char **__restrict__ relative_path);
 
 #ifdef __cplusplus
 }
--- a/libc-bottom-half/sources/chdir.c
+++ b/libc-bottom-half/sources/chdir.c
@@ -8,54 +8,63 @@
 #include <wasi/libc-find-relpath.h>
 #include <wasi/libc.h>
 
-char *__wasilibc_cwd = "/";
+extern char *__wasilibc_cwd;
 static int __wasilibc_cwd_mallocd = 0;
 
 int chdir(const char *path)
 {
-  // Find a preopen'd directory as well as a relative path we're anchored
-  // from which we're changing directories to.
-  char *relative;
-  const char *abs;
-  int parent_fd = __wasilibc_find_relpath(path, &abs, &relative);
-  if (parent_fd == -1) {
-    errno = ENOENT;
-    return -1;
-  }
+    static char *relative_buf = NULL;
+    static size_t relative_buf_len = 0;
 
-  // Make sure that this directory we're accessing is indeed a directory.
-  struct stat dirinfo;
-  int ret = fstatat(parent_fd, relative, &dirinfo, 0);
-  if (ret == -1) {
-    free(relative);
-    return -1;
-  }
-  if (!S_ISDIR(dirinfo.st_mode)) {
-    free(relative);
-    errno = ENOTDIR;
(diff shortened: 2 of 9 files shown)


libc-top-half/musl/src/unistd/getcwd.c
```diff
@@ -4,6 +4,8 @@
 #include <string.h>
 #include "syscall.h"
 
+char __wasilibc_cwd[PATH_MAX] = {'/', 0};
```
sunfishcode, line 7: WASI libc may move to deprecate `PATH_MAX` since every OS has its own limit, and it's desirable for WASI libc to avoid baking in a particular limit. We can use a `malloc` buffer here to avoid having a fixed size limit, with static initialization like this:

```c
char __wasilibc_cwd_buffer[] = {'/', 0};
char __wasilibc_cwd = __wasilibc_cwd_buffer;
```

This will require checking for `__wasilibc_cwd == __wasilibc_cwd_buffer` to avoid doing `free` or `realloc` on the static buffer.

libc-bottom-half/headers/public/wasi/libc-find-relpath.h
```diff
@@ -6,14 +6,22 @@ extern "C" {
...
+ * Returns -1 if no directories were suitable or if an allocation error
+ * happens.
+ *
+ * On success the `relative_path` points to a malloc'd string, so you'll
+ * need to call `free` on it. The `abs_prefix` return does not need to be
+ * free'd.
  */
 int __wasilibc_find_relpath(const char *path,
-                            const char **__restrict__ relative_path);
+                            const char **restrict abs_prefix,
```
sbc100, line 23: __restrict__?

libc-top-half/musl/src/unistd/getcwd.c
```diff
@@ -14,12 +16,21 @@ char *getcwd(char *buf, size_t size)
...
 		return 0;
 	}
+#else
+        size_t len = strlen(__wasilibc_cwd);
+        if (size < len + 1) {
+                errno = ERANGE;
+                return 0;
+        }
+        memcpy(buf, __wasilibc_cwd, len + 1);
+#endif
```
sunfishcode, line 34: In the case where `buf` is `NULL`, we should just `malloc(len + 1)` and `memcpy` into it, rather than artificially limiting it to `PATH_MAX` bytes. A key differences here is that musl is written for Linux where it has to allocate the buffer before making the syscall. For WASI, we can just do the `strlen` right away, check the length if using a pre-allocated buffer, and `malloc` if not.

So I think it makes to have our own `getcwd.c` in libc-bottom-half/sources/getcwd.c rather than base this on musl's getcwd.c.

libc-top-half/musl/src/unistd/chdir.c
```diff
@@ -1,7 +1,46 @@
...
+        }
+
+        // Copy over our new absolute path into `__wasilibc_cwd`
+        close(fd);
+        strcpy(__wasilibc_cwd, abs);
+        if (strcmp(relative, ".") != 0)
+            strcpy(__wasilibc_cwd + strlen(abs), relative);
+        free(relative);
+        return 0;
+#endif
```
sunfishcode, line 45: In cases like this where the entire musl function body is replaced with WASI-specific code, it's better to just create a new file rather than making musl's conditional. We can put this in libc-bottom-half/sources/chdir.c.

libc-top-half/musl/src/unistd/chdir.c
```diff
@@ -1,7 +1,46 @@
...
+        char *relative;
+        const char *abs;
+        int parent_fd = __wasilibc_find_relpath(path, &abs, &relative);
+        if (parent_fd == -1) {
+                errno = ENOENT;
+                return -1;
+        }
+
+        // Make sure that this directory we're accessing is indeed a directory.
+        int fd = __wasilibc_openat_nomode(parent_fd, relative, O_DIRECTORY | O_RDONLY);
```
sunfishcode, line 32: The directory check could be a little cheaper with `fstatat` and `S_ISDIR(stat.st_mode)`.

libc-bottom-half/sources/posix.c
```diff
@@ -18,96 +19,115 @@ int open(const char *path, int oflag, ...) {
...
 
     // If we can't find a preopen for it, indicate that we lack capabilities.
     if (dirfd == -1) {
         errno = ENOTCAPABLE;
         return -1;
     }
 
-    return __wasilibc_openat_nomode(dirfd, relative_path, oflag);
+    int ret = __wasilibc_openat_nomode(dirfd, relative_path, oflag);
+    free(relative_path);
```
sunfishcode, line 33: It's unfortunate to add a `malloc`+`free` on every syscall. Obviously in the case of real host syscalls these will be relatively cheap, but in virtualized WASI environments, this could potentially be significant. Another potential concern is that we may in the future offer a WASI libc mode where `free` is a no-op and we just leak everything, to save the code size and runtime overhead of a real `malloc`. Such approaches are ok with leaking a bit of memory when the program is setting up, but less ok with leaking during operations like `open` which may be called many times.

That said, I also recognize that this allows for greater fidelity with POSIX and existing programs. I don't yet have a good suggestion for what to do here. It's possible we can address these concerns by making some things optional. I'll think about this more.

alexcrichton: I was thinking about this and at a fundamental level I think that malloc has to be supported at least sometimes, for example if you have a preopen of `/`, chdir to `/a`, and then open `b/c`, then you'll want to open `a/b/c` relative to `/`, but that's clearly no longer a subslice of the original path `b/c` since we needed to add `a/` onto it.

I could add an optimization where if a subslice of the original path can be returned then it is, but I'd need to add another return value flag for whether the relative path is owned or not and whether it needs to be deallocated.

sbc100: One traditional (and sometimes problematic) solution to this issue to have an static internal buffer here that gets realloc'd to be as large as is needs to be for each call.   This would be in TLS to allow in to work across threads.  The limitation of course is that successive calls to this function will clobber the output.  However, given that this is strictly a libc-internal function and not user facing, I don't see that as a big problem in this case.  This single buffer would then get a free'd during atexit.

alexcrichton: Ah good idea! I think though that would cause issues with `link` and `rename` which need to call `__wasilibc_find_relpath` twice and call `*at` functions with both path simultaneously.

Alternatively `__wasilibc_find_relpath` could be changed to not do the `malloc` itself but instead require a buffer get passed in, and it could return `ERANGE` when the buffer is too small. There could be a helper function in `posix.c` which would use one of perhaps two globals which cache a global buffer to be used. That would at least remove `malloc` from each per-syscall variant for most cases I think.

sbc100: For those cases that need to call `__wasilibc_find_relpath` twice they could just be aware of the issue and call `strdup()` before calling it a second time.   Since we control all the call sites I think its not so bad to have a few special cases like this.

alexcrichton:
Ok I've pushed up a second commit which should amortize the calls to `malloc` and reuse internal buffers where possible. I didn't register `atexit` handlers yet, though, but that may not be too too important in the context of wasm.

libc-bottom-half/sources/posix.c
```diff
@@ -214,13 +251,16 @@ int symlink(const char *target, const char *linkpath) {
     return symlinkat(target, dirfd, relative_path);
 }
 
+static char *relative_buf2 = NULL;
+static size_t relative_buf_len2 = NULL;
```
sunfishcode, line 255: It would tidy up the callsites that use these to have a function that holds these as function-local statics, similar to `find_relpath`.

libc-bottom-half/sources/preopens.c
```diff
@@ -109,41 +156,32 @@ static bool prefix_matches(const char *prefix, size_t prefix_len, const char *pa
 
 // See the documentation in libc.h
 int __wasilibc_register_preopened_fd(int fd, const char *prefix) {
-    prefix = strdup(prefix);
     return prefix == NULL ? -1 :
```
sunfishcode, line 159: With the preceding `strdup` removed, this NULL check is no longer needed.

libc-bottom-half/sources/chdir.c
```diff
@@ -0,0 +1,70 @@
...
+#include <unistd.h>
+#include <wasi/libc-find-relpath.h>
+#include <wasi/libc.h>
+
+char *__wasilibc_cwd = "/";
+static int __wasilibc_cwd_mallocd = 0;
+
+int chdir(const char *path)
+{
+  static char *relative_buf = NULL;
```
sunfishcode, line 16: Formatting nit: libc-bottom-half/sources sources use 4-space indent.

alexcrichton:
This is something I'd like to enable by default with rust-lang/rust, so if binary size is a concern I think it would be better to handle this for every program rather than just those that don't opt-in. I'd also imagine that almost all programs would want to opt-in given how prevalent the concept of a current directory is. I feel like though this sort of a consequence of the wasi-libc strategy, which is that if we want to put stuff in wasm rather than WASI APIs that's inevitably going to make modules larger.

Is there a way to do weak linking tricks or something like that to only pull in this extra processing when `chdir` is used? If `chdir` is never used then there's really no need for most of this processing since the current directory is always `/`.

sbc100:
> This is something I'd like to enable by default with rust-lang/rust, so if binary size is a concern I think it would be better to handle this for every program rather than just those that don't opt-in. I'd also imagine that almost all programs would want to opt-in given how prevalent the concept of a current directory is. I feel like though this sort of a consequence of the wasi-libc strategy, which is that if we want to put stuff in wasm rather than WASI APIs that's inevitably going to make modules larger.
> 
> Is there a way to do weak linking tricks or something like that to only pull in this extra processing when `chdir` is used? If `chdir` is never used then there's really no need for most of this processing since the current directory is always `/`.

I think that might be possible yes.   I think you would want to co-locate `chdir` and `find_relpath` in the same object file, then have callers of `find_relpath` reference it as weak symbol (maybe you could just move find_relpath2 so only once place would need to check for the weak sym).

sbc100:
> This is something I'd like to enable by default with rust-lang/rust, so if binary size is a concern I think it would be better to handle this for every program rather than just those that don't opt-in. I'd also imagine that almost all programs would want to opt-in given how prevalent the concept of a current directory is. I feel like though this sort of a consequence of the wasi-libc strategy, which is that if we want to put stuff in wasm rather than WASI APIs that's inevitably going to make modules larger.

Perhaps if its very common then it could be opt-out instead, so tiny programs could link with `-D_WASI_EMULATED_CWD=0`?   Although maybe your weak linking idea is preferable?

libc-bottom-half/sources/chdir.c
```diff
@@ -0,0 +1,70 @@
...
+#include <limits.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <wasi/libc-find-relpath.h>
+#include <wasi/libc.h>
+
+char *__wasilibc_cwd = "/";
+static int __wasilibc_cwd_mallocd = 0;
```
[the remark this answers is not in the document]
sbc100: Since PWD is defined to be shared between threads better to leave it a non-TLS for now and include that error message.

Once we support threads we can most likely use some kind of atomic or lock here.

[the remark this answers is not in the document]
sunfishcode: Oops, yeah, `__wasilibc_cwd` is shared; I meant to add this comment to the `relative_buf` variables.

alexcrichton:
Ok so I'm having a really hard time wrangling weak symbols to do what I want. Inevitably they never seem to work for me...

First I refactored a bit where I defined a bunch of weak functions in places and then defined a copy of each function in `chdir.c`. @sunfishcode's sample program still, however, pulled in the symbols from `chdir.c`. I think the problem was that a reference to the symbols were made (e.g. `posix.c` depended on something weakly defined in `preopens.c`), and the linker found a strong version of the symbol in `chdir.c` and used that instead.

Next I refactored everything so the meat was defined in `chdir.c` and it was only referenced weakly from `posix.c` and `preopens.c`. The linker, however, still pulled in `chdir.c` since it seemed to see a reference to the weak symbol and attempted to resolve it.

Basically nothing I did was able to actually work. No matter what I did the linker kept pulling in the strong versions defined in `chdir.c`, even though `chdir` was not a referenced symbol. I feel like what I want is to place the `chdir.o` object into an archive, but I want the archive's symbol table to only list `chdir` for that object, nothing else defined in there. That way the linker, if it searched for symbols, wouldn't find `chdir.c`-defined-symbols unless `chdir` was referenced. I don't think this is an easy feature to implement, however.

I looked a bit though and of the 4k increase it looks like 3k is due to `realloc`. I changed all the `realloc` calls to `free` + `malloc` and the size increase is only 1k at that point.

I'm not really sure how much this is all worth it in these sense of any nontrivial application is almost surely going to have `realloc` in it already. Do y'all have other ideas for weak symbols though?

sbc100:
> The linker, however, still pulled in `chdir.c` since it seemed to see a reference to the weak symbol and attempted to resolve it.

This part is confusing me.  A weak reference to symbol in `chdir.o` should not cause the linker to pulling `chdir.o` if `chdir.o` is part of an archive file.

Are you declaring the function like this within `posix.c/preopens.c`?

```
__attribute__((weak)) void foo(void); 
```

?

To aid debugging you can try `-Wl,--trace-symbol=foo` or `-Wl,--verbose` or if you really want to see all the gory details you can do `-Wl,-mllvm,-debug-only=lld`

sbc100:
Oh.. I see the problem.  The confusion is between weak references and weakly defined symbol.  There are confusingly quite different and have completely different uses.

In this case you want weak references to strongly (normal) defined functions.

I think of it like this:

- weak reference: I don't pull objects out of archives.
- weak symbol: I'm ok with another definition of this symbol overriding me.

We really should not have re-used the word weak for these two different concepts.

alexcrichton:
In `preopens.c` I've got:

```c
int __wasilibc_find_relpath_alloc(
const char *path,
const char **abs,
char **relative,
size_t *relative_len,
int can_realloc
) __attribute__((weak));
````

and it's called like:

```c
    if (__wasilibc_find_relpath_alloc)
        return __wasilibc_find_relpath_alloc(path, abs_prefix, relative_path, &relative_path_len, 0); 
```

and in `chdir.c` I have the symbol actually defined, but when linking I get:

```
$ /opt/wasi-sdk/bin/clang foo.c --sysroot ./sysroot -o foo.wasm -Wl,--trace-symbol=__wasilibc_find_relpath_alloc
./sysroot/lib/wasm32-wasi/libc.a: lazy definition of __wasilibc_find_relpath_alloc
./sysroot/lib/wasm32-wasi/libc.a(preopens.o): reference to __wasilibc_find_relpath_alloc
./sysroot/lib/wasm32-wasi/libc.a(chdir.o): definition of __wasilibc_find_relpath_alloc
```

alexcrichton:
(I can gist the full patch too if it helps, it's just not 100% finished yet)

sbc100:
Can you try `--trace-symbol` chdir to see if anyone else is pulling that object in via that symbol?

alexcrichton:
That just prints out 

```
./sysroot/lib/wasm32-wasi/libc.a: lazy definition of chdir
./sysroot/lib/wasm32-wasi/libc.a(chdir.o): definition of chdir
```

alexcrichton:
Ok so poking around a bit more, if I move `chdir.o` to the end of `libc.a`, then `__wasilibc_find_relpath_alloc` is not pulled in unless `chdir` is used. If `chdir.o` comes before `preopens.o` (which it does on my system) then `__wasilibc_find_relpath_alloc` is pulled in no matter what.

@sbc100 is this perhaps a bug in the `wasm-ld`?

sbc100:
Certainly sounds very odd.    Are there any symbols that those object file both define?  Otherwise I can't think of reason why ordering within an archive could/should matter.   

I would love to take a look at a repro for this..  do you repro steps you could include here?

alexcrichton:
Ok I've pushed up a commit which I think does the weak symbol business. The locally of the example program increases from 38686 to 39021, an increase of 335 bytes.

@sbc100 there's some makefile tweaks here, and if those makefile tweaks are removed then [this example program](https://github.com/bytecodealliance/wasmtime/blob/main/docs/WASI-tutorial.md#from-c) increases 4k in size (pulls in `__wasilibc_find_relpath_alloc`)

libc-bottom-half/sources/chdir.c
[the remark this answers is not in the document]
alexcrichton: I've gone ahead and added a compile time error because this'll definitely all need quite a lot of synchronization.

sbc100:
wasm-ld fix:  https://reviews.llvm.org/D85567

alexcrichton:
Nice! I've updated comments in the `Makefile` about the build logic around `chdir.c` to reference that change.

libc-bottom-half/sources/chdir.c
```diff
@@ -0,0 +1,144 @@
...
+
+int chdir(const char *path)
+{
+    static char *relative_buf = NULL;
+    static size_t relative_buf_len = 0;
+
+    // Find a preopen'd directory as well as a relative path we're anchored
+    // from which we're changing directories to.
+    const char *abs;
+    int parent_fd = __wasilibc_find_relpath_alloc(path, &abs, &relative_buf, &relative_buf_len, 1);
```
sunfishcode, line 26: `__wasilibc_find_relpath_alloc` isn't declared here. Either this file should declare it like posix.c does, or there should be a declaration in one off the headers.

libc-bottom-half/sources/chdir.c
```diff
@@ -0,0 +1,144 @@
...
+    int copy_relative = strcmp(relative_buf, ".") != 0;
+    char *new_cwd = malloc(len + (copy_relative ? strlen(relative_buf) : 0));
+    if (new_cwd == NULL) {
+        errno = ENOMEM;
+        return -1;
+    }
+    new_cwd[0] = '/';
+    strcpy(new_cwd + 1, abs);
+    if (copy_relative)
+        strcpy(new_cwd + 1 + strlen(abs), relative_buf);
```
sunfishcode, line 52: It looks like this needs a `/` between the `abs` part and the `relative_buf` part. Does `relative_buf` always start with a `/` here? If so, it'd be good to comment that somewhere. If not, this should add a `/` between the paths.

sunfishcode (commented):
Looks good, thanks for putting together all the pieces for this! Just a few more comments:

libc-bottom-half/sources/preopens.c
```diff
@@ -71,15 +72,37 @@ static int resize(void) {
     return 0;
 }
 
+// Normalize an absolute path. Removes leading `/` and leading `./`, so the
+// first character is the start of a directory name. Additionally translate `.`
+// to the empty string.
```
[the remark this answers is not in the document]
alexcrichton: I actually thought that this was already how things were supposed to work where either `/foo` or `foo` worked fine! But it's true this would be good to document the default of `/` somewhere user-facing...

libc-bottom-half/sources/chdir.c
alexcrichton: Ah yes indeed, and I've fixed up another issue or two with this.

alexcrichton:
Sorry for the delay, but should be updated now!

sunfishcode (approved):
Looks good!

PR #219, opened 2020-10-05
CI: use $GITHUB_PATH,ENV rather than add-path, set-env

pchickey:
In accordance with this advisory it's recommended we move to a
different scheme of setting env vars and updating PATH:

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


sunfishcode:
Thanks!

sbc100 (commented):
Seems like a better way of doing things... magical strings in stdout always seemed a bit odd to me.

pchickey:
indeed, and in the process of fixing this in another repo i learned those magic strings even count in the commit message's first line, and crash CI if they're malformed :upside_down_face:

PR #221, opened 2020-10-08
Don't truncate directory cookies in `seekdir`/`telldir`

sunfishcode:
POSIX unfortunately requires the return type of `telldir` to be `long`, which is 32 bits on wasm32-wasi, so it isn't big enough to hold an actual directory cookie on WASI, which is 64 bits. For POSIX compatibility, add an indirection table in `dirent` so that `telldir` can return indices into the table rather than the actual cookies.

To accompany this change, add a `d_loc` field to the public `dirent`, which is similar to the traditional `d_off` field, except unlike `d_off` it's not tied to `seekdir`/`telldir` so it avoids the POSIX requirement that it have type `long`, and add a `__wasilibc_seekdir` interface to allow programs wishing to specialize for WASI to avoid the overhead of the indirection table.

This also adds a `d_namlen` field to the public `dirent` struct, which in addition to the changes above, allow it to match the layout of the raw WASI `dirent` struct, making the code simpler. This should also help avoid the need to call `strlen` on the `d_name` field in some situations.

### DIFF
--- a/expected/wasm32-wasi/defined-symbols.txt
+++ b/expected/wasm32-wasi/defined-symbols.txt
@@ -260,6 +260,7 @@ __wasilibc_open_nomode
 __wasilibc_openat_nomode
 __wasilibc_register_preopened_fd
 __wasilibc_rmdirat
+__wasilibc_seekdir
 __wasilibc_tell
 __wasilibc_unlinkat
 __wasm_call_dtors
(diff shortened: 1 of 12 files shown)


sunfishcode:
This has been a useful experiment in what it would take to support `seekdir` and `telldir` correctly, however these functions are so rarely used, and not a good fit for modern filesystems anyway, so it now seems better to just remove `seekdir` and `telldir` directly.

PR #226, opened 2020-12-08
Use override when adding to WASM_CFLAGS

abbec:
This seems to be the intended use case of having `WASM_CFLAGS ?=` at the
top of the Makefile. Otherwise, if WASM_CFLAGS is set on the command
line, flags like `--sysroot` are not added (See
https://www.gnu.org/software/make/manual/make.html#Override-Directive).

Now, with `override` the values are instead appended to whatever is
provided on the command line, or the default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,21 +184,21 @@ LIBC_TOP_HALF_ALL_SOURCES = \
     $(shell find $(LIBC_TOP_HALF_SOURCES) -name \*.c)
 
 # Set the target.
-WASM_CFLAGS += --target=$(TARGET_TRIPLE)
+CFLAGS = $(WASM_CFLAGS) --target=$(TARGET_TRIPLE)
 # WebAssembly floating-point match doesn't trap.
 # TODO: Add -fno-signaling-nans when the compiler supports it.
-WASM_CFLAGS += -fno-trapping-math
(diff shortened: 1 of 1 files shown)


sbc100:
It looks like there are a few other places in this file where we do WASM_CFLAGS.  See `MUSL_PRINTSCAN_OBJS` on 338. 

I wonder if it would make more sense to rename this to variable to just `CFLAGS` and initialize it as:

`CFLAGS = $(WASM_CFLAGS) --sysroot="$(SYSROOT)"`

abbec:
Ah, yeah you are right. Yeah it seems like a better solution to just use another flag for it instead, I just thought it was kinda unexpected behavior that overriding `WASM_CFLAGS` on the command line overwrote the value completely, not just the defaults.

abbec:
I made sure to put `override` in all places at least but I am also happy to rename the variable if we want that.

sbc100:
I find the second variable approach to be a little more aesthetically pleasing that scattering override everywhere... and it should help reduce the excessive line lengths.   WDYT?

abbec:
Yeah, I prefer that too, will fix! 🙂

abbec:
Done!

sbc100 (approved):

PR #228, opened 2021-01-25
Makefile: ignore __indirect_function_table symbol added in LLVM 12

rojamd:
Per https://reviews.llvm.org/D92215#2505592, this symbol should be
ignored when verifying undefined-symbols.txt.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -449,7 +449,7 @@ finish: startup_files libc
 	for undef_sym in $$("$(WASM_NM)" --undefined-only "$(SYSROOT_LIB)"/libc.a "$(SYSROOT_LIB)"/libc-*.a "$(SYSROOT_LIB)"/*.o \
 	    |grep ' U ' |sed 's/.* U //' |LC_ALL=C sort |uniq); do \
 	    grep -q '\<'$$undef_sym'\>' "$(SYSROOT_SHARE)/defined-symbols.txt" || echo $$undef_sym; \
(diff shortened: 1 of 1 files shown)


sbc100 (commented):
For compatibility, we are working on fixing LLVM so that symbol only gets generated when reference types is enabled.   So this change would no longer be needed.   Should we just wait for that to land and avoid this change?

rojamd:
> For compatibility, we are working on fixing LLVM so that symbol only gets generated when reference types is enabled. So this change would no longer be needed. Should we just wait for that to land and avoid this change?

The 12.0.0 release branch will be created tomorrow. Do you think you'll be able to land that change and request a backport before the release cutoff? If I'm understanding https://reviews.llvm.org/D95114 correctly, that would be at RC2 or roughly Feb 23. If that's doable then it should be fine to drop this change.

sbc100:
Oh crap i was not aware of that!   We certainly need to make sure that change doesn't make the final release.    I will coordinate with @wingo to get this fixed.

sbc100:
Actually I think I was confused and the problematic CL has already been reverted: https://reviews.llvm.org/D95005.   So llvm tot should already no need this.

sbc100:
Second revert that actually clears this issue up: https://reviews.llvm.org/D95420

rojamd:
With the changes upstream, this is no longer needed.

PR #230, opened 2021-02-04
AT_FDCWD support.

sunfishcode:
This forward-ports #54 to current main.


sunfishcode:
A brief description of what this is:

>    AT_FDCWD is a special constant in POSIX that can be passed to *at
    functions to indicate the current working directory. Since the
    current working directory is emulated in wasi libc, add emulated
    AT_FDCWD support as well.

Makefile
```diff
@@ -188,6 +188,10 @@ CFLAGS = $(WASM_CFLAGS) --target=$(TARGET_TRIPLE)
 # WebAssembly floating-point match doesn't trap.
 # TODO: Add -fno-signaling-nans when the compiler supports it.
 CFLAGS += -fno-trapping-math
+# We should have proper declarations for everything.
+CFLAGS += \
+    -Werror=missing-declarations \
+    -Werror=implicit-function-declaration \
```
sbc100, line 194: Can we instead use `-Werror` and disabled the warngins that are not fixable or not part in code that we maintain?

libc-bottom-half/cloudlibc/src/libc/fcntl/openat.c
```diff
@@ -11,156 +11,61 @@
 #include <fcntl.h>
 #include <string.h>
 
-#ifdef __wasilibc_unmodified_upstream // fstat
-static_assert(O_APPEND == __WASI_FDFLAG_APPEND, "Value mismatch");
-static_assert(O_DSYNC == __WASI_FDFLAG_DSYNC, "Value mismatch");
-static_assert(O_NONBLOCK == __WASI_FDFLAG_NONBLOCK, "Value mismatch");
-static_assert(O_RSYNC == __WASI_FDFLAG_RSYNC, "Value mismatch");
-static_assert(O_SYNC == __WASI_FDFLAG_SYNC, "Value mismatch");
-#else
```
sbc100, line 20: There removals of the ifdef blocks seem unrelated?    Could they land separately?

Makefile
sbc100: Also maybe this should be its own change.

sunfishcode:
The warnings and ifdef changes are now landed as separate chnages, and this is now rebased on top of them.

sbc100 (approved):

PR #231, opened 2021-02-05
Remove __wasilibc_unmodified_upstream markers from libc-bottom-half.

sunfishcode:
We've already started removing this; this just removes all remaining
ones under the libc-bottom-half directory.

These markers were originally intended to help track upstream changes,
however in practice they created a lot of clutter and weren't that
helpful. And now, upstream cloudlibc is no longer active.


sbc100 (approved):
Great!

PR #234, opened 2021-02-17
Update to the next version of the `witx` crate

alexcrichton:
* Generate adapter functions instead of simply a header file to have a
  place where adapter code can go.
* Implement adapters in terms of the instructions that the `witx` crate
  tells us about.
* Update the interface of functions to what `witx` expects, notably
  string arguments are now only taken as `char*` and `strlen` happens in
  the adapter function.
* Update defined/predefined/undefined symbol lists for types that have
  been updated.

Some precise generated code has changed but the actual APIs should all
be the same except for the change to not take the length of the string
in the raw WASI call, since idiomatically C doesn't pass the length of
strings around.

Eventually it's expected that the shim functions, while sometimes not
necessary today, will implement more checks and more conversions as
necessary for new APIs.


alexcrichton:
Note that this is dependant on https://github.com/WebAssembly/WASI/pull/395, so this should probably wait for that to get settled first.

libc-bottom-half/cloudlibc/src/libc/unistd/symlinkat.c
```diff
@@ -10,8 +10,7 @@
 #include <unistd.h>
 
 int __wasilibc_nocwd_symlinkat(const char *path1, int fd, const char *path2) {
-  __wasi_errno_t error =
-      __wasi_path_symlink(path1, strlen(path1), fd, path2, strlen(path2));
+  __wasi_errno_t error = __wasi_path_symlink(path1, fd, path2);
```
sbc100, line 13: Given that these are now library calls its makes me wonder if we should continue to use the `__` prefix.   Perhaps these should be just `wasi_path_symlink` and then the real imports could be `__wasi_path_symlink`?    Would most likely mean changing all the types too.

expected/wasm32-wasi/undefined-symbols.txt
```diff
@@ -19,48 +19,50 @@ __subtf3
...
+__wasi_real_path_symlink
+__wasi_real_path_unlink_file
+__wasi_real_poll_oneoff
+__wasi_real_proc_exit
+__wasi_real_proc_raise
+__wasi_real_random_get
+__wasi_real_sched_yield
+__wasi_real_sock_recv
+__wasi_real_sock_send
+__wasi_real_sock_shutdown
```
sbc100, line 66: Why is this list two items longer?   (github diff isn't telling me which two are the new ones).

libc-bottom-half/headers/public/wasi/api.h
```diff
@@ -1,6 +1,6 @@
 /**
  * THIS FILE IS AUTO-GENERATED from the following files:
- *   typenames.witx, wasi_snapshot_preview1.witx
+ *   wasi_snapshot_preview1.witx
```
sbc100, line 3: At I see the boilerplate message already doesn't say the tool name, maybe fine to leave that as a followup.

libc-bottom-half/headers/public/wasi/api.h
```diff
@@ -502,160 +502,157 @@ typedef uint64_t __wasi_rights_t;
...
  */
-#define __WASI_RIGHTS_PATH_FILESTAT_GET (UINT64_C(262144))
+#define __WASI_RIGHTS_PATH_FILESTAT_GET ((__wasi_rights_t)(1 << 18))
 
 /**
  * The right to change a file's size (there is no `path_filestat_set_size`).
  * If `path_open` is set, includes the right to invoke `path_open` with `oflags::trunc`.
  */
-#define __WASI_RIGHTS_PATH_FILESTAT_SET_SIZE (UINT64_C(524288))
+#define __WASI_RIGHTS_PATH_FILESTAT_SET_SIZE ((__wasi_rights_t)(1 << 19))
```
sbc100, line 609: Does this change make it any harder to to combine these things without compiler warnings?  e.g. `WASI_XX | WASI_YYY`.  (I guess it shouldn't as long as the all the `__wasi_rights_t` type things are basic integer types).

libc-bottom-half/headers/public/wasi/api.h
```diff
@@ -1625,458 +1589,296 @@ _Static_assert(_Alignof(__wasi_prestat_u_t) == 4, "witx calculated union align")
...
-     */
-    __wasi_size_t *argv_buf_size
-) __attribute__((
-    __import_module__("wasi_snapshot_preview1"),
-    __import_name__("args_sizes_get"),
-    __warn_unused_result__
-));
-
+    __wasi_size_t *retptr0,
+    __wasi_size_t *retptr1
```
sbc100, line 1602: Looks like we lost some information here about what these pointers mean.

sbc100 (commented):
I can't read any of the rust but the output looks good and the direction seems good.

libc-bottom-half/sources/__wasilibc_real.c
```diff
@@ -0,0 +1,652 @@
+/**
+ * THIS FILE IS AUTO-GENERATED from the following files:
+ *   wasi_snapshot_preview1.witx
+ */
+
+#include <wasi/api.h>
+#include <string.h>
+
+int32_t __wasi_real_args_get(int32_t arg0, int32_t arg1) __attribute__((
```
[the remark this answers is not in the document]
alexcrichton: Sure yeah! I went ahead and added "__imported" on the front

expected/wasm32-wasi/undefined-symbols.txt
alexcrichton: The two new items are proc_raise and set_rights. I believe the reason this is showing up now is that wasi-libc never called either of those functions internally, so they were just defined in the header. By having a shim though for all these functions they're all consumed now so they're all showing up as undefined symbols.

libc-bottom-half/cloudlibc/src/libc/unistd/symlinkat.c
alexcrichton: I'm happy to implement either way, but @sunfishcode would probably be best to answer the question (I'm not familiar enough with C idioms to know what best to do here).

libc-bottom-half/headers/public/wasi/api.h
alexcrichton: I think this should be the same because these two definitions are intended to be equivalent. The only difference here is that the type name is used for the cast rather than the unwrapped type which is `UINT64_C` in this case.

libc-bottom-half/headers/public/wasi/api.h
alexcrichton: Yeah unfortunately the documentation got shuffled a bit since each individual parameter is no longer documented in the source file, although I tried to make sure that the documentation was at least preserved somewhat in the `@return` section for the function

sbc100: Any why we can keep the argc and argv_buf_size names some how?

Can these be "retptr_argc" and "retptr_argv_buf_size" perhaps?    retptr0 and retptr1 are not very useful names.

libc-bottom-half/headers/public/wasi/api.h
```diff
@@ -502,160 +509,157 @@ typedef uint64_t __wasi_rights_t;
...
-#define __WASI_RIGHTS_POLL_FD_READWRITE (UINT64_C(134217728))
+#define __WASI_RIGHTS_POLL_FD_READWRITE ((__wasi_rights_t)(1 << 27))
 
 /**
  * The right to invoke `sock_shutdown`.
  */
-#define __WASI_RIGHTS_SOCK_SHUTDOWN (UINT64_C(268435456))
-
-_Static_assert(sizeof(__wasi_rights_t) == 8, "witx calculated size");
-_Static_assert(_Alignof(__wasi_rights_t) == 8, "witx calculated align");
```
pchickey, line 658: we lost these `_Static_assert`s on witx `Constant`s

libc-bottom-half/headers/public/wasi/api.h
```diff
@@ -1277,15 +1263,12 @@ typedef union __wasi_subscription_u_u_t {
     __wasi_subscription_fd_readwrite_t fd_write;
 } __wasi_subscription_u_u_t;
 typedef struct __wasi_subscription_u_t {
-    __wasi_eventtype_t tag;
+    uint8_t tag;
```
pchickey, line 1266: this got less precise, even though __wasi_eventtype_t is typedef'd to uint8_t.

libc-bottom-half/cloudlibc/src/libc/unistd/symlinkat.c
sunfishcode: The `__` prefix continues to make sense :-). For example, users can call WASI APIs directly, and may have their own `wasi_*` wrappers around things, so the `__` prefix here ensures that even if they do that, they can also link in libc at the same time.

sunfishcode: That said, if we wanted to make it easier for people to write code that calls directly into WASI, one thing we could do would be to provide weak aliases named `wasi_*` that point to the `__wasi_` symbols.

sbc100: I just think maybe the `__` prefix is not the right way to carve of our own namespace.   Of course this is unrelated to this PR, but I think `__` is normally reserved for toolchain internal stuff isn't it?  

Again, unrelated to this PR.

libc-bottom-half/headers/public/wasi/api.h
sbc100: SGTM

libc-bottom-half/cloudlibc/src/libc/unistd/symlinkat.c
sunfishcode: The C standard just says "reserve for any use". In practice, musl, glibc, and other libc impls commonly use `__` for their own purposes.

libc-bottom-half/headers/public/wasi/api.h
alexcrichton: Unfortuantely this is mostly due to a change in the `*.witx` file itself -- https://github.com/WebAssembly/WASI/commit/ef8c1a53feb2dfb763d4ea5c7d9e0a0126b45579#diff-2f2ab5a9d79cd5c57a981bf769e772aa60405857803bcdbde3c3e8179315abb8L24-R26

Where previously there were separate `result` variables it's now modeled more closely with the intended interface type, which is a `Result<(usize, usize), Errno>`. This means that there's no way really to auto-generate the names as before since from the source it's just "return pointer for first tuple element" and the second one.

libc-bottom-half/headers/public/wasi/api.h
alexcrichton: That's true, yeah, and is due to tags no longer being modeled as pointers to an `enum` but rather just simply having a particular bit-width as desired. This means that the documentation is lost a bit but at the bindings layer everything should be the same since `__wasi_eventtype_t` is a typedef to `uint8_t` anyway.

libc-bottom-half/headers/public/wasi/api.h
alexcrichton: Oh that was mostly just because I added a special case somewhere and didn't want to add back the static asserts. If desired, however, I can add this back in.

pchickey: ok. its fine, those are most useful for structured types, atoms should be ok.

libc-bottom-half/headers/public/wasi/api.h
pchickey: sgtm

libc-bottom-half/headers/public/wasi/api.h
pchickey: There's a trade-off here where we could get names for these values back by returning a record rather than a tuple, but that would break the ABI today (records always turn into a C struct, which gets passed by one pointer rather than multiple). Its a little but unfortunate but I think losing those names is OK for now.

pchickey (approved):

PR #241, opened 2021-03-22
Emulate `clock`, `times`, and `getrusage` using the monotonic clock.

sunfishcode:
Upcoming WASI snapshots omit the `PROCESS_CPUTIME` clock, since WASI has
no inherent concept of processes, and since implementations which don't
use a process for each instance don't have a way to implement it
efficiently.

However, `clock`, `times`, and `getrusage` are useful functions, so
provide optional emulated version of them, using the `MONOTONIC` clock.
This means these implementations will measure not just the program's
own CPU time, but also time spent suspended while other programs are
running.

Due to this difference in behavior, put these implementations behind
a flag. Users must pass `-D_WASI_EMULATED_PROCESS_CLOCK` and link with
`-lwasi-emulated-process-clocks` to enable them.


pchickey (approved):

expected/wasm32-wasi/defined-symbols.txt
```diff
@@ -442,7 +442,6 @@ cimagl
 clearenv
 clearerr
 clearerr_unlocked
-clock
```
sbc100, line 445: Do we not want a superset of all symbols here?

sbc100 (approved):

expected/wasm32-wasi/defined-symbols.txt
sunfishcode: This diff happens because `clock` is moved to `libwasi-emulated-process-clocks.a` instead of `libc.a`.

I guess we could change `defined-symbols.txt` to include symbols from all the libraries, and not just `libc.a`. I'll look into that.

sbc100: Yup, that is what I was suggesting.  Nice!

zeux:
It would be nice to be able to simply use clock() without having to specify additional defines and libraries... On Windows, clock() already accounts for the amount of time spent waiting, so I'm not sure that the value of the extra warnings/errors in absence of the said define justify themselves...

PR #242, opened 2021-04-02
Enable support for `utimes` and `futimesat`.

sunfishcode:
These can both be implemented in terms of `utimensat`.

### DIFF
--- a/libc-top-half/musl/src/stat/futimesat.c
+++ b/libc-top-half/musl/src/stat/futimesat.c
@@ -2,7 +2,9 @@
 #include <sys/time.h>
 #include <sys/stat.h>
 #include <errno.h>
+#ifdef __wasilibc_unmodified_upstream // WASI has no syscall
 #include "syscall.h"
+#endif
 
 int __futimesat(int dirfd, const char *pathname, const struct timeval times[2])
 {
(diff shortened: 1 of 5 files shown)


sunfishcode:
This is ready to go, and just needs a review.

libc-top-half/musl/src/stat/futimesat.c
```diff
@@ -11,7 +13,12 @@ int __futimesat(int dirfd, const char *pathname, const struct timeval times[2])
 		int i;
 		for (i=0; i<2; i++) {
 			if (times[i].tv_usec >= 1000000ULL)
+#ifdef __wasilibc_unmodified_upstream // WASI has no syscall
 				return __syscall_ret(-EINVAL);
```
sbc100, line 17: I wonder if this a common pattern if it would be worth at least defining `__syscall_ret` macro?

sbc100 (approved):

libc-top-half/musl/src/stat/futimesat.c
sunfishcode: I'm not necessarily opposed to it, but it feels a little awkward, because either:
 - It introduces the Linux-specific convention of using negative values in [-4095,0) to represent errors, or
 - It doesn't use this, which risks having it behave differently than the code calling it expects.

PR #243, opened 2021-04-02
Implement emulated support for `getpid`.

sunfishcode:
Add a simple `getpid` emulation function which just returns a fixed
value, for applications that just use it for logging.


sbc100 (approved):
I won't if this fine grained approach to emulation is scale-able.     

One alternative would be have a single `-lwasi-compat` and `-DWASI_COMPAT` ?

this change lgtm though

sunfishcode:
It kind of feels right to keep them separate right now, because they imply different kinds of emulation. `getpid` is just "here's a bogus value". `mmap` is "i'll actually read the file for you, but in a different way". And so on. But yeah, I could see a "enable all" mode becoming useful.

PR #247, opened 2021-06-15
chdir: fixing a malloc'ed buffer that was not large enough (possible mem corruption)

gerdstolpmann:
I started using `chdir` in my code, and the effect was that it occasionally inserted random bytes into paths (e.g. opening `/abc/def` would try to open `/abc\xab/def`). This PR fixes the problem - it turns out the buffer for the current working directory was not always large enough.

### DIFF
--- a/libc-bottom-half/sources/chdir.c
+++ b/libc-bottom-half/sources/chdir.c
@@ -46,7 +46,7 @@ int chdir(const char *path)
     size_t len = strlen(abs) + 1;
     int copy_relative = strcmp(relative_buf, ".") != 0;
     int mid = copy_relative && abs[0] != 0;
-    char *new_cwd = malloc(len + (copy_relative ? strlen(relative_buf) + mid: 0));
+    char *new_cwd = malloc(len + (copy_relative ? strlen(relative_buf) + mid: 0)+1);
     if (new_cwd == NULL) {
         errno = ENOMEM;
         return -1;
@@ -54,7 +54,7 @@ int chdir(const char *path)
     new_cwd[0] = '/';
     strcpy(new_cwd + 1, abs);
     if (mid)
-        new_cwd[strlen(abs) + 1] = '/';
+        new_cwd[len] = '/';
     if (copy_relative)
         strcpy(new_cwd + 1 + mid + strlen(abs), relative_buf);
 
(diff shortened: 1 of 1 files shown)


libc-bottom-half/sources/chdir.c
```diff
@@ -46,15 +46,15 @@ int chdir(const char *path)
     size_t len = strlen(abs) + 1;
     int copy_relative = strcmp(relative_buf, ".") != 0;
     int mid = copy_relative && abs[0] != 0;
-    char *new_cwd = malloc(len + (copy_relative ? strlen(relative_buf) + mid: 0));
+    char *new_cwd = malloc(len + (copy_relative ? strlen(relative_buf) + mid: 0)+1);
```
gerdstolpmann, line 49: This is the fix.

libc-bottom-half/sources/chdir.c
```diff
@@ -95,9 +95,10 @@ static const char *make_absolute(const char *path) {
     int need_slash = __wasilibc_cwd[cwd_len - 1] == '/' ? 0 : 1;
     size_t alloc_len = cwd_len + path_len + 1 + need_slash;
     if (alloc_len > make_absolute_len) {
-        make_absolute_buf = realloc(make_absolute_buf, alloc_len);
-        if (make_absolute_buf == NULL)
+        char *tmp = realloc(make_absolute_buf, alloc_len);
+        if (tmp == NULL)
             return NULL;
+        make_absolute_buf = tmp;
```
gerdstolpmann, line 101: Also figured this potential problem when reading the code carefully - in the out of mem case, `make_absolute_buf` would be set to `NULL` but `make_absolute_len` would be still positiive, leading to mem corruption the next the function will be called.

libc-bottom-half/sources/chdir.c
```diff
@@ -46,15 +46,15 @@ int chdir(const char *path)
     size_t len = strlen(abs) + 1;
```
sunfishcode, line 46: Would would you think about renaming this to `abs_len` and removing the `+ 1`, folding the `+ 1` into the calculations below as needed? Then we could have code that does `1 + abs_len + ... + 1` which would make it a little easier to see how it relates to the `__wasilibc_cwd` comment above.

Also, there's another redundant `strlen(abs)` below which could use this.

sunfishcode (commented):
Thanks for the report, and the fix! Your analysis here looks right to me. I just have one comment about a possible way of organizing the code which might make it easier to inspect:

libc-bottom-half/sources/chdir.c
gerdstolpmann: makes sense

kateinoigakukun:
@sunfishcode Is there any blocker to merge this PR? We also faced this issue in Ruby porting.

sunfishcode:
Yes; thanks for the ping here. This looks good to me!

Issue #244, opened 2021-04-15
Make dlmalloc optional


aykevl:
This is perhaps a more minimal version of issue #94.

I would like to use a custom allocator because the environment I'm using wasi-libc in (TinyGo) already implements a heap. I think the easiest way to get that working is to not compile dlmalloc as part of wasi-libc.

Any idea how this might be implemented? I'm hitting issues with the file `undefined-symbols.txt` when I comment out the wrapper functions in `dlmalloc/src/dlmalloc.c`.

sunfishcode:
Other parts of libc depend on having a `malloc` and similar functions; does your allocator provide `malloc`-compatible symbols?

aykevl:
I found that I can simply define the functions in my application and they will override the functions defined in the library. I already knew this, just didn't realize it would apply here.

Here is how I did it: https://github.com/tinygo-org/tinygo/pull/1992

> Other parts of libc depend on having a `malloc` and similar functions; does your allocator provide `malloc`-compatible symbols?

Sort of, I've implemented the necessary wrapper functions in the PR above. It's a GC and `free` doesn't currently do anything, but that's fine.

PiotrSikora:
Could we reopen this isssue? I think the ability to link another allocator would be useful.

Notably, using [`wee_alloc`](https://github.com/rustwasm/wee_alloc), [`mimalloc`](https://github.com/purpleprotocol/mimalloc_rust) or another `#[global_allocator]` in Rust works fine for `wasm32-unknown-unknown` targets, but `wasi-libc` always pulls `dlmalloc` when building for `wasm32-wasi` target, which results in multiple allocators in the Wasm bytecode, reintroducing large codebase which removal was one of the main reasons (small code size) for creating `wee_alloc` in the first place.

sbc100:
I would hope that the approach @aykevl uses should be enough to completely override/replace dlmalloc, no?    Just providing your `-lmymalloc` library on the command line should be enough I think?

sunfishcode:
Yes, `-lmymalloc` should work. That said, it'd be great for wasi-libc to have alternative malloc options available. `dlmalloc` is several kilobytes of code.

PiotrSikora:
[...]

Issue #94, opened 2019-08-29
Investigate mimalloc


sunfishcode:
It'd be nice to provide alternative malloc implementations, with different code size vs performance tradeoffs.

See the comment [here](https://github.com/CraneStation/wasi-libc/pull/89#issuecomment-513409067) about mimalloc.

michaelvoronov:
Hi!

I am interested in layout of Wasm linear memory on execution environment. As far as I know, with `wasm-ld` it should look as follows:

[ |`0` ... data ... | ...  stack ... | (`__heap_base`) ... heap ... (`initial`) ...  (`maximum`)| ], 

where `__heap_base` is a place where heap could be started and `initial` usually equals ceil(`__heap_base`) to 2**16, so `initial` >= `__heap_base` (or even >>).

Currently, wasi-libc uses dlmalloc with disabled mmap. And this dlmalloc starts with extending its top chunk with `memory_grow`. Execution environments execute this call as a memory "allocating" from `initial` to `maximum`. So it turns out that if allocator isn't explicitly support `__heap_base` space between `initial` and `__heap_base` isn't used. And this is up to 2**16 - quite a lot. And current wasi-libc allocator doesn't support `__heap_base`. Moreover, wasm-ld in clang 9.0.0 doesn't export `__heap_base` and `data_ends`. Maybe I am wrong somewhere in this.

What do you think about it, does it make sense for WASI? Shall old or new allocator support `__heap_base`?

sunfishcode:
Yes, it does make sense to save this memory.  I expect this could be done within the existing `dlmalloc` allocator, though I haven't looked at specifics. It would be useful for alternate allocators to do too.

wasm-ld still does support `__heap_base` and `__data_end`; it just doesn't emit them if they're not used.

michaelvoronov:
I can prepare a PR for `dlmalloc` this week.

michaelvoronov:
@sunfishcode, I have made a fix, but It is inside dlmalloc `sys_alloc` and I wanted to test it before making a PR somehow. Does wasi-libc have a test suite?

sunfishcode:
Great! For a testsuite, I have started working on setting up libc-test [here](https://github.com/CraneStation/wasi-libc-test) -- for licensing issues we need to keep it in a separate repo. It's admittedly not a lot yet, but it's a start.

michaelvoronov:
[...]

PR #250, opened 2021-07-13
Add build flag to select malloc implementation

aykevl:
Add a build flag MALLOC_IMPL that can be set to dlmalloc or none (defaulting to dlmalloc) which controls the malloc implementation to use. The dlmalloc option is the same as before, but selecting none removes dlmalloc from the libc build.

This flag replaces the BUILD_DLMALLOC flag, which never worked and thus can be removed without breaking any builds. By switching to MALLOC_IMPL, there is a clear path towards a different heap implementation, such as mimalloc.

See https://github.com/WebAssembly/wasi-libc/issues/94 and https://github.com/WebAssembly/wasi-libc/issues/244 for context.

I don't really like the long grep lines, if somebody knows of a better way to remove these symbols I'd like to fix it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,16 +10,12 @@ SYSROOT ?= $(CURDIR)/sysroot
 INSTALL_DIR ?= /usr/local
 # single or posix
 THREAD_MODEL ?= single
+# dlmalloc or none
+MALLOC_IMPL ?= dlmalloc
 # yes or no
-BUILD_DLMALLOC ?= yes
 BUILD_LIBC_TOP_HALF ?= yes
 # The directory where we're store intermediate artifacts.
 OBJDIR ?= $(CURDIR)/build
-
-# Check dependencies.
-ifneq ($(BUILD_DLMALLOC),yes)
-$(error build currently depends on BUILD_DLMALLOC=yes)
-endif
 
 # Variables from this point on are not meant to be overridable via the
 # make command-line.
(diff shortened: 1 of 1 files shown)


sbc100:
What didn't work about `BUILD_DLMALLOC` exactly?  

Why not just fix that option for now, given that we don't have any other malloc options yet? 

If we are going to include options, I wonder if we should include musl's own malloc as a possible option?  Maybe there is no reason to do that though..

I wonder if we can just skip the symbol tests if malloc is not desired, or if think its valuable we could have different `expected.txt` file for each libc variant that we support?

sunfishcode:
Skipping the symbol test if malloc is not desired sounds reasonable to me here.

aykevl:
> What didn't work about `BUILD_DLMALLOC` exactly?

The new `MALLOC_IMPL` is just a more flexible replacement for `BUILD_DLMALLOC` so it would require the exact same changes as I needed to do for `BUILD_DLMALLOC`. You can see this in the diff. Presumably the symbol check is the reason why this flag was disabled.

> Why not just fix that option for now, given that we don't have any other malloc options yet?

To be forward compatible with other allocators. Yes, I could most certainly make this PR in a way that it fixes `BUILD_DLMALLOC` (which is currently defunct and presumably unused for that reason) but then people (like me) might start depending on the flag and require a change once more allocators are added.

If desired I can certainly change this PR to use the flag `BUILD_DLMALLOC` instead of `MALLOC_IMPL` but explicitly providing 'none' or 'dlmalloc' seems more flexible to me.

> If we are going to include options, I wonder if we should include musl's own malloc as a possible option? Maybe there is no reason to do that though..

Sure, that might be possible (I didn't look at the musl allocator and whether it can run on wasm). But it seems better to do that separately, when someone actually needs it.

> I wonder if we can just skip the symbol tests if malloc is not desired, or if think its valuable we could have different `expected.txt` file for each libc variant that we support?

Sure, that step can be skipped in that case. A different defined-symbols.txt and undefined-symbols.txt for each malloc seems like a bad idea to me for maintenance and in case another such flag will be added in the future.

I'll investigate skipping this step.

aykevl:
Updated, skipping the metadata verification step when a non-standard heap allocator is used.

sunfishcode:
Thanks! This looks good.

PR #251, opened 2021-07-27
Fix compilation with clang trunk

keithw:
This PR adds `-Wno-unused-but-set-variable` to CFLAGS (required by [strfmon.c:12](https://github.com/WebAssembly/wasi-libc/blob/main/libc-top-half/musl/src/locale/strfmon.c#L12)) and adds two encoding-related preprocessor macros to [predefined-macros.txt](https://github.com/WebAssembly/wasi-libc/blob/main/expected/wasm32-wasi/predefined-macros.txt) (added in clang trunk by https://github.com/llvm/llvm-project/commit/701d70d4c25c4e02b303ba6dee1495708496f615). With these changes, wasi-libc compiles with the current clang trunk.

### DIFF
--- a/expected/wasm32-wasi/predefined-macros.txt
+++ b/expected/wasm32-wasi/predefined-macros.txt
@@ -2457,7 +2457,6 @@
 #define __FLT_DENORM_MIN__ 1.40129846e-45F
 #define __FLT_DIG__ 6
 #define __FLT_EPSILON__ 1.19209290e-7F
-#define __FLT_EVAL_METHOD__ 0
 #define __FLT_HAS_DENORM__ 1
 #define __FLT_HAS_INFINITY__ 1
 #define __FLT_HAS_QUIET_NAN__ 1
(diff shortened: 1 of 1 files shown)


keithw:
One question might be -- would wasi-libc rather just `grep -v` these new macros out (as was done for __FLT16_ in https://github.com/WebAssembly/wasi-libc/blob/main/Makefile#L538), if the goal is to target multiple versions of clang? It seems like clang will probably keep adding new predefined macros over time, so trying to match them exactly across multiple compiler versions might be tough.

sunfishcode:
wasi-libc does intend to support multiple versions of clang, but I don't think we need predefined-macros.txt to support multiple versions of clang.

The main purpose of the predefined-macros.txt check is to be careful about macro changes that imply ABI changes, and to be careful about accidentally introducing macros that pollute the user namespace. Checking those on the latest supported version of clang is enough to catch the main problems. So yeah, going forward, I think it makes sense to stop trying to add grep lines for everything, and start just disabling the whole test if we're on an older version of clang.

keithw:
Updated to compile the current wasi-libc trunk on the current clang trunk

sunfishcode:
I believe this issues was fixed in #257; please reopen if there are any further issues.

PR #255, opened 2021-09-01
Allow user to provide CC wrapper

### LINKED ISSUE #253: Can't utilize ccache with a custom compiler
Typically one might expect to be able to supply `make WASM_CC="ccache <path to clang>"` to override the compiler & use ccache, but the use of double-quotes invoking the compiler (`"$(WASM_CC)"`) inhibits this. The double-quotes should be removed or perhaps a new `WASM_CC_WRAPPER` introduced that can be used to inject ccache?

vlovich:
If using a custom toolchain not on the path, explicitly invoking
ccache/sccache is required (also easier to debug that the right compiler
is being invoked). Fixes #253 .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 # These variables are specifically meant to be overridable via the make
 # command-line.
 WASM_CC ?= clang
-WASM_NM ?= $(patsubst %clang,%llvm-nm,$(WASM_CC))
-WASM_AR ?= $(patsubst %clang,%llvm-ar,$(WASM_CC))
+WASM_NM ?= $(patsubst %clang,%llvm-nm,$(filter-out ccache sccache,$(WASM_CC)))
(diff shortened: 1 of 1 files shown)


sbc100:
I can see that this works, but I do wonder why we have to do this when other build systems such as automake (and indeed most hand-rolled) Makefiles's don't provide this and assume the user can include ccache in the CC/CXX variable itself.

sbc100:
> I'm fine with that too. I just assumed that the motivation for quoting `WASM_CC` was to allow for spaces in which case this would be the only way to do that.

I think we can just get rid of the quotes then perhaps.

Looking at autoconf/automake it looks like spaces are OK in the CC variable.

When I run `CC="ccache /usr/bin/gcc" ./configure` the generated Makefile contains `CC = ccache /usr/bin/gcc` and it seems to work just fine.

sbc100 (approved):

vlovich:
Is there anything else that needs to be done here to merge?

PR #263, opened 2021-12-04
Use WebAssembly bulk memory opcodes

### LINKED ISSUE #262: Use WebAssembly bulk memory opcodes for memcpy/memset?
WebAssembly bulk memory opcodes have shipped for major browsers for some time. Does it make sense to use that for `memcpy`/`memset` functions in `wasi-libc`? If there are no objections in principle, I'd like to try opening a PR for this.

TerrorJack:
Closes #262.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,13 @@ BUILD_LIBC_TOP_HALF ?= yes
 # The directory where we're store intermediate artifacts.
 OBJDIR ?= $(CURDIR)/build
 
+# When the length is no larger than this threshold, we consider the
+# overhead of bulk memory opcodes to outweigh the performance benefit,
+# and fall back to the original musl implementation. See
+# https://github.com/WebAssembly/wasi-libc/pull/263 for relevant
+# discussion
+BULK_MEMORY_THRESHOLD ?= 32
+
 # Variables from this point on are not meant to be overridable via the
 # make command-line.
 
@@ -393,6 +400,9 @@ $(MUSL_PRINTSCAN_NO_FLOATING_POINT_OBJS): CFLAGS += \
 # https://github.com/llvm/llvm-project/issues/52618 is resolved
 $(BULK_MEMORY_OBJS): CFLAGS += \
         -mbulk-memory
+
+$(BULK_MEMORY_OBJS): CFLAGS += \
+        -DBULK_MEMORY_THRESHOLD=$(BULK_MEMORY_THRESHOLD)
 
 $(LIBWASI_EMULATED_SIGNAL_MUSL_OBJS): CFLAGS += \
 	    -D_WASI_EMULATED_SIGNAL
--- a/Makefile
+++ b/Makefile
@@ -211,7 +211,7 @@ CFLAGS += -Wall -Wextra -Werror \
 
 # Configure support for threads.
 ifeq ($(THREAD_MODEL), single)
-CFLAGS += -mthread-model single
+CFLAGS += -mthread-model single -ftls-model=local-exec
 endif
 ifeq ($(THREAD_MODEL), posix)
 CFLAGS += -mthread-model posix -pthread
--- a/Makefile
+++ b/Makefile
@@ -183,6 +183,10 @@ MUSL_PRINTSCAN_SOURCES = \
     $(LIBC_TOP_HALF_MUSL_SRC_DIR)/stdio/vfscanf.c \
     $(LIBC_TOP_HALF_MUSL_SRC_DIR)/stdlib/strtod.c \
     $(LIBC_TOP_HALF_MUSL_SRC_DIR)/stdlib/wcstod.c
+BULK_MEMORY_SOURCES = \
+    $(LIBC_TOP_HALF_MUSL_SRC_DIR)/string/memcpy.c \
+    $(LIBC_TOP_HALF_MUSL_SRC_DIR)/string/memmove.c \
+    $(LIBC_TOP_HALF_MUSL_SRC_DIR)/string/memset.c
 LIBC_TOP_HALF_HEADERS_PRIVATE = $(LIBC_TOP_HALF_DIR)/headers/private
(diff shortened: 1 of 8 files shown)


TerrorJack:
Ready for first pass of review.

Haven't made microbenchmarks to gather statistics yet, but IIUIC the only scenario where a wasm loop can be faster than native implementation might be: frequent calls to operate on very small chunks. Some engines may have overhead to call native methods, who knows.

TerrorJack:
I also eyeballed the `wasm2wat` CI artifacts, compiled with LLVM 10 & 13. `memcpy`, `memmove`, `memset` all correctly use bulk memory opcodes, but when I attempted to use `__builtin_wmemcpy`, LLVM 10 emits:

```wat
(module
  (type (;0;) (func (param i32 i32 i32) (result i32)))
  (import "env" "__linear_memory" (memory (;0;) 0))
  (import "env" "__indirect_function_table" (table (;0;) 0 funcref))
  (func $wmemcpy (type 0) (param i32 i32 i32) (result i32)
    (local i32)
    local.get 3))
```

And LLVM 13 emits:

```wat
(module
  (type (;0;) (func (param i32 i32 i32) (result i32)))
  (import "env" "__linear_memory" (memory (;0;) 0))
  (func $wmemcpy (type 0) (param i32 i32 i32) (result i32)
    loop (result i32)  ;; label = @1
      br 0 (;@1;)
    end))
```

So the builtin intrinsics doesn't work for wide char versions of the functions. Might be an upstream bug.

Makefile
```diff
@@ -365,6 +370,9 @@ $(MUSL_PRINTSCAN_NO_FLOATING_POINT_OBJS): CFLAGS += \
 	    -D__wasilibc_printscan_no_floating_point \
 	    -D__wasilibc_printscan_floating_point_support_option="\"remove -lc-printscan-no-floating-point from the link command\""
 
+$(BULK_MEMORY_OBJS): CFLAGS += \
+        -mbulk-memory
```
sbc100, line 374: I think if safe/better to pass this to all object files.   No reason not to do that is there?

TerrorJack: I've encountered `fatal error: error in backend: only -ftls-model=local-exec is supported for now on non-Emscripten OSes: variable errno` when compiling some sources with `-mbulk-memory`

sbc100: That sounds like its worth looking into.   I don't see why adding `-mbulk-memory` would generate that error.  Would you mind opening an llvm issue about that?

TerrorJack: Sure, but LLVM bugzilla seems to prohibit new logins for now. The CI log with the clang-13 panicking command line can be found at https://github.com/tweag/wasi-libc/runs/4420917003?check_suite_focus=true#step:8:8. In the meantime, I believe it makes sense to resolve this thread and keep the current workaround as it is?

sbc100: I guess I don't mind landing this patch as workaround for this clang issue, but I would like to be able to add a TODO to the Makefile along with a link to a bug.  

Perhaps a TODO with an explanatory comment would be enough?  Something like: "Limit the use of this feature to just a few compilation units due to clang crash when we try to use it throughout.  TODO: Remove this once clang gets fixed"

Is there some reason you need this patch to land sooner rather than later?

TerrorJack: Fair enough. I have no particular reason to rush getting this merged since we're using our fork anyway, this PR is just meant to avoid duplicate upstream work. I'm fine with waiting for LLVM bug to be filed and fixed then applying `-mbulk-memory` globally.

sbc100:
Might be worth measure that actual performs in shipping runtimes as @kripken suggests in the bug.. otherwise LGTM

sbc100 (approved):

sunfishcode:
Benchmarks would be nice to do, though I'm inclined to merge this either way. LLVM optimizes small fixed-size memcpy/etc. into inline loads and stores already, and beyond that, this is one of the main use cases that bulk-memory was added for, so in theory it shouldn't be slow anywhere. If it does turn out to be slow somewhere, and it isn't just a missing optimization in a particular engine, we can of course revisit this.

kripken:
Just as a warning, the benchmarks I saw were quite slow, so you might be making wasi-libc 2x slower or so on commonly-called functions. There was also debate back then as to whether this was an intended use of bulk memory or not (i.e. should the VM emit fast code for both small and large operations, both aligned and unaligned, etc.). All that was a while ago though, so hopefully it's not a problem any more!

kripken:
Link to some recent discussion on this:

kripken:
Link to some recent discussion on this (with no clear conclusions):

https://github.com/WebAssembly/binaryen/issues/4403

sunfishcode:
I've also now done some benchmarking and instrumentation, and these small unaligned memcpys are more frequent than I had guessed they'd be. For example, they come up when doing formatted I/O, to copy all the individual strings into the I/O buffer. 

@TerrorJack Would you be interested in extending this PR to have fast paths for small lengths?

TerrorJack:
@sunfishcode Hi, of course, but I'd like to know how to obtain the heuristic small length first.

sunfishcode:
@TerrorJack Going by [this data](https://github.com/WebAssembly/binaryen/issues/4403#issuecomment-1006136803) I think the optimal number will be somewhere between 8 and 100. Perhaps we could try starting with 32? That's long enough to easily cover most formatted-I/O use cases, but perhaps short enough to keep the code simple?

TerrorJack:
@sunfishcode Done.

sunfishcode:
Thanks! We can experiment with the threshold and see how it works out in practice.

kripken:
Btw, it might be good to document this somewhere (not sure where though?). I saw someone run into a problem because they used wasi-libc and the code couldn't run in a VM since the VM didn't support bulk memory. The error message "invalid section 12" doesn't really help point people to rebuilding wasi-libc, unforuntately...

sunfishcode:
If you have any ideas where we could document this such that a user would find it when they need to, I'm open to documenting this.

PR #264, opened 2021-12-20
Reset the register stack and counter

lum1n0us:
It turns out that *clang* will surrounding every exported function with `__wasm_call_ctors` and `__wasm_call_dtors`. `__wasm_call_ctors` will call `__cxa_atexit` to register functions which will be triggered by `__wasm_call_dtors`. In other words, the loop of `__cxa_atexit` and `__funcs_on_exit` will be processed multiple times.

but the value of `head` and the value of `slot` will be `NULL` and `COUNT` after ending the for loop in `__funcs_on_exit`. The wrong value leads:
- a memory leak at [L55](https://github.com/WebAssembly/wasi-libc/blob/ad5133410f66b93a2381db5b542aad5e0964db96/libc-top-half/musl/src/exit/atexit.c#L55)
- bind an empty register list but a logically full list at [L60](https://github.com/WebAssembly/wasi-libc/blob/ad5133410f66b93a2381db5b542aad5e0964db96/libc-top-half/musl/src/exit/atexit.c#L60). It will bind *builtin* as a full list but actually it is empty after `__funcs_on_exit`.

### DIFF
--- a/libc-top-half/musl/src/exit/atexit.c
+++ b/libc-top-half/musl/src/exit/atexit.c
@@ -37,6 +37,9 @@ void __funcs_on_exit()
 		func(arg);
 		LOCK(lock);
 	}
+	/* reset */
+	head = &builtin;
+	slot = 0;
 }
 
 void __cxa_finalize(void *dso)


TerrorJack:
> It turns out that clang will surrounding every exported function with __wasm_call_ctors and __wasm_call_dtors

IIUIC this can be avoided by passing `-mexec-model=reactor` to compile to a wasi reactor. If targetting wasi command, only one exported function (`_start`) should be called once.

lum1n0us:
Got it. Thanks!

BTW, does a *reactor* suppose its `_initialize()` should be executed during instantiation before any call from a *command*?

TerrorJack:
Yes, `_initialize()` will be executed once (if present) before any other export is called, see https://github.com/WebAssembly/WASI/blob/main/design/application-abi.md#current-unstable-abi

How `__wasm_call_dtors` is handled in reactor is unclear to me though.

TerrorJack:
Here are some more observations regarding `__wasm_call_dtors`. Consider the example at https://gist.github.com/TerrorJack/c223fce9df844c85318a6e8f115f3ab7:

* There's a constructor function `foo`, a destructor function `bar`
* `_initialize` will call `foo`, then indirectly call `__cxa_atexit` which registers `bar` as a finalizer
* `__wasm_call_dtors` is not exported by default, but the function is indeed in the linked wasm module, and it will call `__funcs_on_exit` to traverse the finalizer list and call them one by one
* The logic of `__cxa_atexit` and `__funcs_on_exit` is defined at https://github.com/WebAssembly/wasi-libc/blob/main/libc-top-half/musl/src/exit/atexit.c

So overall conclusion seems to be:

* If you expect to call any exported non-main function, always target a wasi reactor
* If you are using a wasi reactor and you still want to do finalization, pass `-Wl,--export=__wasm_call_dtors` to export `__wasm_call_dtors` so the entry to finalizer is available as an export

Regardless of wasi command/reactor, I think the assumption of `__wasm_call_ctors`/`__wasm_call_dtors` mustn't be called more than once should hold.

sunfishcode:
That sounds rights, with a few additional caveats:
 - `__wasm_call_dtors` is a compiler-internal symbol, which may be renamed or removed, or may change meaning, in future versions of LLVM, wasi-sdk, and/or wasi-libc, without notice.
 - Once `__wasm_call_dtors` is called (in any way), the instance should be considered finalized and no longer usable, so besides just not calling `__wasm_call_dtors` again, one must ensure that no functions are ever called.

As background, the reactor model was a start at providing more useful tooling in this space, but as we were developing it, it became clear that this functionality would be subsumed by [module linking](https://github.com/WebAssembly/module-linking/). It's an overall goal of WASI to avoid having its own specs which overlap with lower-level specs, so I'm expecting WASI and wasi-sdk will migrate to module linking. This will also be an opportunity to provide more complete functionality in this space (and incidentally, may require changing the meaning of `__wasm_call_dtors`).

TerrorJack:
> As background, the reactor model was a start at providing more useful tooling in this space, but as we were developing it, it became clear that this functionality would be subsumed by module linking

That's totally news to me! Thanks for the reminder.

lum1n0us:
May I ask the version of the *wasi-sdk* in the example? 

I am using [wasi-sdk-14](https://github.com/WebAssembly/wasi-sdk/releases/tag/wasi-sdk-14) and it turns out:
- only a command has `_funcs_on_exit` in the wasm module. 
- the `destructor` and callbacks registered with `atexit` in a reactor won't be executed.

Not sure whether they are expected or not.

lum1n0us:
Plus, Can I assume a *.wasm* module which comes from Rust also follows the same Command/Reactor model?

sbc100:
> Plus, Can I assume a _.wasm_ module which comes from Rust also follows the same Command/Reactor model?

I think so yes.  The command / reactor model is implement in llvm and wasm-ld (the linker).  rust also uses these tools under the hood so it should output compliant binaries.

PR #265, opened 2022-01-11
Adjust the Makefile for LLVM 14

glandium:
After [1], LLVM sets __NO_MATH_ERRNO__

1. https://github.com/llvm/llvm-project/commit/f282b6809105075b65974989459ee420ecd406e9

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -535,6 +535,7 @@ check-symbols: startup_files libc
 	@#
 	@# TODO: Undefine __FLOAT128__ for now since it's not in clang 8.0.
 	@# TODO: Filter out __FLT16_* for now, as not all versions of clang have these.
+	@# TODO: Filter out __NO_MATH_ERRNO_ and a few __*WIDTH__ that are new to clang 14.
 	$(CC) $(CFLAGS) "$(SYSROOT_SHARE)/include-all.c" \
 	    -isystem $(SYSROOT_INC) \
 	    -std=gnu17 \
@@ -553,8 +554,11 @@ check-symbols: startup_files libc
 	    -U__GNUC_PATCHLEVEL__ \
 	    -U__VERSION__ \
 	    -U__FLOAT128__ \
+	    -U__NO_MATH_ERRNO__ \
+	    -U__BITINT_MAXWIDTH__ \
 	    | sed -e 's/__[[:upper:][:digit:]]*_ATOMIC_\([[:upper:][:digit:]_]*\)_LOCK_FREE/__compiler_ATOMIC_\1_LOCK_FREE/' \
 	    | grep -v '^#define __FLT16_' \
+	    | grep -v '^#define __\(BOOL\|INT_\(LEAST\|FAST\)\(8\|16\|32\|64\)\|INT\|LONG\|LLONG\|SHRT\)_WIDTH__' \
 	    > "$(SYSROOT_SHARE)/predefined-macros.txt"
 
 	# Check that the computed metadata matches the expected metadata.


Makefile
```diff
@@ -538,6 +538,7 @@ check-symbols: startup_files libc
 	    -U__GNUC_PATCHLEVEL__ \
 	    -U__VERSION__ \
 	    -U__FLOAT128__ \
+	    -U__NO_MATH_ERRNO__ \
```
sbc100, line 541: Maybe add a TODO above to go with the others... and maybe we should fix those TODOs ?

glandium: There aren't TODOs for the others, though. And I'm not sure what there would be to actually do here...

sbc100: For example `TODO: Undefine __FLOAT128__ for now since it's not in clang 8.0.` above.. it sounds like we no longer need to undefined that macro because we don't need to support llvm 8.0.0 anymore.

Basically some of these we exclusions we want keep forever and other are there to ease the transition from on llvm version to another.  In this case I believe that this is just a transitory addition, right?  Once we require llvm 14 we can remove this line?

glandium:
Well, it didn't last long... there's more "breakage" with even newer clang trunk, coming from https://github.com/llvm/llvm-project/commit/bf7d9970ba0ac5ecfa1a469086f5789de5c94e3f

cuviper:
LLVM release/14.x has branched, along with rc1 -- is it time to finalize the necessary updates?

cuviper:
With `clang-14.0.0-1.fc36.x86_64`, the only differences are those added WIDTH macros.

<details> 
  <summary>Result of <code>make check-symbols</code></summary>

```diff
# Check that the computed metadata matches the expected metadata.
# This ignores whitespace because on Windows the output has CRLF line endings.
diff -wur "/home/jistone/src/wasi-libc/expected/wasm32-wasi" "/home/jistone/src/wasi-libc/sysroot/share/wasm32-wasi"
diff -wur /home/jistone/src/wasi-libc/expected/wasm32-wasi/predefined-macros.txt /home/jistone/src/wasi-libc/sysroot/share/wasm32-wasi/predefined-macros.txt
--- /home/jistone/src/wasi-libc/expected/wasm32-wasi/predefined-macros.txt      2022-04-04 11:11:14.738017260 -0700
+++ /home/jistone/src/wasi-libc/sysroot/share/wasm32-wasi/predefined-macros.txt 2022-04-04 11:15:36.495340552 -0700
@@ -2351,6 +2351,8 @@
 #define __BIGGEST_ALIGNMENT__ 16
 #define __BIG_ENDIAN 4321
 #define __BIND 19950621
+#define __BITINT_MAXWIDTH__ 128
+#define __BOOL_WIDTH__ 8
 #define __BYTE_ORDER __BYTE_ORDER__
 #define __BYTE_ORDER__ __ORDER_LITTLE_ENDIAN__
 #define __CHAR16_TYPE__ unsigned short
@@ -2519,35 +2521,44 @@
 #define __INT_FAST16_FMTi__ "hi"
 #define __INT_FAST16_MAX__ 32767
 #define __INT_FAST16_TYPE__ short
+#define __INT_FAST16_WIDTH__ 16
 #define __INT_FAST32_FMTd__ "d"
 #define __INT_FAST32_FMTi__ "i"
 #define __INT_FAST32_MAX__ 2147483647
 #define __INT_FAST32_TYPE__ int
+#define __INT_FAST32_WIDTH__ 32
 #define __INT_FAST64_FMTd__ "lld"
 #define __INT_FAST64_FMTi__ "lli"
 #define __INT_FAST64_MAX__ 9223372036854775807LL
 #define __INT_FAST64_TYPE__ long long int
+#define __INT_FAST64_WIDTH__ 64
 #define __INT_FAST8_FMTd__ "hhd"
 #define __INT_FAST8_FMTi__ "hhi"
 #define __INT_FAST8_MAX__ 127
 #define __INT_FAST8_TYPE__ signed char
+#define __INT_FAST8_WIDTH__ 8
 #define __INT_LEAST16_FMTd__ "hd"
 #define __INT_LEAST16_FMTi__ "hi"
 #define __INT_LEAST16_MAX__ 32767
 #define __INT_LEAST16_TYPE__ short
+#define __INT_LEAST16_WIDTH__ 16
 #define __INT_LEAST32_FMTd__ "d"
 #define __INT_LEAST32_FMTi__ "i"
 #define __INT_LEAST32_MAX__ 2147483647
 #define __INT_LEAST32_TYPE__ int
+#define __INT_LEAST32_WIDTH__ 32
 #define __INT_LEAST64_FMTd__ "lld"
 #define __INT_LEAST64_FMTi__ "lli"
 #define __INT_LEAST64_MAX__ 9223372036854775807LL
 #define __INT_LEAST64_TYPE__ long long int
+#define __INT_LEAST64_WIDTH__ 64
 #define __INT_LEAST8_FMTd__ "hhd"
 #define __INT_LEAST8_FMTi__ "hhi"
 #define __INT_LEAST8_MAX__ 127
 #define __INT_LEAST8_TYPE__ signed char
+#define __INT_LEAST8_WIDTH__ 8
 #define __INT_MAX__ 2147483647
+#define __INT_WIDTH__ 32
 #define __IS_CX(x) (__IS_FP(x) && sizeof(x) == sizeof((x)+I))
 #define __IS_FP(x) (sizeof((x)+1ULL) == sizeof((x)+1.0f))
 #define __IS_REAL(x) (__IS_FP(x) && 2*sizeof(x) == sizeof((x)+I))
@@ -2569,9 +2580,11 @@
 #define __LDBL_MIN__ 3.36210314311209350626267781732175260e-4932L
 #define __LITTLE_ENDIAN 1234
 #define __LITTLE_ENDIAN__ 1
+#define __LLONG_WIDTH__ 64
 #define __LONG_LONG_MAX__ 9223372036854775807LL
 #define __LONG_MAX __LONG_MAX__
 #define __LONG_MAX__ 2147483647L
+#define __LONG_WIDTH__ 32
 #define __NAMESER 19991006
 #define __NEED_FILE
 #define __NEED___isoc_va_list
@@ -2667,6 +2680,7 @@
 #define __RETCAST_REAL(x)
 #define __SCHAR_MAX__ 127
 #define __SHRT_MAX__ 32767
+#define __SHRT_WIDTH__ 16
 #define __SID ('S' << 8)
 #define __SIG_ATOMIC_MAX__ 2147483647L
 #define __SIG_ATOMIC_WIDTH__ 32
make: *** [Makefile:513: check-symbols] Error 1
```

</details>

glandium:
There is another problem, caused by https://reviews.llvm.org/D114109, which might be more of a problem for wasi-sdk rather than wasi-libc itself, but the end result is that `<atomic>` ends up included when it wasn't before, with unintended consequences.

sunfishcode:
Thanks!

PR #270, opened 2022-03-16
Try to restore some build system conventions in a low-impact way

### LINKED ISSUE #269: A few packaging nits, coming from Nixpkgs/NixOS
Hi, you can see how we currently package this in https://github.com/nixos/nixpkgs/blob/master/pkgs/development/libraries/wasilibc/default.nix. 

There are no imminent problems (the situation is less fraught than I thought) but I few things I think could be improved.

## Install dirs

We generally like to have "mutliple outputs" for our libraries. That means we install to something like `/nix/store/random-foo/lib`, `/nix/store/random-bar/include`, etc. notice the `random-{foo,bar}` are not the same.

We generally use the "GNU-style: install dirs" to achieve this. C.f.

 - Autoconf: https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Installation-Directory-Variables.html
 - CMake: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
 - Meson: https://mesonbuild.com/Builtin-options.html#directories

You have
https://github.com/WebAssembly/wasi-libc/blob/079adff840032c3455eb1cb34dc9ceaa0b2bfc0c/Makefile#L8
and
https://github.com/WebAssembly/wasi-libc/blob/079adff840032c3455eb1cb34dc9ceaa0b2bfc0c/Makefile#L246-L248

for this. It would be nice to make the second ones `?=` too, and relax the assumption that that they share a parent directory.

That list bit part means getting rid of other uses of `SYSROOT` (it can just be a way to default the defaults for those three, so e.g.:

## Tool env vars

You use `WASM_*`, e.g. `WASM_CC`. The convention however is that `CC`, `LD`, etc. are the tools for the *current* package, for the *current* platform, one is installing too. Conversely, if one needed to build e.g. some temporary codegen tool (a common use case, but not on in wasi libc to my knowledge), one would use `CC_FOR_BUILD`, `LD_FOR_BUILD` etc.

This means the `WASM_` prefix is not needed.

I would suggest just getting rid of them. The wrapper `wasi-sdk` already does recursive make:
https://github.com/WebAssembly/wasi-sdk/blob/20c5bcd791e22895ea1c9721c9bd88081e3ed1f6/Makefile#L92-L96
```make
build/wasi-libc.BUILT: build/llvm.BUILT
	$(MAKE) -C $(ROOT_DIR)/src/wasi-libc \
		WASM_CC=$(BUILD_PREFIX)/bin/clang \
		...
```
This makes scoped variables easy, so you can just do
```make
build/wasi-libc.BUILT: build/llvm.BUILT
	$(MAKE) -C $(ROOT_DIR)/src/wasi-libc \
		CC=$(BUILD_PREFIX)/bin/clang \
		...
```
instead, and not worry about clashes with any "global" `CC`.

Ericson2314:
Fixes #269

The "don't move the variables independently if you want to `make
install`" part is a gross, but as neither `wasi-sdk` or Nixpkgs actually
needs `make install` (we both can just build right into the directories
we want) I figured it wasn't actually so heinous in practice.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,9 +19,9 @@ jobs:
         curl -fsSLO https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/LLVM-10.0.0-win64.exe
         7z x LLVM-10.0.0-win64.exe -y -o"llvm"
         echo "$(pwd)/llvm/bin" >> $GITHUB_PATH
-        echo "WASM_CC=$(pwd)/llvm/bin/clang.exe" >> $GITHUB_ENV
-        echo "WASM_AR=$(pwd)/llvm/bin/llvm-ar.exe" >> $GITHUB_ENV
-        echo "WASM_NM=$(pwd)/llvm/bin/llvm-nm.exe" >> $GITHUB_ENV
+        echo "CC=$(pwd)/llvm/bin/clang.exe" >> $GITHUB_ENV
+        echo "AR=$(pwd)/llvm/bin/llvm-ar.exe" >> $GITHUB_ENV
+        echo "NM=$(pwd)/llvm/bin/llvm-nm.exe" >> $GITHUB_ENV
       if: matrix.os == 'windows-latest'
 
     - name: Override llvm-nm with one from rustup (Windows)
@@ -29,7 +29,7 @@ jobs:
         rustup update stable
         rustup default stable
         rustup component add llvm-tools-preview
-        echo "WASM_NM=$(rustc --print sysroot|sed 's|C:|/c|'|sed 's|\\|/|g')/lib/rustlib/x86_64-pc-windows-msvc/bin/llvm-nm.exe" >> $GITHUB_ENV
+        echo "NM=$(rustc --print sysroot|sed 's|C:|/c|'|sed 's|\\|/|g')/lib/rustlib/x86_64-pc-windows-msvc/bin/llvm-nm.exe" >> $GITHUB_ENV
       if: matrix.os == 'windows-latest'
 
     - name: Install LLVM tools (MacOS)
@@ -38,9 +38,9 @@ jobs:
         curl -sSfL https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/clang+llvm-10.0.0-x86_64-apple-darwin.tar.xz | tar xJf -
         export CLANG_DIR=`pwd`/clang+llvm-10.0.0-x86_64-apple-darwin/bin
         echo "$CLANG_DIR" >> $GITHUB_PATH
-        echo "WASM_CC=$CLANG_DIR/clang" >> $GITHUB_ENV
-        echo "WASM_AR=$CLANG_DIR/llvm-ar" >> $GITHUB_ENV
-        echo "WASM_NM=$CLANG_DIR/llvm-nm" >> $GITHUB_ENV
+        echo "CC=$CLANG_DIR/clang" >> $GITHUB_ENV
+        echo "AR=$CLANG_DIR/llvm-ar" >> $GITHUB_ENV
+        echo "NM=$CLANG_DIR/llvm-nm" >> $GITHUB_ENV
       if: matrix.os == 'macos-latest'
 
     - name: Install LLVM tools (Linux)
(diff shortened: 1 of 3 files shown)


sbc100:
Seems reasonable. 

I wonder if we should just remove the `install` target if its not being used anywhere (and therefore untested).

sbc100:
I don't the github actions running here... any idea whats up with that?

sbc100:
Looks like github actions might be down/struggling: https://www.githubstatus.com/

Makefile
```diff
@@ -185,7 +195,7 @@ LIBC_TOP_HALF_ALL_SOURCES = \
     $(shell find $(LIBC_TOP_HALF_SOURCES) -name \*.c)
 
 # Set the target.
-CFLAGS = $(WASM_CFLAGS) --target=$(TARGET_TRIPLE)
+CFLAGS += --target=$(TARGET_TRIPLE)
```
sunfishcode, line 198: If someone overwrites `CFLAGS` on the make command line, does it override the `--target=$(TARGET_TRIPLE)` part too, or just the `-O2 -DNDEBUG` part?

Makefile
```diff
@@ -483,10 +491,10 @@ check-symbols: startup_files libc
 	@# TODO: Use llvm-nm --extern-only instead of grep. This is blocked on
 	@# LLVM PR40497, which is fixed in 9.0, but not in 8.0.
 	@# Ignore certain llvm builtin symbols such as those starting with __mul
-	@# since these dependencies can vary between llvm versions.
-	"$(WASM_NM)" --defined-only "$(SYSROOT_LIB)"/libc.a "$(SYSROOT_LIB)"/libwasi-emulated-*.a "$(SYSROOT_LIB)"/*.o \
+	@# since these dependencies caisysroot  vary between llvm versions.
```
sunfishcode, line 494: Is "caisysroot" a typo?

Ericson2314: Oops! Yes it is.

Makefile
Ericson2314: I think just the `O2 -DNDEBUG` part, but will check.

Makefile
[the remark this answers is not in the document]
Ericson2314: Indeed. Sorry I didn't notice that line was touched!

Makefile
Ericson2314: @sunfishcode I tested it, and it indeed the CLI was overriding everything. I added a bunch of `override` so that this is no longer the case, and just the `-O2 -DNDEBUG` part gets overridden.

Makefile
```diff
@@ -185,12 +195,12 @@ LIBC_TOP_HALF_ALL_SOURCES = \
     $(shell find $(LIBC_TOP_HALF_SOURCES) -name \*.c)
 
 # Set the target.
-CFLAGS = $(WASM_CFLAGS) --target=$(TARGET_TRIPLE)
+override CFLAGS += --target=$(TARGET_TRIPLE)
```
sbc100, line 198: Could we avoid all those ugly overrides by instead doing something like:

```
CFLAGS = $(EXTRA_CFLAGS) --target=$(TARGET_TRIPLE)
```

I believe `EXTRA_CFLAGS` is a relatively standard way to inject extra `CFLAGS`.

Ericson2314: Yeah sure!

Ericson2314:
Oops forgot about CI and the README too. I did a grep and there should be no more pesky `WASM_` at all, now.

sbc100 (approved):

sunfishcode:
The CI is now getting "fatal error: 'wasi/api.h' file not found" errors, and it's not immediately obvious what the problem is.

Ericson2314:
@sunfishcode I think I am using `-isystem`, but we in Nixpkgs were getting that error *before* too, and hacked around it. I tried various things, but nothing is making sense, so I messaged for help in the Zulip.

.github/workflows/main.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           curl -fsSLO https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/LLVM-10.0.0-win64.exe
           7z x LLVM-10.0.0-win64.exe -y -o"llvm"
           echo "$(pwd)/llvm/bin" >> $GITHUB_PATH
-          echo "WASM_AR=$(pwd)/llvm/bin/llvm-ar.exe" >> $GITHUB_ENV
+          echo "AR=$(pwd)/llvm/bin/llvm-ar.exe" >> $GITHUB_ENV
```
Ericson2314, line 22: Oh oops, I set `AR` twice, but it is on the path so it still works. Maybe someone tell what what style they prefer?

sunfishcode: Either way seems fine, but right now the other platforms set AR below, and it seems best to be consistent about it.

Ericson2314:
Closing since I broke up into bits.

PR #273, opened 2022-03-21
Use `-isystem` instead of `--sysroot` for the installed headers

Ericson2314:
This is less of a sledgehammer, and prepares the way for #270.


sunfishcode:
Looks reasonable!

PR #274, opened 2022-03-21
Allow varying the include, lib, and share dirs independently

Ericson2314:
Progress towards #270

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,18 @@ WASM_CC ?= clang
 WASM_NM ?= $(patsubst %clang,%llvm-nm,$(filter-out ccache sccache,$(WASM_CC)))
 WASM_AR ?= $(patsubst %clang,%llvm-ar,$(filter-out ccache sccache,$(WASM_CC)))
 WASM_CFLAGS ?= -O2 -DNDEBUG
-# The directory where we build the sysroot.
+# These variables describe the locations of various files and
+# directories to install.
+#
+# `SYSROOT` is just used to define the defaults, and in the install
+# rule. That means you should not override `SYSROOT_*` if you are going
+# to run `make install`, but if you don't need to, you are free to move
+# the `SYSROOT_*` directories around freely, not preserving their
+# default locations relative `SYSROOT`.
 SYSROOT ?= $(CURDIR)/sysroot
+SYSROOT_LIB ?= $(SYSROOT)/lib/$(MULTIARCH_TRIPLE)
+SYSROOT_INC ?= $(SYSROOT)/include
+SYSROOT_SHARE ?= $(SYSROOT)/share/$(MULTIARCH_TRIPLE)
 # A directory to install to for "make install".
 INSTALL_DIR ?= /usr/local
 # single or posix
@@ -252,12 +262,6 @@ LIBWASI_EMULATED_GETPID_OBJS = $(call objs,$(LIBWASI_EMULATED_GETPID_SOURCES))
 LIBWASI_EMULATED_SIGNAL_OBJS = $(call objs,$(LIBWASI_EMULATED_SIGNAL_SOURCES))
 LIBWASI_EMULATED_SIGNAL_MUSL_OBJS = $(call objs,$(LIBWASI_EMULATED_SIGNAL_MUSL_SOURCES))
 
-# These variables describe the locations of various files and
-# directories in the generated sysroot tree.
-SYSROOT_LIB := $(SYSROOT)/lib/$(MULTIARCH_TRIPLE)
-SYSROOT_INC = $(SYSROOT)/include
-SYSROOT_SHARE = $(SYSROOT)/share/$(MULTIARCH_TRIPLE)
-
 # Files from musl's include directory that we don't want to install in the
 # sysroot's include directory.
 MUSL_OMIT_HEADERS :=
@@ -420,7 +424,7 @@ $(LIBWASI_EMULATED_PROCESS_CLOCKS_OBJS): CFLAGS += \
     -I$(LIBC_BOTTOM_HALF_CLOUDLIBC_SRC)
 
 include_dirs:
-	$(RM) -r "$(SYSROOT)"
+	$(RM) -r "$(SYSROOT_INC)"
 
 	#
 	# Install the include files.
@@ -472,7 +476,11 @@ finish: startup_files libc
 	done
 
 	#
-	# The build succeeded! The generated sysroot is in $(SYSROOT).
+	# The build succeeded! The generated files are in
+	#
+	# lib:     $(SYSROOT_LIB)
+	# include: $(SYSROOT_INC)
+	# share:   $(SYSROOT_SHARE)
 	#
 
 # The check for defined and undefined symbols expects there to be a heap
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -420,7 +424,7 @@ $(LIBWASI_EMULATED_PROCESS_CLOCKS_OBJS): CFLAGS += \
     -I$(LIBC_BOTTOM_HALF_CLOUDLIBC_SRC)
 
 include_dirs:
-	$(RM) -r "$(SYSROOT)"
+	$(RM) -r "$(SYSROOT_INC)"
```
sbc100, line 427: I think its we are building the sysroot from scratch we probably want remove the entire thing somewhere. (doesn't have to be here).

Ericson2314: @sbc100 well the idea with this change is that there may not be a "sysroot" anymore --- these dirs are independent can couldn't have `/` even as their most recent common parent. I just didn't rename the variables to avoid churn.

sbc100: I think we should still default to the normal "build an entire sysroot" approach.   This new approach your are proposing probably isn't want to want for wasi-sdk.. or, I imagine, what most folks building wasi-libc will want, so maybe it should opt-in?

sbc100: BTW are you actually planning to use this approach of having the sysroot split up like this, or is it just to conform to the norms of your build system?   i.e. in practice, will your package of wasm-libc still have sysroot structure (i.e. will your users be running clang with --sysroot to access wasi-libc?  If not, how will they be running it?)

sbc100: The concept of "sysroot" is something clang understands so I think it will still be useful if in the final thing you package has a sysroot structure.. but maybe I'm misunderstanding something.

Ericson2314: Sorry I just saw these now. We do use this, and we do install in directories with a more distant common ancestor.

Ericson2314: @sbc100 you may be interested in https://github.com/mesonbuild/meson/pull/9903#issuecomment-1071105428 which is us trying to do this thing for Meson (and therefore fix it for a lot of packages in one swoop). A few other folks with similar needs chimed in there too.

sbc100: I wonder if you can get away with building the entire sysroot like we do (locally in a single root that we control), and then just modifying the `install` phase.    Technically its the install phase you care about right?  Its not until we get to installation that we want to honor the build systems target directories is it?

sbc100: If we allow SYSROOT_INC to be externally controlled it might not be something we can just remove with `rm -rf` like this.   What if I was to build with `SYSROOT_INC=/usr/local/include`..  this also makes me think is that `install` phase you want to be working with.

Ericson2314:
@sbc100 (think it is might be better to discus wider issue outside the code comment err keep in mind what you observed before:

Yes, I can build normally but and install specially, but.....

1. Neither you all (`wasi-sdk`) or us currently use `make install`. As you remarked before, we could just delete it!

2. `make install` is the only issue where relative locations of those dirs actually matter! Everything else you propose maybe shouldn't change is trivial to change. Conversely, I don't know how to write a more genearl `make install` rule.

Makefile
```diff
@@ -558,8 +566,25 @@ check-symbols: startup_files libc
...
 	mkdir -p "$(INSTALL_DIR)"
 	cp -r "$(SYSROOT)/lib" "$(SYSROOT)/share" "$(SYSROOT)/include" "$(INSTALL_DIR)"
+.PHONY: install
+endif
+
+clean:
+	$(RM) -r "$(OBJDIR)"
+	$(RM) -r "$(SYSROOT_INC)"
+	$(RM) -r "$(SYSROOT_LIB)"
+	$(RM) -r "$(SYSROOT_SHARE)"
```
Ericson2314, line 588: @sbc100 I added a `make clean` rule, so there is still an easy and standard way to clean up after everything.

Makefile
Ericson2314: I replied below to what you wrote just above.

As to your original question

> I think its we are building the sysroot from scratch we probably want remove the entire thing somewhere. (doesn't have to be here).

I made `make clean` rule below. I dislike the "sysroot" terminology, but I have no issue with wanting to make a rule to "start over"!

Makefile
```diff
@@ -558,8 +566,25 @@ check-symbols: startup_files libc
...
+
+# If the 3 dirs are not within the standard locations, the install rule
+# is undefined. This protects the user from doing something silly by
+# mistake.
+ifeq ($(SYSROOT)/lib $(SYSROOT)/include $(SYSROOT)/share,$(strip $(NORMAL_SYSROOT)))
 install: finish
 	mkdir -p "$(INSTALL_DIR)"
 	cp -r "$(SYSROOT)/lib" "$(SYSROOT)/share" "$(SYSROOT)/include" "$(INSTALL_DIR)"
+.PHONY: install
+endif
```
Ericson2314, line 582: I made `make install` go away for when the directories are in a custom layout, to get rid of the footgun of a broken `make install` doing something the user did not intend.

But per what I wrote in the main thread, and what you wrote in #270, we could just delete `make install` too!

sbc100:
> @sbc100 (think it is might be better to discus wider issue outside the code comment err keep in mind what you observed before:
> 
> Yes, I can build normally but and install specially, but.....
> 
> 1. Neither you all (`wasi-sdk`) or us currently use `make install`. As you remarked before, we could just delete it!
> 2. `make install` is the only issue where relative locations of those dirs actually matter! Everything else you propose maybe shouldn't change is trivial to change. Conversely, I don't know how to write a more genearl `make install` rule.

I think it would be better for you to make your modifications to the `make install` rule (and start using it).   Something like this:

```
install: finish                                                                     
        mkdir -p "$(INSTALL_LIB)"                                                   
        mkdir -p "$(INSTALL_INCLUDE)"                                               
        mkdir -p "$(INSTALL_SHARE)"                                                 
        cp -r "$(SYSROOT)/lib/*" "$(INSTALL_LIB)"                                     
        cp -r "$(SYSROOT)/include/*" "$(INSTALL_INCLUDE)"                             
        cp -r "$(SYSROOT)/share/*" "$(INSTALL_SHARE)"     
```

This has the advantage of cleanly separating the phase (`make`) from the installation phase (`make install`)

(If you dislike the `$SYSROOT` name we can consider renaming it to `$BUILDDIR`, but perhaps we can discuss that as a separate issue.  I quite like the sysroot terminology myself, having worked on compilers and sdk for so many years).

Ericson2314:
@sbc100 Well the problem is the build products are being put in e.g. `$(SYSROOT)/lib/$(MULTIARCH_TRIPLE)` and so with your plan those are installed in `$(INSTALL_LIB)/$(MULTIARCH_TRIPLE)`, but I want to install them in `$(INSTALL_LIB)`. Because we have separate prefixes for everything we never use those arch-specific "libsubdir"s.

Ericson2314:
If you all like, I could make a PR that just did the 

> Generate a test file that includes all public header files

part of this, but keep everything else the same. Then this would be a sort of "secret, completely unstable, accidental feature" I could use in Nixpkgs by overriding the `SYSROOT_*` variables on the command line.

Generally a risky move for downstream, but I'm willing to gamble and take the blame if it ends up breaking.

sbc100:
> @sbc100 Well the problem is the build products are being put in e.g. `$(SYSROOT)/lib/$(MULTIARCH_TRIPLE)` and so with your plan those are installed in `$(INSTALL_LIB)/$(MULTIARCH_TRIPLE)`, but I want to install them in `$(INSTALL_LIB)`. Because we have separate prefixes for everything we never use those arch-specific "libsubdir"s.

We can be more precise then maybe:

```
install: finish                                                                     
        mkdir -p "$(INSTALL_LIB)"                                                   
        mkdir -p "$(INSTALL_INCLUDE)"                                               
        mkdir -p "$(INSTALL_SHARE)"                                                 
        cp -r "$(SYSROOT)/lib/$(MULTIARCH_TRIPLE)/*" "$(INSTALL_LIB)"                                     
        cp -r "$(SYSROOT)/include/*" "$(INSTALL_INCLUDE)"                             
        cp -r "$(SYSROOT)/shared/$(MULTIARCH_TRIPLE)/*" "$(INSTALL_SHARE)"     
```

sbc100:
> If you all like, I could make a PR that just did the
> 
> > Generate a test file that includes all public header files

The changes to the header checks look uncontroversial, yes

Ericson2314:
OK I can prepare both.

PR #275, opened 2022-03-22
Use standardized `CC`, `AR`, `NM` env vars

Ericson2314:
Progress towards #269

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
         echo "$(pwd)/llvm/bin" >> $GITHUB_PATH
         echo "CC=$(pwd)/llvm/bin/clang.exe" >> $GITHUB_ENV
         echo "AR=$(pwd)/llvm/bin/llvm-ar.exe" >> $GITHUB_ENV
-        echo "WASM_NM=$(pwd)/llvm/bin/llvm-nm.exe" >> $GITHUB_ENV
+        echo "NM=$(pwd)/llvm/bin/llvm-nm.exe" >> $GITHUB_ENV
       if: matrix.os == 'windows-latest'
 
     - name: Override llvm-nm with one from rustup (Windows)
@@ -29,7 +29,7 @@ jobs:
         rustup update stable
         rustup default stable
         rustup component add llvm-tools-preview
-        echo "WASM_NM=$(rustc --print sysroot|sed 's|C:|/c|'|sed 's|\\|/|g')/lib/rustlib/x86_64-pc-windows-msvc/bin/llvm-nm.exe" >> $GITHUB_ENV
+        echo "NM=$(rustc --print sysroot|sed 's|C:|/c|'|sed 's|\\|/|g')/lib/rustlib/x86_64-pc-windows-msvc/bin/llvm-nm.exe" >> $GITHUB_ENV
       if: matrix.os == 'windows-latest'
 
     - name: Install LLVM tools (MacOS)
@@ -40,7 +40,7 @@ jobs:
         echo "$CLANG_DIR" >> $GITHUB_PATH
         echo "CC=$CLANG_DIR/clang" >> $GITHUB_ENV
         echo "AR=$CLANG_DIR/llvm-ar" >> $GITHUB_ENV
-        echo "WASM_NM=$CLANG_DIR/llvm-nm" >> $GITHUB_ENV
+        echo "NM=$CLANG_DIR/llvm-nm" >> $GITHUB_ENV
       if: matrix.os == 'macos-latest'
 
     - name: Install LLVM tools (Linux)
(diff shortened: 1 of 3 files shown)


Ericson2314:
@sbc100 this one look good?

.github/workflows/main.yml
```diff
@@ -13,44 +13,45 @@ jobs:
...
-    - name: Install clang (Linux)
+    - name: Install LLVM tools (Linux)
       shell: bash
       run: |
         curl -sSfL https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/clang+llvm-10.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz | tar xJf -
         export CLANG_DIR=`pwd`/clang+llvm-10.0.0-x86_64-linux-gnu-ubuntu-18.04/bin
         echo "$CLANG_DIR" >> $GITHUB_PATH
         echo "WASM_CC=$CLANG_DIR/clang" >> $GITHUB_ENV
+        echo "WASM_AR=$CLANG_DIR/llvm-ar" >> $GITHUB_ENV
+        echo "WASM_NM=$CLANG_DIR/llvm-nm" >> $GITHUB_ENV
```
sbc100, line 54: Why can't we continue to infer the second two based on the first? 

Is the plan to require all 3 going foreword?  Isn't that more/unnecessary work for somebody who wants to build wasi-libc locally?

Ericson2314: We can, and I don't have any intention of taking the `?= $(patsubst %clang,%llvm-nm,$(filter-out ccache sccache,$(WASM_CC)))` out of the Makefile :).

The problem is just the Github Actions provide no good way to clear the environment of whatever crap the OS dropped in there, without repeatedly doing it in every step. (`GITHUB_ENV` doesn't support `unset`, and merely extends an ambient environment.)

That is not a problem with `WASM_*` but is a problem with the standard variables, which shouldn't be initialized for native compilation when Ubuntu and friends have no idea what, if any, compilation we want to do.

So I am doing this so the last step of getting fid of the `WASM_` prefixes can be "just" a `sed`.

sbc100: I see.   Do you know if github actions actually set CC, AR, or NM?

Ericson2314: I am not sure, but by the names of the actions, I don't think so. I think Ubuntu and MSYS might `CC=cc`, though that clashes with the standard of what those are supposed to mean.

(Indeed, I am quite sympathetic to people making their own custom env vars in response to `CC` being set. But this to me is the moral equivalent of implementing `echo foo > file` by opening up a new file descriptor for `file` in deference to `stdout` being "always is for the PTY". `CC` is supposed to convey what C compiler is used for building things that will be installed, rather than point to the "default, native" C compiler which may or may not be suitable for that task. Both `CC` and `stdout` are parameters defined by how the parameters should used, not by what default arguments those parameters are instantiated with.)

Ericson2314: I did this in response to issues that the Windows CI *did* have `CC` set by default, as did the the macOS with `AR`, so I am quite sure this is not hypothetical.

If you mean, just put this commit in the same PR that strips the `WASM_` prefixes, sure. I thought it just might be useful to separate it out so that was "just a sed" and easier to review.

sbc100: I'm confused then.  I asked above "Do you know if github actions actually set CC, AR, or NM?" and your response was "I am not sure, but by the names of the actions, I don't think so.".  

But now it seems you are saying you have indeed seen evidence that the CI *is* setting these env vars?   If that is true then yes this change seem reasonable.   But lets combine it with the change to using bare CC/AR/NM.. then the history will make it clear why needed to start setting them.

Ericson2314: Sorry let me clear this up:

1. I know those variables are getting set.
2. I don't know whether it is an action this CI depends on or the "base image" setting them.

OK can combine.

Ericson2314:
> But now it seems you are saying you have indeed seen evidence that the CI is setting these env vars? If that is true then yes this change seem reasonable. But lets combine it with the change to using bare CC/AR/NM.. then the history will make it clear why needed to start setting them.

OK this PR is retitled and now includes both changes as requested.

Ericson2314:
Aye! it's the accursed
```bash
 	done | grep -v "^__mul" > "/Users/runner/work/wasi-libc/wasi-libc/sysroot/share/wasm32-wasi/undefinod-symbols.txt"
```
`undefined`->`undefinod` error on macOS again. What the hell....

Ericson2314:
So I renamed it to `xundefined-symbols`, and then weird `undefinod` went way. What the hell!!

Ericson2314:
I thought it might be some bit-pattern thing but `xxaaaaaaaaa-symbols` is also fine.

Ericson2314:
`aaaaaaaoa-symbols.txt` Aha! so it has something to do with the length.

Looks like macOS has a stupid bug. Tsk Tsk.

sunfishcode:
Emojis cannot convey how bizarre that bug is.

Ericson2314:
@sunfishcode @sbc100 Looks like it was a `make` bug, and me pulling out the filename into a variable worked around it!

Ericson2314:
@sunfishcode @sbc100 Looks like it was a `make` bug, and me pulling out the filename into a variable worked around it!

sunfishcode:
Cool. This looks good to me; @sbc100 did you have any outstanding concerns?

Ericson2314:
@sbc100 ping :)

Ericson2314:
@sbc100 Ping week 2 :)

sbc100 (approved):
Sorry I was on vacation.  This looks good to me now.

Ericson2314:
Horray! Thanks @sbc100.

PR #277, opened 2022-03-22
Add `make clean` rule, don't have `include_dirs` nuke other things

Ericson2314:
I think this separation of concerns is good, though I don't need it for
Nixpkgs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -420,8 +420,6 @@ $(LIBWASI_EMULATED_PROCESS_CLOCKS_OBJS): CFLAGS += \
     -I$(LIBC_BOTTOM_HALF_CLOUDLIBC_SRC)
 
 include_dirs:
-	$(RM) -r "$(SYSROOT)"
-
 	#
 	# Install the include files.
 	#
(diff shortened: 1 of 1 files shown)


Ericson2314:
@sbc100 this one good?

sbc100 (approved):
Nice!

Makefile
```diff
@@ -420,7 +420,7 @@ $(LIBWASI_EMULATED_PROCESS_CLOCKS_OBJS): CFLAGS += \
     -I$(LIBC_BOTTOM_HALF_CLOUDLIBC_SRC)
 
 include_dirs:
-	$(RM) -r "$(SYSROOT)"
+	$(RM) -r "$(SYSROOT_INC)"
```
sbc100, line 423: I think maybe we should completely remove this line?  If we are separating the cleaning from the building?

Ericson2314: OK

Ericson2314: I thought it would break `make; make`, but it doesn't!

sbc100 (approved):

sbc100:
Side note: I wonder if renaming `$(SYSROOT)` to `$(STAGING_ROOT)` would make it more clear what is going on, and perhaps more acceptable to you in terms of terminology?

Ericson2314:
@sbc100 I don't mind the names :). I may not like sysroots, but I do believe in alpha-equivalence :).

PR #278, opened 2022-03-23
Remove the signal constants and proc_raise function.

### LINKED ISSUE #271: Confusion around SIGCHLD value?
What's the right value to use for SIGCHLD in a wasi app... 16 or 17?  

```
$ find . -type f -print0 | xargs -0 grep CHLD | egrep 'wasm|wasi'
./libc-top-half/musl/arch/wasm32/bits/signal.h:#define SIGCHLD   17
./expected/wasm32-wasi/predefined-macros.txt:#define __WASI_SIGNAL_CHLD (UINT8_C(16))
./libc-bottom-half/headers/public/wasi/api.h:#define __WASI_SIGNAL_CHLD (UINT8_C(16))
```

### LINKED ISSUE #272: WASI signal numbers out of sync with wasi-libc signal numbers
if we compare `wasi-libc/libc-top-half/musl/arch/wasm32/bits/signal.h` with `wasi/api.h`, we see wasi-libc defines `SIGSTKFLT` as signal 16, but this signal does not exist in the WASI spec at all.  this causes all remaining signals to be off-by-one.

for example, the first few around the desync:
| signal | wasi-libc | wasi API |
|---|---|---|
| SIGPIPE | 13 | 13 |
| SIGALRM | 14 | 14 |
| SIGTERM | 15 | 15 |
| SIGSTKFLT | 16 | - |
| SIGCHLD | 17 | 16 |
| SIGCONT | 18 | 17 |
| SIGSTOP | 19 | 18 |

this makes working with signals kind of a pita as we have to maintain 2 tables and their mappings between them.  from a simplicity pov, wasi-libc should just use the same set of signals & signal numbers as wasi api and call it a day.

sunfishcode:
The WASI signal constants and proc_raise function were removed in the
latest [ephemeral], which had been scheduled to be in the next snapshot,
however WASI itself is now transitioning away from the snapshot system.

WASI libc will also be transitioning to updated wit specs once they're
ready, however until that time, we can make the simple change of
removing these signal constants to avoid confusion.

Fixes #271.
Fixes #272.

[ephemeral]: https://github.com/WebAssembly/WASI/tree/main/phases/ephemeral/witx


libc-bottom-half/headers/public/wasi/api.h
```diff
@@ -1300,200 +1300,6 @@ typedef uint32_t __wasi_exitcode_t;
 _Static_assert(sizeof(__wasi_exitcode_t) == 4, "witx calculated size");
 _Static_assert(_Alignof(__wasi_exitcode_t) == 4, "witx calculated align");
 
-/**
```
sbc100, line 1303: Don't we have tests that verify that this auto-generated file is up-to-date with the witx file?

sunfishcode: Ah, yes we do, and they're now failing :-}. I'm going to propose we just disable it for now, and just be careful not to make any ABI-breaking changes for now.

sbc100: Sure thing, maybe modify the comment at the top of the file to indicate this.

sbc100 (approved):

libc-bottom-half/headers/public/wasi/api.h
sunfishcode: Comment updated.

Issue #284, opened 2022-04-14
Support for `accept{,4}`


rvolosatovs:
With #282 in place, it should be now possible to implement `accept` and `accept4`.

rvolosatovs:
cc @Deepansharora27 perhaps you could be interested in contributing to this?

Deepansharora27:
@rvolosatovs Apologies Roman, I missed the notification of you tagging me here 😅

I thought you were working on this for the `0.4` release so I thought to not work on this then. 

Apologies again for missing the tag message

Deepansharora27:
https://github.com/WebAssembly/wasi-libc/blob/7302f33f99233482de8eedfcd21495c20858e0e4/libc-top-half/musl/src/network/accept.c#L1-L7

https://github.com/WebAssembly/wasi-libc/blob/7302f33f99233482de8eedfcd21495c20858e0e4/libc-top-half/musl/src/network/accept4.c#L7-L19

As I was iterating and trying to get an understanding of the repository, I have found prototype implementations of both the `accept()` and `accept4()` functions. 
Since the issue description entails that the required task is to add support for `accept()` and `accept4()`, I am not sure,How can I take this ahead from here given that these two are the prototype implementations for the function.

sunfishcode:
Those accept.c and accept4.c files are in the libc-top-half/musl directory and are part of musl. musl upstream only supports Linux, so to make it work for wasi, we only use musl's "top half"&mdash;things like printf, qsort, etc.&mdash;and disable its "bottom half&mdash;things like read, open, but also accept and accept4 as seen here. Specifically, in the code in these musl files, `socketcall_cp` and `__syscall` perform Linux syscalls, so the code won't work as-is.

For wasi-libc, we'll need to add libc-bottom-half/sources/accept.c and libc-bottom-half/sources/accept4.c, which likely do something similar, but which call the WASI functions instead of doing Linux syscalls. Also, WASI doesn't have `O_CLOEXEC` (as it doesn't have an `exec`), so it doesn't need `SOCK_CLOEXEC`/`FD_CLOEXEC` code.

PR #286, opened 2022-04-29
Adding Support for accept and accept4

Deepansharora27:
I have made a draft PR for this and have made some inital changes

Fixes #284

PR #287, opened 2022-05-02
feat: add support for accept and accept4

haraldh:
Since the socket address of the accepted socket is unknown,
all bytes are set to zero and the length is truncated to the size
of the generic `struct sockaddr`.

Fixes: #284 
Supersedes: #286 

Signed-off-by: Harald Hoyer <harald@profian.com>

### DIFF
--- a/libc-bottom-half/sources/accept.c
+++ b/libc-bottom-half/sources/accept.c
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include <sys/socket.h>
+
+#include <assert.h>
+#include <wasi/api.h>
+#include <errno.h>
+#include <string.h>
+
+int accept(int socket, struct sockaddr *restrict addr, socklen_t *restrict addrlen) {
+  int ret = -1;
+
+  __wasi_errno_t error = __wasi_sock_accept(socket, 0, &ret);
+
+  if (error != 0) {
+    errno = error;
+    return -1;
+  }
+
+  // Clear sockaddr to indicate undefined address
+  memset(addr, 0, *addrlen);
+  // might be AF_UNIX or AF_INET
+  addr->sa_family = AF_UNSPEC;
+  *addrlen = sizeof(struct sockaddr);
+
+  return ret;
+}
+
+int accept4(int socket, struct sockaddr *restrict addr, socklen_t *restrict addrlen, int flags) {
+  int ret = -1;
+
+  if (flags & ~(SOCK_NONBLOCK | SOCK_CLOEXEC)) {
+    errno = EINVAL;
+    return -1;
+  }
+
+  __wasi_errno_t error = __wasi_sock_accept(socket, (flags & SOCK_NONBLOCK) ? __WASI_FDFLAGS_NONBLOCK : 0, &ret);
+
+  if (error != 0) {
+    errno = error;
+    return -1;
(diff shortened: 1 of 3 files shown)


haraldh:
Side note: while writing test code, I encountered:

```
In file included from main.c:4:
In file included from wasi-libc/sysroot/include/sys/socket.h:5:
In file included from wasi-libc/sysroot/include/__header_sys_socket.h:5:
wasi-libc/sysroot/include/__struct_sockaddr.h:10:14: error: use of undeclared identifier 'max_align_t'
    _Alignas(max_align_t) sa_family_t sa_family;
```

I hope, this is just me using `wasi-sdk-14.0` mixed with this HEAD branch's `sysroot`

npmccallum:
@sunfishcode This is the feature I mentioned on the phone. We should consider this a blocker before starting the `wit-bindgen` work.

libc-bottom-half/cloudlibc/src/libc/sys/socket/accept.c
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: BSD-2-Clause
+
+#include <common/errno.h>
```
sunfishcode, line 3: In cloudlibc, this header handles translating from cloudabi error codes such as `ENOTCAPABLE` to POSIX error codes, but in wasi-libc, we decided to just have the WASI calls return the desired error codes directly, so this header and `errno_fixup_socket` doesn't do anything. I think if we removed this header, it would be straightforward to move this file into libc-bottom-half/sources.

sunfishcode (commented):
Overall looks good!

When adding new files, such as this accept.c, we should add them in libc-bottom-half/sources, since they aren't derived from cloudlibc. See below for how to resolve the common/errno.h header:

libc-bottom-half/cloudlibc/src/libc/sys/socket/accept.c
haraldh: done

sunfishcode:
Looks good!

libc-bottom-half/sources/accept.c
```diff
@@ -0,0 +1,51 @@
...
+  addr->sa_family = AF_UNSPEC;
+  *addrlen = sizeof(struct sockaddr);
+
+  return ret;
+}
+
+int accept4(int socket, struct sockaddr *restrict addr, socklen_t *restrict addrlen, int flags) {
+  int ret = -1;
+
+  if (flags & ~(SOCK_NONBLOCK | SOCK_CLOEXEC)) {
```
codefromthecrypt, line 32: I suppose the rationale for validating, but dropping the only flag besides nonblock is because it isn't in wasi, right? https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md#fdflags cc @deadprogram who's implementing tinygo https://github.com/tinygo-org/tinygo/pull/2748 🎖️

codefromthecrypt: sorry tagged wrong person meant @rvolosatovs :blush:

rvolosatovs: Yes, this behavior is not relevant for the TinyGo implementation, since this is a wrapper for `sock_accept`, whereas in the TinyGo PR we call directly into WASI

PR #288, opened 2022-05-05
Override Make's default for CC and AR.

glandium:


sunfishcode:
I can confirm that `make` works this way. I expect @Ericson2314's use case is explicitly overriding `CC` and `AR`. I'm going to merge this, to unbreak the tree for users, but @Ericson2314 please feel free to comment if this isn't intended.

PR #291, opened 2022-05-24
Fix utimensat to avoid passing uninitialized values into WASI calls.

sunfishcode:
Previously, utimensat would leave the mtim and/or atim timestamps
uninitialized when the `MTIM_NOW` or `ATIM_NOW` were in use, because
that means the respective timestamps are not used.

However, clang now automatically adds `noundef` to the arguments in
functions like `__wasi_path_filestat_set_times`, and there are cases
where simplifycfg can see paths where the uninitialized values are
passed to those `noundef` arguments.

To fix this, change the utimens code to zero out the timestamps when
they aren't in use, to avoid passing uninitialized arguments.


sbc100:
Why does clang automatically add `noundef` to those argument?

sunfishcode:
I'm not current on the latest thinking about uninitialized variables, but I imagine it's considered UB in C to pass uninitialized values as function arguments. Clang 14 appears to automatically add noundef to all C function arguments.

sbc100 (approved):
Fair enough.

PR #294, opened 2022-05-25
Delete several blocks of unused code.

sunfishcode:
Delete several pieces of code from libc-bottom-half/cloudlibc that aren't in
use on wasi-libc.


sunfishcode:
This deletes several blocks of code, making it easier to figure out what code is being used in wasi-libc and what code is not. In particular, it clarifies that `errno_fixup_directory` and `errno_fixup_socket` are no longer needed.

PR #296, opened 2022-06-02
Split import declaration and glue function for each function

kateinoigakukun:
If a program links any WASI function, `__wasilibc_real.c` is linked and it brings import entries for all WASI functions.
This patch allows optional import for WASI function. It's useful for avoiding imports of newly added features like `sock_accept`.


sbc100:
Can you explain eactly how/when this helps?   My understanding (as the author of wasm-ld) is that only the imports that you actually use from `__wasilibc_real.c` should end up as imports in the final binary.

Your final binary should never import a function it doesn't actually need.    Unless maybe you are building with `--no-gc-sections`?

kateinoigakukun:
@sbc100 Sorry for lack of context. And I forgot the fact that we are using `--no-gc-sections`...

The motivation of this patch was for Swift, which depends on `--no-gc-sections`.
Currently Swift runtime heavily depends on `__start_` and `__stop_` symbols for named data segments to scan matadata in data segments dynamically. However, wasm-ld doesn't mark a data segment live if `__start_` `__stop_` symbols for the segment are used. but no symbol in the segment is used.
So Swift compiler driver passes `--no-gc-sections` to linker as a workaround to avoid stripping those data segments. (and I forgot that fact.. 😓 sorry)

I realized that the right solution is telling wasm-ld to mark the `__start_`/`__stop_` symbols live.

sbc100:
> @sbc100 Sorry for lack of context. And I forgot the fact that we are using `--no-gc-sections`...
> 
> The motivation of this patch was for Swift, which depends on `--no-gc-sections`. Currently Swift runtime heavily depends on `__start_` and `__stop_` symbols for named data segments to scan matadata in data segments dynamically. However, wasm-ld doesn't mark a data segment live if `__start_` `__stop_` symbols for the segment are used. but no symbol in the segment is used. So Swift compiler driver passes `--no-gc-sections` to linker as a workaround to avoid stripping those data segments. (and I forgot that fact.. sweat sorry)
> 
> I realized that the right solution is telling wasm-ld to mark the `__start_`/`__stop_` symbols live.

Is there some change to wasm-ld that we could make that would address this issue?   Are you saying if that you would like all data that lives in segment foo to be kept alive if the `__start_foo` and/or `__stop_foo` is referenced?

Perhaps we could add a `--no-gc-data-sections` flag that would only effect data and not function symbols?

kateinoigakukun:
> Is there some change to wasm-ld that we could make that would address this issue?

I haven't made a patch yet, but IIUC lld for ELF does it [here](https://github.com/llvm/llvm-project/blob/5c6ed60c517c47b25b6b25d8ac3666d0e746b0c3/lld/ELF/MarkLive.cpp#L355-L361)

> Are you saying if that you would like all data that lives in segment foo to be kept alive if the __start_foo and/or __stop_foo is referenced?

I mean wasm-ld should keep data segments if any symbol in an object file, which contains the data segment, is referenced  and start/stop symbols are referenced.

For example, think the following situation:

```c
// in foo.c, which is archived as libfoo.a
__attribute__((__section__("foo_md")))
char FOO_MD[] = "bar";

void foo_func(void) {}

// in main.c, which links foo via -lfoo

#include <stdio.h>

extern const char __start_foo_md;
extern const char __stop_foo_md;

extern void foo_func(void);

int main(void) {
    printf("__start_foo_md = %p, __stop_foo_md = %p\n", &__start_foo_md, & __stop_foo_md);
    foo_func();
    return 0;
}
```

Then, compile it by

```console
$ clang -c foo.c && llvm-ar cr libfoo.a foo.o
$ clang main.c -lfoo -L. -fuse-ld=lld
```

It works with ELF lld. The `foo.o` is linked since `foo_func` is referenced from `main.o`, and section `foo_md` is also linked.
But it doesn't work with wasm-ld, with following message:

```
wasm-ld: error: /tmp/main-c796ab.o: undefined symbol: __start_foo_md
wasm-ld: error: /tmp/main-c796ab.o: undefined symbol: __stop_foo_md
```

kateinoigakukun:
> Or perhaps it would be enough to make all your metadata as "used". I believe this maps to the WASM_SYMBOL_NO_STRIP flag which prevents gc-sections from removing it:

I think it also works, but I prefer marking sections in linker to follow what ELF lld does.

PR #300, opened 2022-07-08
Add a `getpagesize` function.

sunfishcode:
This adds a `getpagesize` function. This interface is deprecated in POSIX,
but it's sufficiently widely used and not problematic in practice.


sbc100:
Why not just compile the existing `libc-top-half/musl/src/legacy/getpagesize.c`?  Why re-implement?

sbc100 (approved):

sunfishcode:
I've now switched it to use the musl version.

PR #303, opened 2022-07-18
[WIP] Fixes for the THREAD_MODEL=posix build

sbc100:


sbc100:
Still need to deal with preopens.c

libc-bottom-half/sources/getentropy.c
```diff
@@ -1,18 +1,18 @@
...
 int __getentropy(void *buffer, size_t len) {
+    int cs;
     if (len > 256) {
         errno = EIO;
         return -1;
     }
 
+    pthread_setcancelstate(PTHREAD_CANCEL_DISABLE, &cs);
     int r = __wasi_random_get(buffer, len);
+    pthread_setcancelstate(cs, 0);
```
sunfishcode, line 15: I propose that we leave out considerations of cancellation out for now. Pthread cancellation is something of an anti-sweet-spot, being particularly difficult and expensive to implement, while being particularly rarely used.

One of the particular complexities is what to do if one thread tries to cancel another thread that's currently executing host code. That might require all host code to be pthread cancel-safe, which would be very difficult for many popular host implementations today.

libc-bottom-half/sources/environ.c
```diff
@@ -10,9 +10,9 @@
 // that it's initialized whenever user code might want to access it.
 char **__wasilibc_environ;
 extern __typeof(__wasilibc_environ) _environ
-    __attribute__((weak, alias("__wasilibc_environ")));
+    __attribute__((__weak__, alias("__wasilibc_environ")));
 extern __typeof(__wasilibc_environ) environ
-    __attribute__((weak, alias("__wasilibc_environ")));
```
sunfishcode, line 15: I'm not opposed to changing these `weak`s to `__weak__`s, though (a) could you split this out into a separate PR, and (b) it's worth noting that most of these aren't header files, so so don't strictly need most of these.

libc-top-half/musl/arch/wasm32/pthread_arch.h
```diff
@@ -1,12 +1,11 @@
-#ifdef _REENTRANT
-#error "multiple threads not supported in musl yet"
+uintptr_t __get_tp(void) {
+#if _REENTRANT
+  int val;
+  __asm__(".globaltype g_needs_dynamic_alloc, i32\n"
+          "global.get __wasi_libc_pthread_self\n"
```
sunfishcode, line 5: Where are `g_needs_dynamic_alloc` and `__wasi_libc_pthread_self` defined?

As a minor style nit, wasi-libc usually uses identifiers named `__wasilibc_*` rather than `__wasi_libc_*`.

libc-bottom-half/cloudlibc/src/libc/time/nanosleep.c
```diff
@@ -5,6 +5,7 @@
 #include <errno.h>
 #include <threads.h>
 #include <time.h>
+#include <_/cdefs.h> // for __strong_reference
```
sunfishcode, line 8: Instead of adding this include, let's just delete the `thrd_sleep` alias here. If we want `thrd_sleep`, it's better to use the musl definition, which correctly implements the C11 behavior.

libc-bottom-half/sources/chdir.c
```diff
@@ -90,18 +92,30 @@ static const char *make_absolute(const char *path) {
...
         make_absolute_len = alloc_len;
     }
     strcpy(make_absolute_buf, __wasilibc_cwd);
+    __wasilibc_cwd_unlock();
+
+#ifdef _REENTRANT
+    if (path[0] == 0 || !strcmp(path, ".") || !strcmp(path, "./")) {
+        return make_absolute_buf;
+    }
+#endif
```
sunfishcode, line 117: Could you add a comment about what this block of code does?

libc-top-half/musl/arch/wasm32/pthread_arch.h
```diff
@@ -1,12 +1,11 @@
...
+          "local.set %0" : "=r" (val));
+  return val;
+#else
+  return 0;
```
sunfishcode, line 9: Does `__get_tp` need to be defined in non-`_REENTRANT` mode? It seems better to get an undefined-symbol link error if we try to use it than to have it silently return NULL, especially with NULL being a valid address.

sbc100: Thanks for all the comments.  I will address all of them, this PR is certainly not ready to land.  I just wanted to share what I had done so far.

sunfishcode: Yep, makes sense. I was just looking through it so that I could follow along, and happened to have comments along the way :-).

sbc100:
Closing in favor of #311

libc-bottom-half/sources/getentropy.c
sbc100: Sounds reasonable, at least for a first pass.   How about we just avoid using `pthread_setcancelstate` et al in the bottom half and just stub them out (make them no-ops) when compiling top-half code.

PR #306, opened 2022-07-25
Use MUSL's `weak*` feature in bottom half

abrown:
This change extracts the `weak*`-related parts of #303 as a separate PR.
Note that this is slightly strange in that it uses some top-half MUSL
headers in the bottom-half code, but discussion around this led me to
believe that the advantages of, e.g., `LOCK` made this worthwhile.
Beyond just changing uses of `weak` to `__weak__`, we also MUSL's `weak`
and `weak_alias` macros in a few more places.


libc-bottom-half/headers/public/wasi/libc-find-relpath.h
```diff
@@ -70,7 +70,7 @@ int __wasilibc_find_relpath_alloc(
     char **relative,
     size_t *relative_len,
     int can_realloc
-) __attribute__((weak));
+) __attribute__((__weak__));
```
abrown, line 73: Here is a place that I could have used just `weak` but when I do the build fails:

```
#
# Test that it compiles.
#
clang -O2 -DNDEBUG --target=wasm32-wasi -fno-trapping-math -Wall -Wextra -Werror -Wno-null-pointer-arithmetic -Wno-unused-parameter -Wno-sign-compare -Wno-unused-variable -Wno-unused-function -Wno-ignored-attributes -Wno-missing-braces -Wno-ignored-pragmas -Wno-unused-but-set-variable -Wno-unknown-warning-option -mthread-model single -isystem "/home/abrown/Code/wasi-libc/sysroot/include" -fsyntax-only "/home/abrown/Code/wasi-libc/sysroot/share/wasm32-wasi/include-all.c" -Wno-\#warnings
In file included from /home/abrown/Code/wasi-libc/sysroot/share/wasm32-wasi/include-all.c:166:
/home/abrown/Code/wasi-libc/sysroot/include/wasi/libc-find-relpath.h:73:3: error: expected function body after function declarator
) weak;
  ^
1 error generated.
```

I guess there is some other place to add an include path if we want to use `weak` everywhere. (Not sure I'm fully convinced that `weak` is better than `__attribute((__weak__))`, though... Just trying to be consistent.)

sbc100 (approved):
Nice!

libc-bottom-half/headers/public/wasi/libc-find-relpath.h
sbc100: Adding `#include <features.h>` should do it.

abrown: That seemed like the right thing to do so I tried it but it blew up with the same error...

sunfishcode:
LGTM!

libc-bottom-half/sources/__wasilibc_initialize_environ.c
```diff
@@ -11,7 +11,7 @@
 /// Statically-initialize it to an invalid pointer value so that we can
 /// detect if it's been explicitly initialized (we can't use `NULL` because
 /// `clearenv` sets it to NULL.
-char **__wasilibc_environ __attribute__((weak)) = (char **)-1;
+char **__wasilibc_environ weak = (char **)-1;
```
sbc100, line 14: I think this reads way easier with `weak` at the start of the line.. and I think that seems to be how musl uses it.

libc-bottom-half/sources/__wasilibc_initialize_environ.c
```diff
@@ -87,7 +87,7 @@ void __wasilibc_deinitialize_environ(void) {
 }
 
 // See the comments in libc-environ.h.
-__attribute__((weak))
+weak
 void __wasilibc_maybe_reinitialize_environ_eagerly(void) {
```
sbc100, line 91: Combine these onto one line?

PR #308, opened 2022-07-26
Enable socket options definitions

### LINKED ISSUE #304: Provide socket option definitions
Hi,
I've noticed there's no definition in [__header_sys_socket.h](https://github.com/WebAssembly/wasi-libc/blob/main/libc-bottom-half/headers/public/__header_sys_socket.h) for [socket options](https://github.com/WebAssembly/wasi-libc/blob/main/libc-top-half/musl/include/sys/socket.h#L209). Is that intentional? I see other definitions (e.g. `AF_INET`) are already added to the header, but I can't find it anywhere in the WASI spec (so I'm not sure why e.g. AF_INET is defined as 1).
If lack of those definitions is not intentional, I can provide a PR for that; however, I'm not sure what numbers should be assigned to a specific definition - is there any specification already? (as I said, I couldn't find one for AF_INET either).

Many thanks

loganek:
Fixes #304

### DIFF
--- a/expected/wasm32-wasi-threads/predefined-macros.txt
+++ b/expected/wasm32-wasi-threads/predefined-macros.txt
@@ -33,10 +33,10 @@
 #define ADJ_TAI 0x0080
 #define ADJ_TICK 0x4000
 #define ADJ_TIMECONST 0x0020
-#define AF_INET 1
-#define AF_INET6 2
+#define AF_INET PF_INET
+#define AF_INET6 PF_INET6
 #define AF_UNIX 3
-#define AF_UNSPEC 0
+#define AF_UNSPEC PF_UNSPEC
 #define ALT_DIGITS 0x2002F
 #define AM_STR 0x20026
 #define ANYMARK 0x01
@@ -1315,6 +1315,9 @@
 #define PAGE_SIZE PAGESIZE
 #define PATH_MAX 4096
 #define PDP_ENDIAN __PDP_ENDIAN
+#define PF_INET 1
+#define PF_INET6 2
+#define PF_UNSPEC 0
 #define PM_STR 0x20027
 #define POLLERR 0x1000
 #define POLLHUP 0x2000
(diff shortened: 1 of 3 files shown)


sunfishcode:
My main question here is whether it's better to define *all* of these macros, or to selectively define only the ones which WASI is expected to support. For example, let's say something like `SO_ATTACH_BPF` doesn't happen for the foreseeable future. Is it better for WASI to define it and have it fail at runtime, or to leave it undefined?

loganek:
> My main question here is whether it's better to define _all_ of these macros, or to selectively define only the ones which WASI is expected to support. For example, let's say something like `SO_ATTACH_BPF` doesn't happen for the foreseeable future. Is it better for WASI to define it and have it fail at runtime, or to leave it undefined?

Yeah, I did think about it; on one hand it's better to handle unimplemented properties as early as possible (in that case, at compilation time). On the other hand, that will make porting code to WASM more difficult. This will be especially hard for the code that doesn't use non-supported properties, but e.g. relies on 3rd party dependency that under the hood uses it in some branches that are very unlikely to be hit. Not sure if there's a general guidance or if there were similar situations before - we either prioritize safety and avoid runtime errors (even for properties that aren't likely to be used anyway) or we lean towards ease of porting existing codebase. I'm personally less concerned about runtime errors as this can be handled at the testing stage; at the same time I'm aware of the status of test coverage for multiple projects around... I'm curious to know what are the community thought on that.

sbc100:
We could always introduce more macros like `_WASI_EMULATED_SIGNAL`.. perhaps we could have a single one that enables maximum source compatibility, for folks that just want to get something building?

yamt:
> > My main question here is whether it's better to define _all_ of these macros, or to selectively define only the ones which WASI is expected to support. For example, let's say something like `SO_ATTACH_BPF` doesn't happen for the foreseeable future. Is it better for WASI to define it and have it fail at runtime, or to leave it undefined?
> 
> Yeah, I did think about it; on one hand it's better to handle unimplemented properties as early as possible (in that case, at compilation time). On the other hand, that will make porting code to WASM more difficult. This will be especially hard for the code that doesn't use non-supported properties, but e.g. relies on 3rd party dependency that under the hood uses it in some branches that are very unlikely to be hit. Not sure if there's a general guidance or if there were similar situations before - we either prioritize safety and avoid runtime errors (even for properties that aren't likely to be used anyway) or we lean towards ease of porting existing codebase. I'm personally less concerned about runtime errors as this can be handled at the testing stage; at the same time I'm aware of the status of test coverage for multiple projects around... I'm curious to know what are the community thought on that.

while i can understand the argument for things like SO_LINGER,
i don't think it makes much sense to have very linux-specific stuff, which is not portable regardless of wasi.

Issue #209, opened 2020-07-14
Add pthread.h in wasi sysroot to support pthread


xujuntwt95329:
Hi,

The pthread library has been supported in [wasm micro runtime](https://github.com/bytecodealliance/wasm-micro-runtime). The pthread related APIs in wasm application should be imported from the runtime, such as pthread_create, pthread_join, etc. so the runtime can create an actual thread to execute the wasm function.

For example, suppose we have such a C code:
``` C
#include <pthread.h>

void *func(void *arg)
{

}

int main()
{
    pthread_t tid;
    pthread_create(&tid, NULL, func, NULL);
    return 0;
}
```

The `pthread_create` should be compiled as an import function:
``` wasm
(import "env" "pthread_create" (func $pthread_create (type $t1)))
......
```

In this situation, I think there should be a simple way to support pthread in wasi-sdk as we only need a few things:
1. type definition for pthread_t, pthread_mutex_t and so on
2. function prototype for pthread APIs such as pthread_create, pthread_join, pthread_self, etc.
3. undefined-symbol lists

To achieve 1 and 2, we just need a `pthread.h` under `sysroot/include`, just like [WAMR's pthread.h](https://github.com/bytecodealliance/wasm-micro-runtime/blob/main/wamr-sdk/app/libc-builtin-sysroot/include/pthread.h), we also need to undef some macros like `__NEED_pthread_t` to avoid duplicate definition with `sysroot/include/bits/alltypes.h`
And to achieve 3, we need to add pthread API symbols into `sysroot/share/wasm32-wasi/undefined-symbols.txt` just like wasi APIs

In this way, we don't need to provide libpthread.a in wasi-libc, as the pthread related APIs should be implemented by the runtime.

Do you think this is a reasonable way to support pthread ?

xujuntwt95329:
Furthermore, I'm trying to build some C++ workload which use `std::thread`, but I get this error:
> error: `<atomic>` is not supported on this single threaded system

I noticed that the `std::thread` depends on pthread. In this case, if we have supported pthread, is there any other thing to do to support `std::thread` ?

Thanks a lot !

sunfishcode:
The current situation is that there is a consensus among the people working on threads support in the WebAssembly core spec is that the current semantics are not yet suitable for use outside of Web contexts. See the following issues for background:

https://github.com/WebAssembly/threads/issues/8
https://github.com/WebAssembly/threads/issues/95
[...]

PR #311, opened 2022-07-27
Fix `make THREAD_MODEL=posix`

abrown:
This adds my changes on top of #303 in order to have `make THREAD_MODEL=posix` build successfully.

Fixes: #209


sbc100:
Should I abandon #303 now in favor of this?  (seems like I should)

expected/wasm32-wasi/defined-symbols.txt
```diff
@@ -78,6 +81,48 @@ __fwritable
...
+__get_tp
+__get_tp
+__get_tp
+__get_tp
+__get_tp
+__get_tp
+__get_tp
+__get_tp
+__get_tp
+__get_tp
```
sbc100, line 125: I guess we need to run `uniq` on this list?

expected/wasm32-wasi/predefined-macros.txt
```diff
@@ -2290,53 +2324,53 @@
...
+#define _TIME_H
+#define _UCHAR_H
+#define _UNISTD_H
+#define _UTIME_H
+#define _VALUES_H
+#define _VA_LIST
+#define _WCHAR_H
+#define _WCHAR_T
+#define _WCTYPE_H
+#define _WINT_T
```
sbc100, line 2373: Whats up with the whitespace changes here?

abrown:
> Should I abandon #303 now in favor of this? (seems like I should)

Probably, yes... this contains your commit but rebased on the latest `main`.

expected/wasm32-wasi/predefined-macros.txt
abrown: Probably my IDE ripping out trailing whitespace? Should it not?

sbc100: These files are auto-generated, and should not be edited by hand.  If you want to remove the spaces (not sure its worth doing) then I would suggest doing is part of that auto-generation process.

sbc100: If you do want to do this we should make it a separate PR to keep this one focused.

(BTW I personally recommend against automatically stripping trailing white space on save, and instead binding a keyboard shortcut to it.. then you are in control of if/when you want to do the stripping, which makes it easier to work on lots of different projects, when not all of them are already free of trailing whitespace.  I use leader-w in vim for this).

Makefile
```diff
@@ -597,7 +597,7 @@ check-symbols: startup_files libc
 
 	# Check that the computed metadata matches the expected metadata.
 	# This ignores whitespace because on Windows the output has CRLF line endings.
-	diff -wur "$(CURDIR)/expected/$(MULTIARCH_TRIPLE)" "$(SYSROOT_SHARE)"
+	diff -wur "$(CURDIR)/expected/$(MULTIARCH_TRIPLE)/$(THREAD_MODEL)" "$(SYSROOT_SHARE)"
```
sbc100, line 600: I'm not sure this change in necessary, or at least we wouldn't want the default build to install into single.

I'm hoping that eventually we can just ship a single copy of libc since modern engines can handler lowering atomic accesses to non-atomics ones, and the linker can lower TLS to regular static data.

abrown: I'm not sure what you're proposing... is there some other way in the current state of the PR to have all the same symbols for both thread models?

expected/wasm32-wasi/posix/defined-symbols.txt
```diff
@@ -199,22 +200,17 @@ __sinl
 __small_printf
 __stderr_FILE
 __stderr_used
-__stderr_used
```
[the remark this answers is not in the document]
abrown: I could but that's sort of a secondary concern right now. Things that need to get figured out:
- [ ] how to lock `preopens.c` (see my Zulip question), 
- [ ] how best to expose MUSL's `thrd_sleep` instead of cloud ABI's (https://github.com/WebAssembly/wasi-libc/pull/303#discussion_r923921656)
- [ ] where to define, e.g., `__wasilibc_pthread_self` (https://github.com/WebAssembly/wasi-libc/pull/303#discussion_r923920481)
- [ ] remove uses thread cancellation (https://github.com/WebAssembly/wasi-libc/pull/303#discussion_r923918195)
- [ ] decide whether `__get_tp` should even be a defined symbol in single-threaded model (https://github.com/WebAssembly/wasi-libc/pull/303#discussion_r923923837)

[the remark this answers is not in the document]
sbc100: 1. For pre-opens yes, its seems reasonable to use fairly coarse grain locking, at least initially.
2. Seems like we can just remove the `thrd_sleep` alias from `libc-bottom-half/cloudlibc/src/libc/time/nanosleep.c` and start compiling `libc-top-half/musl/src/thread/thrd_sleep.c` which should work for both posix and single threaded builds (that can be its one change if you are looking to pieces of this to land right away).
3. `/libc-top-half/musl/src/thread/pthread_self.c` seems like a reasonable place to define this global.
4. Seems reasonable to skip over thread cancellation initially.  No need to add any calls to the bottom half, and for the top half we can just stub out the impl.
5. I think its reasonable to define `__get_tp` to just always returns a constant in the single threaded build.
6. Indeed, it looks like we should maybe just skip building the threaded build for clang v10 and only build it for recent clang.  The CI only builds with clang v10 to tests that we work with older clang, and I think its reasonable for the threaded build to require a more recent clang.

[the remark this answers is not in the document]
abrown: > 3. `/libc-top-half/musl/src/thread/pthread_self.c` seems like a reasonable place to define this global.

Can you show me a place that demonstrates the right way to do this for the WebAssembly target? I'm just not super familiar with how to map a WebAssembly global directly in C...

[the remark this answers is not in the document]
sbc100: You have to use inline assembly, or in a separate `.s` assembly file.  We recently starting doing it in emscripten here:

https://github.com/emscripten-core/emscripten/blob/517cc860e204b8e659c41649e5542d534bc5cad8/system/lib/pthread/emscripten_tls_init.c#L34-L40

expected/wasm32-wasi/single/defined-symbols.txt
```diff
@@ -37,6 +37,9 @@ __des_setkey
 __do_des
 __duplocale
 __env_rm_add
```
sbc100, line 39: This looks out-of-date.

libc-bottom-half/sources/chdir.c
```diff
@@ -8,10 +8,8 @@
 #include <wasi/libc-find-relpath.h>
 #include <wasi/libc.h>
 
-#ifdef _REENTRANT
-#error "chdir doesn't yet support multiple threads"
-#endif
-
+void __wasilibc_cwd_lock(void);
```
sbc100, line 11: Wrap these in `#if _REENTRANT`?

We could do `#define __wasilibc_cwd_lock()` in the else block if you want to avoid more ifdefs in the code?

libc-bottom-half/sources/getcwd.c
```diff
@@ -1,30 +1,41 @@
...
-// For threads this needs to synchronize with chdir
-#ifdef _REENTRANT
-#error "getcwd doesn't yet support multiple threads"
-#endif
+#include "lock.h"
 
 char *__wasilibc_cwd = "/";
+static volatile int lock[1];
+
+void __wasilibc_cwd_lock(void)
```
sbc100, line 9: Wrap these in `#ifdef _REENTRANT`

libc-top-half/musl/src/thread/pthread_self.c
```diff
@@ -1,6 +1,12 @@
 #include "pthread_impl.h"
 #include <threads.h>
 
+#if defined(_REENTRANT)
```
sbc100, line 4: `if defined(__wasm__) &&  defined`_REENTRANT)`

Also wrap in `__wasilibc_unmodified_upstream`?

libc-top-half/musl/src/thread/pthread_setcancelstate.c
```diff
@@ -2,10 +2,12 @@
 
 int __pthread_setcancelstate(int new, int *old)
 {
+	#ifdef _REENTRANT
```
sbc100, line 5: #ifdef should be in column zero I think.

Wrap in `__wasilibc_unmodified_upstream`

sbc100 (commented):
Looking good! I think we can land this with a couple of nits fixed!

expected/wasm32-wasi/posix/include-all.c
```diff
@@ -0,0 +1,171 @@
+#include <__errno.h>
```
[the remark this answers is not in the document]
abrown: Can we do a follow-up to that later? It's quite convenient and clear to use `$(THREAD_MODEL)` throughout the Makefile...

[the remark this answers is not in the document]
sbc100: I'd rather not rename these expectations file just to rename them again later.

Can you find a way to at least leave the existing files in place?

How about something like this:

```
#if THREAD_MODULE == posix
EXPECTATION_SUBDIR = '/pthreads'
#else
EXPECTATION_SUBDIR = ''
#endif
```

I know that isn't Makefile syntax but you get the idea.  This way the old files stay in place, and the new files have more meaningful names.

Makefile
```diff
@@ -184,6 +185,14 @@ LIBC_TOP_HALF_MUSL_SOURCES = \
                  %/cimagf.c %/cimag.c %cimagl.c, \
                  $(wildcard $(LIBC_TOP_HALF_MUSL_SRC_DIR)/complex/*.c)) \
     $(wildcard $(LIBC_TOP_HALF_MUSL_SRC_DIR)/crypt/*.c)
+
+# Add some additional sources when threads are enabled.
+ifeq ($(THREAD_MODEL), posix)
+LIBC_TOP_HALF_MUSL_SOURCES += $(addprefix $(LIBC_TOP_HALF_MUSL_SRC_DIR)/, \
+    thread/pthread_setcancelstate.c \
```
sbc100, line 192: I think at the very least you will need to include `thread/__lock.c` here

libc-bottom-half/sources/preopens.c
```diff
@@ -32,6 +29,10 @@ static preopen *preopens;
 static size_t num_preopens;
 static size_t preopen_capacity;
 
+/// Access to the the above preopen must be protected in the presence of
+/// threads.
+static volatile int lock[1];
```
sbc100, line 34: Put this inside of `ifdef _REENTRANT` .. also perhaps `#define` LOCK and UNLOCK to no-op in the else case?

i.e. do the same thing we did for `__wasilibc_cwd_unlock`?

libc-top-half/musl/src/thread/pthread_setcancelstate.c
```diff
@@ -2,11 +2,13 @@
...
-	if (new > 2U) return EINVAL;
-	struct pthread *self = __pthread_self();
-	if (old) *old = self->canceldisable;
-	self->canceldisable = new;
-	return 0;
+#if defined(__wasilibc_unmodified_upstream) || defined(_REENTRANT)
+    if (new > 2U) return EINVAL;
+    struct pthread *self = __pthread_self();
+    if (old) *old = self->canceldisable;
+    self->canceldisable = new;
```
sbc100, line 9: Leave the indentation alone here to keep the diff small?

libc-top-half/musl/src/thread/thrd_sleep.c
```diff
@@ -1,11 +1,17 @@
...
 #include "syscall.h"
+#endif
 
 int thrd_sleep(const struct timespec *req, struct timespec *rem)
 {
+#ifdef __wasilibc_unmodified_upstream
 	int ret = -__clock_nanosleep(CLOCK_REALTIME, 0, req, rem);
+#else
+	int ret = -clock_nanosleep(CLOCK_REALTIME, 0, req, rem);
+#endif
```
sbc100, line 14: Add `weak_alias(__clock_nanosleep, clock_nanosleep);` to the file where clock_nanosleep is defined (like musl does) ... to avoid this diff?

sbc100 (commented):
Getting pretty close.. have you tried actually linking any programs against this yet?  Or should we not even make that a goal of this intial PR?

Makefile
sbc100: Actually I think you should probably include that majority of the files in the `thread/` directory?  But maybe you want to save that for a followup PR?

abrown: I tried to include more files here but it was a game of whack-a-mole (i.e., the build failures). There are many other bits that need to be fixed so I'm hoping this PR can be merged with this very limited subset of the pthreads API and future PRs can add more of the API.

sbc100: Sounds like a good plan.

In that case, maybe just remove this block completely for now? (We really need `pthread_setcancelstate` do we?)

abrown: Yeah, I can just remove it.

expected/wasm32-wasi/posix/include-all.c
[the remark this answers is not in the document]
abrown: Ok, I tried this and the issue is that we are using `diff` recursively so a structure like the following doesn't really work:
```
- expected
--- wasm-32-wasi
----- defined-symbols.txt
----- ...
----- posix (or pthreads)
------- defined-symbols.txt
------- ...
```

The posix/pthreads directory gets in the way of the `diff`.

libc-bottom-half/sources/preopens.c
abrown: `LOCK` and `UNLOCK` are already defined this way; I guess I'll add `#ifdef _REENTRANT` around this `lock` but it seems a bit weird since in the no-threads case `lock` will not be defined at all.

libc-top-half/musl/src/thread/thrd_sleep.c
abrown: Hm, not sure I understand the "why" of what you're proposing: then we would have a weak alias in one direction in the bottom half and another weak alias in the other direction in the top half? (See `libc-top-half/musl/src/time/clock_nanosleep.c`).

sbc100: We don't compile the top half version of clock_nanosleep right?  IIUC it gets completely replaced by the bottom half version in `libc-bottom-half/cloudlibc/src/libc/time/clock_nanosleep.c`, right?

I'm suggesting that you make the bottom half version match the top half version by adding `weak_alias(__clock_nanosleep, clock_nanosleep);` at the end the file.. then this patch would not need needed.

libc-bottom-half/sources/preopens.c
sbc100: Oh great, i forget LOCK/UNLOCK were already defined like that.

sbc100: I that case I guess `__wasilibc_cwd_lock/unlock` are not actually needed at all .. no harm I guess in naming the macro but we could just use LOCK/UNLOCK

libc-top-half/musl/src/thread/thrd_sleep.c
abrown: Ok, didn't understand the "non-compiled top-half bits" of this comment; thanks. I ad

expected/wasm32-wasi/posix/include-all.c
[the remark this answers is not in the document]
abrown: It seems like we'll need two separate directory trees to `diff` with. Here are some options:
 a. append `/pthreads` or `/no-pthreads` to the expected directory path; this at least means that the naming is a bit more accurate
 b. append something, e.g., `pthreads`, to the target, so we would have `wasm32-wasi` and `wasm32-wasi-pthreads`

libc-bottom-half/sources/getentropy.c
```diff
@@ -1,12 +1,9 @@
...
 #include <errno.h>
 #include <unistd.h>
-
-#ifdef _REENTRANT
-#error With threads support, getentropy is not intended to be a cancellation point.
-#endif
+#include <wasi/api.h>
 
 int __getentropy(void *buffer, size_t len) {
+    int cs;
```
sbc100, line 6: Remove this line?

sbc100 (approved):
lgtm with one last nit

libc-bottom-half/sources/getentropy.c
abrown: Not sure how this got here... removed.

PR #313, opened 2022-07-31
Sort lists of input files

bmwiedemann:
Sort lists of input files
for reproducible `.a` files in spite of non-deterministic
filesystem readdir order

Without this patch, find returned files in filesystem order
and llvm-ar used that order to create .a files.

This PR was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).


sbc100 (approved):
Make sense to me

PR #316, opened 2022-08-22
threads: implement support for unnamed semaphores

abrown:
[POSIX semaphores] come in two forms: named and unnamed. Roughly, named
semaphores use files to implement locking across processes; unnamed
semaphores use a shared memory region to implement locking across
threads in the same process. Since WASI currently has no process concept
(and it is relatively unclear how to map the WASI files as shared
memory), only the unnamed semaphores are supported by this changed. This
means that `sem_open`, `sem_close`, and `sem_unlink` will not available
to programs compiled with a threads-enabled `wasi-libc`.

[POSIX semaphores]: https://man7.org/linux/man-pages/man7/sem_overview.7.html

### DIFF
--- a/libc-top-half/musl/src/thread/sem_timedwait.c
+++ b/libc-top-half/musl/src/thread/sem_timedwait.c
@@ -8,7 +8,9 @@ static void cleanup(void *p)
 
 int sem_timedwait(sem_t *restrict sem, const struct timespec *restrict at)
 {
+#ifdef __wasilibc_unmodified_upstream
 	pthread_testcancel();
+#endif	
 
 	if (!sem_trywait(sem)) return 0;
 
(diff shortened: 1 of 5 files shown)


libc-top-half/musl/src/thread/sem_timedwait.c
```diff
@@ -8,7 +8,9 @@ static void cleanup(void *p)
 
 int sem_timedwait(sem_t *restrict sem, const struct timespec *restrict at)
 {
+#ifdef __wasilibc_unmodified_upstream
 	pthread_testcancel();
+#endif
```
sbc100, line 13: `src/thread/pthread_testcancel.c` is super trivial so maybe easier to just include that now rather than add ifdef here.

See https://github.com/WebAssembly/wasi-libc/blob/main/libc-top-half/musl/src/thread/pthread_testcancel.c

sbc100 (approved):
lgtm % comment

PR #320, opened 2022-09-02
threads: implement support for pthread_condattr

loganek:


Makefile
```diff
@@ -264,6 +268,9 @@ ifeq ($(THREAD_MODEL), posix)
 # Specify the tls-model until LLVM 15 is released (which should contain
 # https://reviews.llvm.org/D130053).
 CFLAGS += -mthread-model posix -pthread -ftls-model=local-exec
+
+# Include cloudlib's directory to access the structure definition of clockid_t
+CFLAGS += -I$(LIBC_BOTTOM_HALF_CLOUDLIBC_SRC)
```
loganek, line 273: For `pthread_condattr_setclock` (in top half) we need a definition of the `clockid_t` struct, which is defined in bottom half. Definition of `clockid_t` at least for now is not required to be a part of public interface, therefore I don't think we have to make it public at that stage. At the same time, I'm a bit worried that adding the whole cloudlib source dir as include by default to top half will lead to some weird usage if we're not careful enough. Another option would be to have a dedicated folder with headers shared between bottom and top half (but not available in public interface).

I'm open for suggestions.

sbc100: This solution seems ok to me.

libc-top-half/musl/src/thread/pthread_condattr_setclock.c
```diff
@@ -1,9 +1,11 @@
...
+#include <common/clock.h>
+
 int pthread_condattr_setclock(pthread_condattr_t *a, clockid_t clk)
 {
-	if (clk < 0 || clk-2U < 2) return EINVAL;
+	if (clk->id < 0 || clk->id-2U < 2) return EINVAL;
 	a->__attr &= 0x80000000;
-	a->__attr |= clk;
+	a->__attr |= clk->id;
 	return 0;
```
sbc100, line 10: For changes to musl sources like this we wrap them in `__wasilibc_unmodified_upstream`

libc-top-half/musl/src/thread/pthread_condattr_setclock.c
```diff
@@ -1,9 +1,11 @@
 #include "pthread_impl.h"
 
+#include <common/clock.h>
+
 int pthread_condattr_setclock(pthread_condattr_t *a, clockid_t clk)
 {
-	if (clk < 0 || clk-2U < 2) return EINVAL;
+	if (clk->id < 0 || clk->id-2U < 2) return EINVAL;
```
sbc100, line 7: I wonder if we can drop `cloudlibc/src/common/clock.h` completely and just use __wasi_clockid_t as the definition for clockid_t?.. avoiding the need for this change?

sbc100: Actually I guess the comments in `libc-bottom-half/cloudlibc/src/common/clock.h` are pretty clear this was a deliberate choice.   I guess this change is needed.

sbc100 (approved):

Makefile
loganek: ok, let's keep it as it is for now then. Thanks for the feedback.

libc-top-half/musl/src/thread/pthread_condattr_setclock.c
loganek: updated

libc-top-half/musl/src/thread/pthread_condattr_setclock.c
loganek: ok

abrown (approved):

PR #321, opened 2022-09-04
Implement support for pthread_rwlock

loganek:
For now only tested if the change builds (with `make -j15 THREAD_MODEL=posix`)


sbc100 (approved):

sbc100:
@abrown has been leading he charge on getting pthread support added so we should make sure this doesn't conflict with stuff he is already working on.

abrown (approved):

PR #323, opened 2022-09-05
threads: implement support for conditional variables

loganek:
The implementation is not as efficient as for native Linux platform due
to lack of FUTEX_REQUEUE-like system call in WASI.
    
For now we wake all the waiters which is inefficient; if that becomes
a bottleneck, I suggest we'll revisit the implementation.

The change is built on top of https://github.com/WebAssembly/wasi-libc/pull/320

### DIFF
--- a/libc-top-half/musl/src/thread/pthread_cond_timedwait.c
+++ b/libc-top-half/musl/src/thread/pthread_cond_timedwait.c
@@ -1,5 +1,9 @@
 #include "pthread_impl.h"
 
+#ifndef __wasilibc_unmodified_upstream
+#include <common/clock.h>
+#endif
+
 /*
  * struct waiter
  *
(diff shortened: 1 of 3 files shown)


libc-top-half/musl/src/thread/pthread_cond_timedwait.c
```diff
@@ -156,7 +173,7 @@ int __pthread_cond_timedwait(pthread_cond_t *restrict c, pthread_mutex_t *restri
 		if (val>0) a_cas(&m->_m_lock, val, val|0x80000000);
 		unlock_requeue(&node.prev->barrier, &m->_m_lock, m->_m_type & (8|128));
 	} else if (!(m->_m_type & 8)) {
-		a_dec(&m->_m_waiters);		
+		a_dec(&m->_m_waiters);
```
sbc100, line 176: revert these 3 whitespace-only changes?

loganek: I guess that's to keep the source as close to musl as possible? I've reverted that (and two similar) changes.

abrown:
Rebase now that #320 is merged?

loganek:
@abrown rebased

abrown (approved):

PR #324, opened 2022-09-05
threads: enable support for spinlock

loganek:


sbc100 (approved):

sunfishcode:
Could you say more about the context for this? WASI doesn't currently expose hooks for configuring the scheduler, so it seems like it'd be risking priority inversions and deadlocks. An alternative might be to implement the spin lock API with regular locks, however we should ideally look at how people expect to use these APIs.

loganek:
@sunfishcode I'm just trying to compile sample applications to run on WASM runtime and I don't have any specific use case for this for now. I'm a bit reluctant implementing a spinlock differently than spinlocks are normally implemented as this might be confusing and won't meet certain expectations that people have when use it.
Having said that, I do understand your concerns, and I don't mind waiting for a bit to gather more feedback on that.

sbc100:
I don't see how these are any different to spinlocks that one can already write using raw atomic instructions.  I don't see any reason to diverge from the upstream musl code here.  Spinlocks are already something that one should be using at ones own risk. 

The fact that we don't expose "hooks for configuring the scheduler" IMHO makes them, if anything, slightly safer to use that on other platforms that do expose such things since the likelyhood of threads within a wasm program having a different priority is therefore lower (than in a program that is able to access such hooks and meddle with priorities).

sunfishcode:
I myself don't know of any real-world code that people want to run on wasm that uses the pthread spin lock API. If anyone has any such code, I'm interested in learning about it, and what assumptions it might be relying on.

I don't have a major objection to adding this, I just want to take what appears to be an opportunity to ask this question.

yamt (approved):
honestly speaking, i think pthread spin lock is a misfeaure with no good applications.
however, as it's in the standard, it isn't surprising if some applications are using it.

loganek:
I still think we should enable it; spin lock has very clear definition and it's up to the user to use it or not.

abrown (approved):
I agree with @sunfishcode's point that spin locks in WebAssembly seem a bit weird, but I do think that, for compatibility sake, having this functionality available in wasi-libc makes sense. Let's merge it!

PR #325, opened 2022-09-13
threads: implement `pthread_create`

abrown:
As described in the [`wasi-threads`] proposal, this change implements `pthread_create` using the new `wasi_thread_spawn(void *arg)` API. As described there, `wasi-libc` exports the thread entry point with the expected name, `wasi_thread_start`, and then unwraps the passed argument `struct` to invoke the user function with the user argument `struct`.

[`wasi-threads`]: https://github.com/WebAssembly/wasi-threads/pull/5


libc-top-half/musl/src/thread/pthread_create.c
```diff
@@ -230,15 +277,26 @@ static void init_file_lock(FILE *f)
...
 }
 
+int32_t __imported_wasi_snapshot_preview1_thread_spawn(int32_t arg0) __attribute__((
+    __import_module__("wasi_snapshot_preview1"),
+    __import_name__("thread_spawn")
+));
+
+static int wasi_thread_spawn(void* start_args) {
+	return __imported_wasi_snapshot_preview1_thread_spawn((int32_t) start_args);
+}
```
abrown, line 287: I don't think this is the right way to import new WASI functions... what _is_ the right way?

libc-top-half/musl/src/thread/pthread_create.c
```diff
@@ -211,6 +240,24 @@ static int start_c11(void *p)
...
+__attribute__((export_name("wasi_thread_start")))
+int wasi_thread_start(int tid, void *p)
+{
+	struct start_args *args = p;
+	// Set the thread ID on the pthread structure and save this structure as
+	// the global `pthread_self`.
+	args->thread->tid = tid;
+	__asm__("local.set %0\n"
+		  "global.set __wasilibc_pthread_self\n"
+          : "=r"(args->thread));
```
abrown, line 253: Perhaps there is a better way to set this global?

libc-top-half/musl/src/thread/pthread_create.c
```diff
@@ -364,6 +450,26 @@ int __pthread_create(pthread_t *restrict res, const pthread_attr_t *restrict att
...
+		if (wait_ret == 1) {
+			// Loaded value != expected value; the pthread was not set up
+			// correctly.
+			return -EINVAL;
+		}
+		if (wait_ret == 2) {
+			// The timeout expired.
+			return -ETIMEDOUT;
+		}
+	}
```
abrown, line 471: We can get rid of this `wait` (and the corresponding `notify` in `wasi_thread_start`) if we return the TID from `wasi_thread_spawn`.

libc-top-half/musl/src/thread/pthread_create.c
```diff
@@ -364,6 +450,26 @@ int __pthread_create(pthread_t *restrict res, const pthread_attr_t *restrict att
...
 			__wait(&args->control, 0, 3, 0);
 	}
+#else
+	/* As in the unmodified version, all spawn failures translate to EAGAIN.
+	 * If the spawn did succeed, then wait for the TID to be populated by the
+	 * child thread. */
+	if (ret != 0) {
+		ret = -EAGAIN;
+	} else {
+		int timeout_ns = -1; // TODO
```
abrown, line 460: How long to wait here for the child thread to report it's TID?

libc-top-half/musl/src/thread/pthread_create.c
sbc100: Currently the only way to set a real wasm global is via assembly (either .s file or inline like this).

libc-top-half/musl/src/thread/pthread_create.c
sbc100: I don't see why we would just wait forever.   If the thread doesn't start that would be bug in the hos wouldn't it?  I don't think need to worry such things at runtime.

libc-top-half/musl/src/thread/pthread_create.c
sbc100: I think we normally expect the low level import to be declared in `libc-bottom-half/headers/public/wasi/api.h`.    Although I'm not sure we still have the capability to regenerate that from the witx anymore.

Perhaps it should be in a different module name rather than adding to the old `wasi_snapshot_preview1`?

libc-top-half/musl/src/thread/pthread_create.c
sbc100: Presumably we can also get rid of the wait by using a purely userspace/fake (e.g. monotonically increasing atomic) tid?

libc-top-half/musl/src/thread/pthread_create.c
abrown: You mean "I don't see why we would _not_ just wait forever"?

sbc100: Yes sorry!

abrown: Yeah, I guess I'll just document the "wait forever... bug in the host" bit and leave it at `-1`.

libc-top-half/musl/src/thread/pthread_create.c
abrown: Yes, that would be another way to do this. In the discussion we had, others considered it necessary for the host to define this TID since it may need to maintain of mapping of that to real TIDs.

libc-top-half/musl/src/thread/pthread_create.c
abrown: After talking to @sunfishcode I manually moved them to the bottom half behind `#ifdef _REENTRANT`. I imported from `wasi_snapshot_preview2` because I didn't know where else to import from (?).

libc-top-half/musl/src/thread/pthread_create.c
abrown: Actually, thinking about this more, I now think a monotically increasing atomic TID controlled by the Wasm side (not the host) will still require a `wait`/`notify`: there are multiple threads that need the TID to continue &mdash the parent thread and the child thread. If the TID is incremented in the parent, then the child must be notified that the TID has been assigned before it continues to run the user function so that the TID is available to that code. If the TID is incremented in the child, then the parent must be notified of the new TID so that it can return with a correctly filled out `struct pthread`. Since both threads need the TID to continue, we would still need the `wait`/`notify`.

sbc100: To repeat what was just said out meeting, I think if you have a completely user-space TID it can be full created/decided prior the thread_spawn syscall (e.g. it can just be part of the pthread_t struct that is passed to the child).

loganek: What ID will the main thread get? Is it going to be `0`, and the new TIDs start from `1`? Also, given that the host won't have access to those TIDs, what's the plan to implement (in the future) functions that manipulates native threads (e.g. setting priority)? Is the TID created in the userspace being passed to the `wasi_thread_spawn` so the runtime can keep track on the mapping?

sbc100: We need to decide if we want to bake in the idea that that host and the module should share a concept of TID from day 1.   With out initial goal of just creating `thread_spawn` there is no need to such a TID concept.

The need for way to pass a shared TID to the host will only exist once we add API such as `set_priority` or `thread_kill`.. which we don't have immediate plans for but can see the possible need for in the future.

I guess it comes down of two options:

1) add a shared TID concept from day today by adding a (currently useless) argument to `thread_spawn` or to the thread entry point function.
2) Add a new argument to one of those functions at the point when we add those new APIs. 
3) Add a new API that is something like `get_current_thread_id()` at the point when we add those new functions.

I think (3) has some downsides, but if we go with (1) then we can delay this discussion and focus just one what we need today for MVP.

loganek: I *guess* sooner or later we'll need a host to know about the thread ID. Option 1. sounds reasonable; I'd rather stay away from 2. as I think we'd have to break the API to make that change.

libc-top-half/musl/src/thread/pthread_create.c
```diff
@@ -211,6 +240,24 @@ static int start_c11(void *p)
 	__pthread_exit((void *)(uintptr_t)start(args->start_arg));
 	return 0;
 }
+#else
+__attribute__((export_name("wasi_thread_start")))
+int wasi_thread_start(int tid, void *p)
```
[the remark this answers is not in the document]
sbc100: Because if we pass out a function pointer, then in order to have it be useable/callable, we also need to export the entire table from the wasm module.   We discussed the the cost/benefit of doing this (sorry I can't remember exactly where).    

My personal feeling is that exporting the entire table just for this purpose is not worth it.   In addition of over-sharing, there is also at least one binaryen optimization that is prevented when the table is exported (The important detail is that a non-exported table can be considered read-only when there are no table.set instructions and binaryen can do it calls "directize"ing in this case.. i.e. it can turn some indirect calls into direct calls).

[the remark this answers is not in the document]
sbc100: I guess once we have first class function references then maybe we could switch to using them instead of a fixed export?  (i.e. we could do table.get and then pass the result out of the host).

libc-top-half/musl/src/thread/pthread_create.c
loganek: Yes. My comment was related to the case where the thread ID is generated in the userspace - in that case, we probably want to add another parameter to `wasi_thread_spawn` to tell the runtime the calculated thread id (so it can maintain the mapping).

libc-top-half/musl/src/thread/pthread_create.c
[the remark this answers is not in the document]
abrown: If we pass a function pointer, then the Wasm module must export its function `table` under a fixed name so the host can access the table the function pointer lives in. What's the difference between exporting a function and a table? I can't recall all the details from our last discussion, but IIRC it was trickier to export the function table (toolchain-wise) and, in my opinion, just setting a function name is a bit more direct and clear.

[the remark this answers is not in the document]
abrown: (@sbc100, seems like we raced!)

libc-top-half/musl/src/thread/pthread_create.c
sbc100: Yes I kind of like that option... since it means that user space never sees anything but its own TID which it generated.   

Then the host doesn't have to generate or share TIDs of its own and there is no possible figerprinting that can be done based on the host-provided TIDs.

expected/wasm32-wasi/posix/defined-symbols.txt
```diff
@@ -38,9 +39,11 @@ __ctype_toupper_loc
 __cxa_atexit
 __cxa_finalize
 __des_setkey
+__dl_thread_cleanup
```
sunfishcode, line 42: I'd prefer to avoid defining this symbol, since it's in support of `dlopen`.

libc-top-half/musl/src/thread/pthread_create.c
```diff
@@ -251,9 +300,13 @@ int __pthread_create(pthread_t *restrict res, const pthread_attr_t *restrict att
 		init_file_lock(__stdin_used);
 		init_file_lock(__stdout_used);
 		init_file_lock(__stderr_used);
+#ifdef __wasilibc_unmodified_upstream
 		__syscall(SYS_rt_sigprocmask, SIG_UNBLOCK, SIGPT_SET, 0, _NSIG/8);
+#endif
 		self->tsd = (void **)__pthread_tsd_main;
+#ifdef __wasilibc_unmodified_upstream
 		__membarrier_init();
+#endif
```
sunfishcode, line 309: __membarrier_init is commented out here, but I see above that the function is still defined, presumably to the "dummy" version since the actual version does a Linux syscall. For tidiness, it'd be good to avoid defining this symbol too.

libc-top-half/musl/src/thread/pthread_create.c
sunfishcode: @sbc100 Can you say more about your reasons for wanting userspace to never see anything but its own TID? What kind of fingerprinting are you concerned about here?

sbc100: I was thinking that any information leaks from the host are a potential boon to the attacker, and TIDs could be of particular interest.

For example, I imagine that observing a particular pattern of TIDs could leak the identity of the host OS.  It doesn't seem like a major risk, just one that might be good to avoid if there are is an easy way to do so.

Making the TID (a) optional, and (b) created in user space (at the point of thread creation) , seems like a easy design choice that avoids host-side TIDs from being passed to the client module.  

What if we made the TID a (optional) second parameter to the thread creation function?:

```
i32: Thread ID that can be used in later system calls to identify the newly created thread.  0 means the thread cannot be referred to by such system calls.  Thread creation will fail if ID is already used by currently running thread.
```

sunfishcode: My thinking was that hosts where TIDs are sensitive can easily use a hash table or so to hide the host TIDs from wasm. Wasm-generated TIDs would ultimately need the exact same kind of hash table, to map wasm-generated TIDs to host TIDs, so it wouldn't be significantly different. And it would avoid having two different kinds of threads, with-TID and without-TID, which could potentially be an awkward special case in some settings.

abrown: For later reference: the main reason for host-generated TIDs is that, in a scenario where multiple modules are linked together from different thread paradigms, it would be difficult to coordinate a central "atomically-incremented TID counter." E.g., if a Rust module and a C module both had Wasm-generated TIDs and were linked togethers, they would duplicate each other without some tricky toolchain coordination.

sbc100 (approved):
Great!

loganek (approved):

PR #328, opened 2022-09-28
Don't run static constructors on arbitrary user exports.

sunfishcode:
Previously, "new-style commmands" considered every user-defined export to be a potential command entrypoint, so wasi-libc and wasm-ld cooperated to run the user's static constructors on each entrypoint.

This form of new-style command turned out not to be useful, and it interferes with some use cases, so disable it.

This is done by making an explicit call to `__wasm_call_ctors`, which tells wasm-ld that it shouldn't synthesize any calls to `__wasm_call_ctors` on its own.


sbc100 (approved):

yamt (approved):
lgtm, thank you.

PR #329, opened 2022-09-30
Add a check to command modules to ensure that they're only started once.

sunfishcode:
Wasm command modules should only be called once per instance, because the programming model doesn't leave linear memory in a reusable state when the program exits. As use cases arise for loading wasm modules in environments that want to treat them like reactors, add a safety check to ensure that command modules are used according to their expectations.

### DIFF
--- a/libc-bottom-half/crt/crt1-command.c
+++ b/libc-bottom-half/crt/crt1-command.c
@@ -3,8 +3,18 @@ extern void __wasm_call_ctors(void);
 extern int __main_void(void);
 extern void __wasm_call_dtors(void);
 
+// Commands should only be called once per instance. This simple check ensures
+// that the `_start` function isn't started more than once.
+static volatile int started = 0;
+
 __attribute__((export_name("_start")))
 void _start(void) {
+    // Don't allow the program to be called multiple times.
+    if (started != 0) {
+	__builtin_trap();
+    }
+    started = 0;
+
     // The linker synthesizes this to call constructors.
     __wasm_call_ctors();
 
--- a/libc-bottom-half/crt/crt1-command.c
+++ b/libc-bottom-half/crt/crt1-command.c
@@ -13,7 +13,7 @@ void _start(void) {
     if (started != 0) {
 	__builtin_trap();
     }
-    started = 0;
+    started = 1;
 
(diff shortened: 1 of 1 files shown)


libc-bottom-half/crt/crt1-command.c
```diff
@@ -3,8 +3,20 @@ extern void __wasm_call_ctors(void);
 extern int __main_void(void);
 extern void __wasm_call_dtors(void);
 
+// Commands should only be called once per instance. This simple check ensures
+// that the `_start` function isn't started more than once. `address_space(1)`
+// tells clang to put this variable in a wasm global.
```
sbc100, line 8: Does that actually work?  It seems a little risky to depend on this now.  Why not just is a simple static?

libc-bottom-half/crt/crt1-command.c
```diff
@@ -3,8 +3,20 @@ extern void __wasm_call_ctors(void);
...
+__attribute__((address_space(1)))
+static volatile int started = 0;
+
 __attribute__((export_name("_start")))
 void _start(void) {
+    // Don't allow the program to be called multiple times.
+    if (started != 0) {
+	__builtin_trap();
+    }
+    started = 0;
```
sbc100, line 18: Seems like it would be good place for `#ifndef NDEBUG` but we don't have a debug version of libc yet IIUC

libc-bottom-half/crt/crt1-command.c
sunfishcode: No big reason; and depending on it breaks LLVM 11, which we still support for now, so I'll switch to a static.

libc-bottom-half/crt/crt1-command.c
sunfishcode: I care about code size, but this is just a few bytes. And this hazard comes up in both component use cases and JS embedding use cases, where creating an instance is separated from running the command and there's no other way to prevent the command from being run multiple times, causing UB.

sbc100: `started = 1;` here

sbc100 (approved):
lgtm with comment

sunfishcode:
Oops; fixed.

PR #330, opened 2022-09-30
[WIP] Add CMake build system

### LINKED ISSUE #322: Migrate build system to CMake
I understand that a great thing about Makefile (versus CMake) is less dependencies. However, I see it comes with its own limitations, like not great Windows support, lack of ninja build (if that's preferred) or a bit obscure syntax (although that's very subjective). I've seen a PR https://github.com/WebAssembly/wasi-libc/pull/154/ although that seemed abandoned. Not sure if there was any reason for that or was it simply lost of interest? I'd be happy to pick up the work and finish it unless there was a reason to keep Makefiles.

loganek:
This is a work-in-progress PR and still require a few actions. I open the PR to get early feedback. Some of the changes are clone from the other attempt of migrating this repository to CMake build system: #154. One major difference between this change and the previous one is that I'm using install target to create sysroot; I find this to be a bit more natural than specifying destination for every target.

I tested the changes by diff-ing generated install directory with the sysroot generated by Makefile. I also compiled and ran a few apps on WAMR runtime.

- [ ] remove Makefile
- [ ] update CI files

Resolves #322


sbc100:
I know you mention some rational in #322, but I'm curious what you personal primary motivation is for wanting to make this change?  I guess better native windows support is the big one?   Without needing to setup something like mingw?

I personally find cmake syntax and semantics pretty confusing (more so than the relative simplicity of a Makefile), but I get that its personal preference.

This does seem to make some things more complex.  For example, it looks like it takes roughly twice as many lines of code to implement.   We would also be adding a python dependency.    I'm generally a fan of python over shell scripts for this kind of thing but others might feel differently.

cmake/libc-bottom-half/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
...
+
+# FIXME(https://reviews.llvm.org/D85567) - due to a bug in LLD the weak
+# references to a function defined in `chdir.c` only work if `chdir.c` is at the
+# end of the archive, but once that LLD review lands and propagates into LLVM
+# then we don't have to do this.
+set(CHDIR_SRC "${LIBC_BOTTOM_HALF_SRC_DIR}/chdir.c")
+list(REMOVE_ITEM LIBC_BOTTOM_HALF_ALL_SOURCES ${CHDIR_SRC})
+list(APPEND LIBC_BOTTOM_HALF_ALL_SOURCES ${CHDIR_SRC})
+
+add_library(libc-bottom-half OBJECT
```
sbc100, line 27: I guess this is some kind of "library" that getting somehow inlined into the main libc?   More like a list of object files that a library per-say?

I assume all the object files that make up libc-bottom-half will still show up as individual object files within libc.a?

sbc100 (commented):
Could you add some instructions (maybe as a mod to the README) so I can try it out?  I've love to give it go.  Does this setup support out-of-tree building?

abrown:
> I personally find cmake syntax and semantics pretty confusing

I'm also not sold on replacing `make` with `cmake`. `make` is available on Windows in various ways: WSL, MinGW, and Visual Studio. I'm not opposed to replacing it, but I feel `cmake` would have to offer something better than what currently exists. Also, I'm a bit concerned that people working on the project may be familiar with a `Makefile` but would be slower to hack on a `CMakeLists.txt` (as @sunfishcode mentions in #322). Just some thoughts...

penzn:
In my opinion, it provides more watertight ways to ensure you have the pieces you need, also it has "first class types" for sub-targets like libraries, including dependence detection, this reduces risk of mistyping or having parts of the build go out of sync. It isn't a silver bullet, though.

loganek:
> I know you mention some rational in https://github.com/WebAssembly/wasi-libc/issues/322, but I'm curious what you personal primary motivation is for wanting to make this change? 

Pretty much what @penzn already mentioned but also having experience with both Makefile and CMake, I find both learning CMake and maintaining CMake project easier. I also think Makefile has a steeper learning curve than CMake but that may be because I ~~learned~~ used Makefile first.

> This does seem to make some things more complex. For example, it looks like it takes roughly twice as many lines of code to implement.

I wouldn't use number of lines as a metric for complexity. If needed, I could probably optimize (most likely at a cost of readability, although I didn't look into it).

> We would also be adding a python dependency. I'm generally a fan of python over shell scripts for this kind of thing but others might feel differently.

I don't think this is relevant to Makefile vs CMake discussion. They both can run shell and python scripts; I copied it (and slightly modified) from previous PR because I found it more readable.

> Also, I'm a bit concerned that people working on the project may be familiar with a `Makefile` but would be slower to hack on a `CMakeLists.txt` (as @sunfishcode mentions in #322). Just some thoughts...

What I found is that creating a CMake project from scratch requires some knowledge, but most of the modifications don't often require looking into documentation, as you can do most of what you need based on already written code (I guess just like Makefile)

loganek:
> Could you add some instructions (maybe as a mod to the README) so I can try it out? I've love to give it go. Does this setup support out-of-tree building?

I can definitely update the README file; meanwhile, here's a command you can use:
```
# assume you're in wasi-libc
cd .. # go up; we'll do out-of-tree build
mkdir wasi-libc-build && cd wasi-libc-build
# configure the project with ninja (-G). The argument can be omitted and it
# will use default generator (On linux, most likely it will be Unix Makefiles)
cmake -DCMAKE_C_COMPILER=/usr/bin/clang-15 -DCMAKE_BUILD_TYPE=Release -DTHREAD_MODEL=posix -DCMAKE_INSTALL_PREFIX=../cmake-install -G Ninja ../wasi-libc 
ninja install 
```

sunfishcode:
One concrete advantage of moving to CMake is that the current wasi-libc Makefile doesn't do dependency management correctly, so we effectively don't have incremental builds right now.

CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+cmake_minimum_required(VERSION 3.16.3)
+
+project(wasi-libc LANGUAGES C)
+
+enable_testing()
+
+if("${CMAKE_C_COMPILER}" MATCHES "(.*)clang(-[0-9]+)?$")
+  set(CMAKE_NM "${CMAKE_MATCH_1}llvm-nm${CMAKE_MATCH_2}")
```
compnerd, line 8: I would recommend against this.  This prevents the user from specifying the name tool to use.

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+  wasi-emulated-signal-musl-objs
+  wasi-emulated-signal-objs
+)
+add_custom_command(TARGET wasi-emulated-signal POST_BUILD
+  COMMAND ${CMAKE_C_COMPILER_AR} crs $<TARGET_FILE:c>
+)
+
+add_library(c)
+
+set (LIBC_LINK_LIBRARIES
```
compnerd, line 33: This is an anti-pattern.  Prefer to directly use the targets with `target_link_libraries`

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+  libc-bottom-half
+)
+
+if(${BUILD_LIBC_TOP_HALF} STREQUAL "ON")
```
compnerd, line 37: I tend to prefer the more concise spelling:

```suggestion
if(BUILD_LIBC_TOP_HALF)
```

This has the benefit of allowing the user to specify the value with a truthy value that CMake permits (ON, YES, 1).

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+    list(APPEND LIBC_LINK_LIBRARIES
+      libc-top-half
+    )
```
compnerd, line 40: Again, an anti-pattern.  Prefer to use a generator expression with `target_link_libraries`.

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+endif()
+
+if(${MALLOC_IMPL} STREQUAL "dlmalloc")
+  list(APPEND LIBC_LINK_LIBRARIES
+    dlmalloc)
```
compnerd, line 45: Similar

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+endif()
+
+target_link_libraries(c
+  PRIVATE
+    ${LIBC_LINK_LIBRARIES}
+)
+
+add_custom_command(TARGET c POST_BUILD
+  COMMAND ${CMAKE_C_COMPILER_AR} crs $<TARGET_FILE:c>
+)
```
compnerd, line 59: Why do this?  Add a proper target for a static archive.  There is no guarantee the librarian accepts the GNU style arguments.

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+
+add_dependencies(c sysroot-headers)
+
+# Create empty placeholder libraries.
+foreach(stub m rt pthread crypt util xnet resolve dl)
+  set(lib_path ${SYSROOT_LIB_DIR}/lib${stub}.a)
+  add_custom_command(OUTPUT ${lib_path}
+    COMMAND ${CMAKE_COMMAND} -E make_directory ${SYSROOT_LIB_DIR}
+    COMMAND ${CMAKE_C_COMPILER_AR} crs ${lib_path})
```
compnerd, line 68: Prefer to use `add_library` over a custom command.

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+    wasi-emulated-process-clocks
+    wasi-emulated-getpid
+    wasi-emulated-signal
+  ARCHIVE DESTINATION ${INSTALL_LIB_DIR}
+  LIBRARY DESTINATION ${INSTALL_LIB_DIR}
+)
+
+install(
+  DIRECTORY ${SYSROOT_INC_DIR}
+  DESTINATION .
```
compnerd, line 88: This seems odd, why is the destination `.`?

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+)
+
+install(
+  DIRECTORY ${SYSROOT_SHARE_DIR}
+  DESTINATION share
+)
+
+install(
+  DIRECTORY ${SYSROOT_LIB_DIR}
```
compnerd, line 97: Why the custom variables rather than CMake's install variables (use `include(GNUInstallDirs)`).

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+  DESTINATION lib
+)
+
+add_test(NAME check-metadata
+  COMMAND diff -wur
```
compnerd, line 102: `diff` is not guaranteed to be present in the path on Windows.

cmake/CMakeLists.txt
```diff
@@ -0,0 +1,112 @@
...
+    ${PROJECT_SOURCE_DIR}/expected/${MULTIARCH_TRIPLE}/${THREAD_MODEL}/predefined-macros.txt
+    ${CMAKE_BINARY_DIR}/sysroot/share/${MULTIARCH_TRIPLE}/predefined-macros.txt)
+
+add_test(NAME check-headers
+  COMMAND ${CMAKE_C_COMPILER}
+    -target ${CMAKE_C_COMPILER_TARGET}
```
compnerd, line 108: This prevents the use of any compiler other than clang.

cmake/dlmalloc/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+set(DLMALLOC_DIR ${PROJECT_SOURCE_DIR}/dlmalloc)
```
compnerd, line 1: Why the variable for a single use?

cmake/global-compiler-options.cmake
```diff
@@ -0,0 +1,39 @@
+# WebAssembly floating-point match doesn't trap.
+# TODO: Add -fno-signaling-nans when the compiler supports it.
+add_compile_options(-fno-trapping-math)
```
compnerd, line 3: You should validate that the compiler supports the option.

cmake/global-compiler-options.cmake
```diff
@@ -0,0 +1,39 @@
...
+  -Wno-unused-parameter
+  -Wno-sign-compare
+  -Wno-unused-variable
+  -Wno-unused-function
+  -Wno-ignored-attributes
+  -Wno-missing-braces
+  -Wno-ignored-pragmas
+  -Wno-unused-but-set-variable
+  -Wno-unknown-warning-option
+)
```
compnerd, line 18: Similar

cmake/libc-bottom-half/CMakeLists.txt
```diff
@@ -0,0 +1,41 @@
+set(LIBC_BOTTOM_HALF_CLOUDLIBC_SRC_INC_DIR ${LIBC_BOTTOM_HALF_CLOUDLIBC_SRC_DIR}/include)
+set(LIBC_BOTTOM_HALF_SRC_DIR ${LIBC_BOTTOM_HALF_DIR}/sources)
+set(LIBC_BOTTOM_HALF_HEADERS_PRIVATE_DIR ${LIBC_BOTTOM_HALF_DIR}/headers/private)
+
+file(GLOB_RECURSE
+  LIBC_BOTTOM_HALF_ALL_SOURCES
+  CONFIGURE_DEPENDS
+  ${LIBC_BOTTOM_HALF_CLOUDLIBC_SRC_DIR}/*.c
+  ${LIBC_BOTTOM_HALF_SRC_DIR}/*.c)
```
compnerd, line 9: This is a highly dubious.  This will lose the ability to alter the sources.  The glob is applied once at configure time and *never* again.  You should explicitly list the sources.   Similar throughout.

compnerd (commented):
Part of why I had lost interest was the difficulty to get the proper support for CMake added due to upstream resistance.

PR #331, opened 2022-10-05
Add a separate install target for threaded libc

penzn:
Use `MULTIARCH_TRIPLE` to append 'pthread' and `TARGET_TRIPLE` as more generic `wasm32-wasi`.

To use this sysroot, one would have to compile with `--triple=wasm32-wasi-pthread`.


Makefile
```diff
@@ -37,6 +37,12 @@ BULK_MEMORY_THRESHOLD ?= 32
 TARGET_TRIPLE = wasm32-wasi
 MULTIARCH_TRIPLE = wasm32-wasi
 
+# Threaded version necessitates a different traget, as objects from different
+# targets can't be mixed together while linking
+ifeq ($(THREAD_MODEL), posix)
+MULTIARCH_TRIPLE = wasm32-wasi-pthread
```
sbc100, line 43: I wonder if its work being explicit here and calling this `wasm32-unknown-wasi-pthread` ?

sbc100: Also, is there a reason with have TARGET_TRIPLE and MULTIARCH_TRIPLE separated here?  I would that thought that we want to just use a single triple everywhere?

sunfishcode: I propose we handle the vendor string in the same way as the non-pthread builds, so that the only difference is the suffix.

Makefile
```diff
@@ -658,7 +664,7 @@ check-symbols: startup_files libc
 
 	# Check that the computed metadata matches the expected metadata.
 	# This ignores whitespace because on Windows the output has CRLF line endings.
-	diff -wur "$(CURDIR)/expected/$(MULTIARCH_TRIPLE)/$(THREAD_MODEL)" "$(SYSROOT_SHARE)"
+	diff -wur "$(CURDIR)/expected/$(TARGET_TRIPLE)/$(THREAD_MODEL)" "$(SYSROOT_SHARE)"
```
sbc100, line 667: Oh I see.   Now that we have the thread model as part of the triple we don't need the extra directory level here.. we can just remove `/$(THREAD_MODEL)`

Makefile
penzn: > Also, is there a reason with have TARGET_TRIPLE and MULTIARCH_TRIPLE separated here? I would that thought that we want to just use a single triple everywhere?

Should we update both to include pthreads or should we only keep one variable? Both are viable solutions I think.

Makefile
penzn: Good point, I can change that.

Makefile
penzn: Maybe I misread your question initially. I don't know why there are two variables, but can try to make it into just one.

sunfishcode: I don't know the precise state of things here, but my suggestion was just to avoid making the `unknown` vendor field explicit here, since it's implicit everywhere else and making it explicit here would make more things different.

penzn: Oh, I meant @sbc100's comment right before yours - it looks like he was asking why there are two variables.

penzn:
I've added a commit to that would leave only one triple variable in the Makefile. I am not squashing it yet, so that it is easier to drop it if that is not what we need.

sbc100 (commented):
LGTM!

Have you tested this?  Does clang/llvm already support `wasm32-wasi-pthread`?

penzn:
> Have you tested this? Does clang/llvm already support `wasm32-wasi-pthread`?

I just realized I didn't build `compiler-rt` with it, working through that, but Clang would happily accept `wasm32-wasi-pthread` today.

penzn:
@sbc100 how do you normally test something like that? I have built `wasi-sdk` pointing to this branch with `THREAD_MODEL=posix` on, I can get through full build of a simple program and it gets atomics in the output.

sbc100:
> @sbc100 how do you normally test something like that? I have built `wasi-sdk` pointing to this branch with `THREAD_MODEL=posix` on, I can get through full build of a simple program and it gets atomics in the output.

Until very recently there was no testing at all in this repo and testing via integration with `wasi-sdk` was the only way to go.   There was recently some minimal testing added to the `test/` directory.  I think those get run during CI so if its green you didn't break them.

If you have have manually verified via integration with wasi-sdk and the CI here is green I think thats good enough.

sbc100 (approved):

penzn:
I don't have commit privileges here, can you merge this for me?

PR #332, opened 2022-10-06
Add braces to `if` statements whose bodies have multiple statements.

sunfishcode:
Add braces to `if` statements so that the whole intended body is covered by the conditional.


sbc100:
I thought clang had a warning that for these days?

sunfishcode:
It appears the warning is disabled when there's a preprocessor macro involved, such as `UNLOCK` in this case.

PR #336, opened 2022-10-10
threads: enable pthread_join

loganek:


sbc100 (approved):

abrown (approved):
LGTM; I have some doubts about whether `__pthread_setcancelstate` here even needs to be (or should be) called but I think it is fine for now. And this `free` looks right; the `pthread_create` is doing the `malloc`.

PR #339, opened 2022-10-14
Support threads in the new crt1-command.c ctor check.

sunfishcode:
Use an atomic compare-and-swap for checking whether constructors have been run, when threads are enabled.

### DIFF
--- a/libc-bottom-half/crt/crt1-command.c
+++ b/libc-bottom-half/crt/crt1-command.c
@@ -8,7 +8,7 @@ void _start(void) {
     // Commands should only be called once per instance. This simple check
     // ensures that the `_start` function isn't started more than once.
 #ifdef _REENTRANT
-    static volatile _Atomic int started = 0;
+    static volatile int started = 0;
     if (__sync_val_compare_and_swap(&started, 0, 1) != 0) {
 	__builtin_trap();
     }
--- a/libc-bottom-half/crt/crt1-command.c
+++ b/libc-bottom-half/crt/crt1-command.c
@@ -3,17 +3,22 @@ extern void __wasm_call_ctors(void);
 extern int __main_void(void);
 extern void __wasm_call_dtors(void);
 
-// Commands should only be called once per instance. This simple check ensures
-// that the `_start` function isn't started more than once.
-static volatile int started = 0;
-
 __attribute__((export_name("_start")))
 void _start(void) {
-    // Don't allow the program to be called multiple times.
+    // Commands should only be called once per instance. This simple check
+    // ensures that the `_start` function isn't started more than once.
+#ifdef _REENTRANT
+    static volatile _Atomic int started = 0;
+    if (__sync_val_compare_and_swap(&started, 0, 1) != 0) {
+	__builtin_trap();
+    }
+#else
+    static volatile int started = 0;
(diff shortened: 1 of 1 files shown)


abrown (approved):
This makes sense to me.

abrown:
cc: @haraldh, this might be helpful for you.

libc-bottom-half/crt/crt1-command.c
```diff
@@ -3,17 +3,22 @@ extern void __wasm_call_ctors(void);
...
-    // Don't allow the program to be called multiple times.
+    // Commands should only be called once per instance. This simple check
+    // ensures that the `_start` function isn't started more than once.
+#ifdef _REENTRANT
+    static volatile int started = 0;
+    if (__sync_val_compare_and_swap(&started, 0, 1) != 0) {
+	__builtin_trap();
+    }
+#else
+    static volatile int started = 0;
```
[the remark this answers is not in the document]
sunfishcode: No reason; I've now factored it out.

libc-bottom-half/crt/crt1-command.c
```diff
@@ -3,17 +3,21 @@ extern void __wasm_call_ctors(void);
...
-static volatile int started = 0;
-
 __attribute__((export_name("_start")))
 void _start(void) {
-    // Don't allow the program to be called multiple times.
+    // Commands should only be called once per instance. This simple check
+    // ensures that the `_start` function isn't started more than once.
+    static volatile int started = 0;
+#ifdef _REENTRANT
+    if (__sync_val_compare_and_swap(&started, 0, 1) != 0) {
```
sbc100, line 12: What about using C11 atomics here?  (https://www.ibm.com/docs/en/zos/2.1.0?topic=c11-atomic-compare-exchange-strong).  Maybe we don't want to depend on C11?

libc-bottom-half/crt/crt1-command.c
```diff
@@ -7,12 +10,14 @@ __attribute__((export_name("_start")))
...
-    static volatile int started = 0;
 #ifdef _REENTRANT
-    if (__sync_val_compare_and_swap(&started, 0, 1) != 0) {
+    static volatile _Atomic int started = 0;
+    int expected = 0;
+    if (!atomic_compare_exchange_strong(&started, &expected, 1)) {
 	__builtin_trap();
     }
 #else
+    static volatile int started = 0;
```
sbc100, line 20: No need for volatile here I think?

sbc100 (approved):

libc-bottom-half/crt/crt1-command.c
sunfishcode: I've now added a comment about `volatile`. It's admittedly a very theoretical concern. But also, this variable is read and written to exactly once, so there's nothing to optimize about it. So I'm inclined to keep the `volatile`.

sunfishcode: Persisting an in-memory variable past the point of an `_exit` call isn't something that's even possible to think about in C, and yet that's what this code is doing, so `volatile` helps make sure that the compiler just does what we're telling it to do, because it won't understand what we're actually doing.

sbc100 (approved):

Issue #319, opened 2022-08-29
dlmalloc has license incompatible with Fedora policies


khardix:
Hi all,
this is an inquiry more that an issue: Is there any way to use this libc without dlmalloc?

Recently, [Fedora removed approval for CC0-licensed code](https://lists.fedoraproject.org/archives/list/legal@lists.fedoraproject.org/thread/RRYM3CLYJYW64VSQIXY6IF3TCDZGS6LM/), generally due to possible patent trouble with such code. I do not really think that will ever happen here (although better safe than sorry); however, I would still like to comply with the policy when packaging.

I was so far unsuccessful in trying to build wasi-libc with any of the included muls malloc implementations, and I'm not sure that's possible without having `mmap`.

What can be done? Is there any other malloc implementation that could be used? Can I use the mmap-emulation code and some of the musl malloc(s) on top of it, or is that a bad idea?

Any help greatly appreciated.

TerrorJack:
`mimalloc` also claims to work with `wasm32`. I suppose it's possible to vendor `mimalloc` in this tree and have some degree of testing, and there's preliminary work on allowing building without `dlmalloc`.

sunfishcode:
The mmap emulation code works by calling malloc, so it unfortunately won't work for implementing a malloc.

As @TerrorJack mentions, there is a build parameter for changing the malloc implementation, though the current options are `dlmalloc` or `none`.
 
`mimalloc` is a possibility. We could certainly add that to the wasi-libc tree as an option, if someone wanted to port it. It'd also be something we could consider making the default allocator, though we'd want code size and performance comparisions. Also, I am concerned about the fact that mimalloc uses spin-wait loops for some things, though I haven't studied it in depth.

Another option would be to ask Fedora for an exception for dlmalloc. It's used in a *lot* of places.

kripken:
Another option is [emmalloc](https://github.com/emscripten-core/emscripten/blob/6153cb3a2a9c6d8446a46d9c300528c402f17b82/system/lib/emmalloc.c). Like dlmalloc it uses `sbrk`, and it's used in wasm today in emscripten as a drop-in dlmalloc replacement. It should just work in wasi-libc (as long as you don't `#define` the emscripten verbose logging code). License is MIT (compatible with wasi-libc).

[...]

PR #340, opened 2022-10-14
Port Emscripten's emmalloc to wasi-libc

sunfishcode:
As suggested by @kripken in https://github.com/WebAssembly/wasi-libc/issues/319#issuecomment-1230558688, port emmalloc to wasi-libc. At a quick experiment, it's 1781 bytes smaller than the current default dlmalloc. It's not enabled by default; to enable it, build wasi-libc with `MALLOC_IMPL=emmalloc`. I've only done very light testing so far.

I made a few modifications; the patch is split into two commits, with the first commit checking in the unmodified code, and the second commit showing the changes. The two most significant changes are:
 - Port the functionality in wasi-libc's dlmalloc for allocating memory from `__heap_base` before using memory.grow
 - Eliminate the use of a constructor, which is relevant to https://github.com/WebAssembly/WASI/pull/497.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -7,6 +7,7 @@ Portions of this software are derived from third-party works covered by
 their own licenses:
 
 dlmalloc/                     - CC0; see the notice in malloc.c for details
+emmalloc/                     - MIT; see the notice in emmalloc.c for details
 libc-bottom-half/cloudlibc/   - BSD-2-Clause; see the LICENSE file for details
 libc-top-half/musl/           - MIT; see the COPYRIGHT file for details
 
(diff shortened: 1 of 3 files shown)


sbc100 (commented):
lgtm, although I a little confused about the switch away form using a ctor.

Aren't we currently looking at converting dlmalloc to use a ctor in : https://github.com/WebAssembly/wasi-libc/pull/338?   Have we decided that isn't a good direction?

sunfishcode:
My thought for removing the constructor was just that avoiding static constructors is one way to avoid the cost of the wrappers in that LLVM patch.

From our discussion offline, another potential concern is future use cases that want to do allocation before calling normal user functions.

khardix:
I've just noticed this PR coming from #319, and I'm very interested in having this merged :) I'm not sure from the comments above if anything is actually blocking this being merged – is there still an issue to be solved? Can I help in any way?

sunfishcode:
I don't think anything is blocking this; we even have an lgtm comment :-). I've switched this out of draft and will land it soon if there are no objections.

sbc100 (approved):

PR #341, opened 2022-10-16
Improve build scripts + update Actions workflow

theduke:
- Updated sysroot so that it supports wasm64-wasix
- Fixed an issue where return types were wrong on the file operations - put them back to the originals
- First go at wasm64-wasi
- Updated the target from wasix to wasi so all the libraries coded against wasi still work
- Fixed the expected check
- Removed submodule
- Updated the library so it can compile both wasm32-wasi and wasm64-wasi
- Updated the library so it can compile both wasm32-wasi and wasm64-wasi
- Now compiling wasix for wasm32-wasi
- Added the first round of implementations for sockets within libc
- Added sendto and sendmsg implementations
- Added an implementation for recvmsg
- The receive operations can now process flags
- Fixed the data truncated flag
- Implemented the remaining socket functions for libc
- Added getifaddrs methods
- Simplified some of the socket syscalls
- Added the sendfile syscalls
- Added the sendfile syscalls
- Added more networking constructs
- Added more flags for file system operations
- Added more constants
- Added a function that allows the FD to be duplicated
- Added another dup fctrl
- Added the ability to get the protocol type
- Added more macros and constants needed for networking
- Added SEQPACKET
- Added getpid and fixed up chdir/getpwd
- Added getpid and fixed up chdir/getpwd
- Refactored the WASI interface
- Fixed the relative path and added events
- Made the GIT update an optional step
- Updated to point to the renamed WITX file
- Updated libc with the latest WASIX
- Implemented the futex syscalls
- Added a memory allocatio function to the thread spawn syscall
- Removed the malloc method parameter from thread_spawn
- Simplified the threading and bus callbacks so that they dont need named functions as parameters
- Now passing in a topic hash for the calls and subcalls
- Added support for multithreading in WebAssembly
- Added a fix so that the preopen initializer only runs once
- Implemented more operations for WASIX
- Implemented enhancements required for stack save and restore
- Added long jump capability for WASIX
- Added all the functionality needed to compile dash
- Added real signals to WASIX
- Finished the stack unwinding and winding functionality
- Finished the stack unwinding and winding functionality
- Added the vfork functionality
- Added spawning examples
- Further enhancements for more functions exposed to WASIX
- More functions enabled for WASIX
- Fixed the build which was missing the string source files
- Now automatically pulling the WITX properly again
- Modified the proc_join function so that it notifies what PID was exited
- Reduced the memory consumption of stack snapshots by moving them into the runtime
- Improve build script compatibility and resiliency
- ci: Update Github workflow


theduke:
Sorry, wrong repository!

PR #344, opened 2022-10-26
Adjust Makefile for LLVM trunk (16) as of 2022-11-08

glandium:
https://github.com/llvm/llvm-project/commit/1e4e2433bcd1a0296ef1043c462252f0d087d90c enabled sign-ext and mutable-globals by default, which adds corresponding __wasm_-prefixed #defines.


glandium:
ping @sunfishcode

sunfishcode:
Thanks!

PR #345, opened 2022-11-24
If `fd_readdir` returns a zero inode, call `fstatat` to get the inode value.

sunfishcode:
On some systems, `fd_readdir` may not implement the `d_ino` field and may set it to zero. When this happens, have wasi-libc call `fstatat` to get the inode number.

See the discussion in
https://github.com/WebAssembly/wasi-filesystem/issues/65 for details.

### DIFF
--- a/libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c
+++ b/libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c
@@ -88,6 +88,7 @@ struct dirent *readdir(DIR *dirp) {
     // the inode number is unknown. In that case, do an `fstatat` to get the
     // inode number.
     off_t d_ino = entry.d_ino;
+    unsigned char d_type = entry.d_type;
     if (d_ino == 0) {
       struct stat statbuf;
       if (fstatat(dirp->fd, dirent->d_name, &statbuf, AT_SYMLINK_NOFOLLOW) != 0) {
@@ -98,9 +99,15 @@ struct dirent *readdir(DIR *dirp) {
         return NULL;
       }
 
+      // Fill in the inode.
       d_ino = statbuf.st_ino;
+
+      // In case someone raced with us and replaced the object with this name
+      // with another of a different type, update the type too.
+      d_type = __wasilibc_iftodt(statbuf.st_mode & S_IFMT);
     }
     dirent->d_ino = d_ino;
+    dirent->d_type = d_type;
 
     dirp->cookie = entry.d_next;
     dirp->buffer_processed += entry_size;
--- a/libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c
+++ b/libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c
@@ -3,6 +3,9 @@
 // SPDX-License-Identifier: BSD-2-Clause
 
 #include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <fcntl.h>
 
 #include <assert.h>
 #include <wasi/api.h>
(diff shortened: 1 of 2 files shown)


libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c
```diff
@@ -38,6 +41,8 @@ static_assert(DT_UNKNOWN == __WASI_FILETYPE_UNKNOWN, "Value mismatch");
   } while (0)
 
 struct dirent *readdir(DIR *dirp) {
+  struct stat statbuf;
```
loganek, line 44: [minor] the scope of this variable could be reduced to the block starting in [L#93](https://github.com/WebAssembly/wasi-libc/pull/345/files#diff-39125f185a8b15746cc411f8d6fe6f79838f07c3bd95b773b12644a62da4eb80R93)

libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c
```diff
@@ -77,10 +82,23 @@ struct dirent *readdir(DIR *dirp) {
...
     memcpy(dirent->d_name, name, entry.d_namlen);
     dirent->d_name[entry.d_namlen] = '\0';
+
+    // `fd_readdir` implementations may set the inode field to zero if the
+    // the inode number is unknown. In that case, do an `fstatat` to get the
+    // inode number.
+    off_t d_ino = entry.d_ino;
+    if (d_ino == 0) {
+      if (fstatat(dirp->fd, dirent->d_name, &statbuf, AT_SYMLINK_NOFOLLOW) != 0) {
+        return NULL;
```
yamt, line 95: ENOENT etc can happen because of concurrent activities.
aborting on a fstat error here seems like an overreaction to me.

libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c
```diff
@@ -77,10 +82,23 @@ struct dirent *readdir(DIR *dirp) {
...
          offsetof(struct dirent, d_name) + entry.d_namlen + 1);
     struct dirent *dirent = dirp->dirent;
-    dirent->d_ino = entry.d_ino;
     dirent->d_type = entry.d_type;
     memcpy(dirent->d_name, name, entry.d_namlen);
     dirent->d_name[entry.d_namlen] = '\0';
+
+    // `fd_readdir` implementations may set the inode field to zero if the
+    // the inode number is unknown. In that case, do an `fstatat` to get the
+    // inode number.
```
yamt, line 91: is it (ie. d_ino can be 0) going to be a part of the official spec of fd_readdir?
otherwise i'm not sure if it's a good idea to have a workaround here.

libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c
sunfishcode: Good idea about `ENOENT`. I've made it continue on that.

libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c
sunfishcode: Yes, the idea here would be that we'd document that `d_ino` can be zero at the WASI level. I've now filed https://github.com/WebAssembly/wasi-filesystem/issues/71 to track this.

libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c
sunfishcode: Fixed.

loganek (approved):
You might consider squashing some commits before the merge to keep the history clean.

loganek:
I think that's an interesting scenario worth covering in https://github.com/WebAssembly/wasi-testsuite, would you mind adding a test for that?

sunfishcode:
Unfortunately, I don't know of a way to write a test for this with our current infrastructure. It requires an `fd_readdir` implementation which returns a 0 inode, which isn't something we can easily arrange. This might be doable in the future with virutalization though.

sunfishcode:
In terms of standardizing this behavior, worst case if we don't standardize it is that wasi-libc makes redundant `stat-at`calls which don't have any effect.

A potential concern is atomicity. With a separate stat-at call, there is a potential window between when one set of directory entries is read and when their inodes are read. My sense is that this isn't a problem, because directory traversal isn't atomic in general. If there are other processes creating and removing and renaming files in a directory at the same time as a `readdir` is iterating through the directory, the `readdir` isn't guaranteed to see an atomic snapshot.

That said, thinking about this led me to think about the case where a file is replaced by another filesystem object with a different type. That's straightforward to handle, so I've now added code to this PR to handle it.

sunfishcode:
There haven't been any objections to this approach on the spec side, and this doesn't break compatibility, so I'm going to go ahead and land this.

libc-bottom-half/cloudlibc/src/libc/dirent/readdir.c
```diff
@@ -77,10 +80,35 @@ struct dirent *readdir(DIR *dirp) {
...
+    off_t d_ino = entry.d_ino;
+    unsigned char d_type = entry.d_type;
+    if (d_ino == 0) {
+      struct stat statbuf;
+      if (fstatat(dirp->fd, dirent->d_name, &statbuf, AT_SYMLINK_NOFOLLOW) != 0) {
+	if (errno == ENOENT) {
+	  // The file disappeared before we could read it, so skip it.
+	  continue;
```
yamt, line 97: don't you need to advance buffer_processed?

jedisct1: +1

sunfishcode: Good catch! I've now filed https://github.com/WebAssembly/wasi-libc/pull/352 to fix that and one other bug.

PR #346, opened 2022-11-28
test: run a subset of tests using `libc-test`

abrown:
This change introduces a `test` directory that retrieves the `libc-test` suite, compiles a subset of the tests using `wasi-libc`, and runs them with Wasmtime. It also enables the running of these tests in CI.

For now, only a small subset of `libc-test` is used. This PR only introduces the ability to run tests and in the future more tests (and more kinds of tests) can be enabled. Currently only some `src/functional` tests are enabled but the `src/math` and `src/api` tests could also be useful.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -51,6 +51,7 @@ jobs:
         curl -sSfL https://github.com/llvm/llvm-project/releases/download/llvmorg-${{ matrix.clang_version }}/clang+llvm-${{ matrix.clang_version }}-x86_64-linux-gnu-ubuntu-18.04.tar.xz | tar xJf -
         export CLANG_DIR=`pwd`/clang+llvm-${{ matrix.clang_version }}-x86_64-linux-gnu-ubuntu-18.04/bin
         echo "$CLANG_DIR" >> $GITHUB_PATH
+        echo "CLANG_DIR=$CLANG_DIR" >> $GITHUB_ENV
         echo "CC=$CLANG_DIR/clang" >> $GITHUB_ENV
         echo "AR=$CLANG_DIR/llvm-ar" >> $GITHUB_ENV
         echo "NM=$CLANG_DIR/llvm-nm" >> $GITHUB_ENV
@@ -59,6 +60,22 @@ jobs:
     - name: Build libc
       shell: bash
       run: make -j4
+
+    - name: Test
+      shell: bash
+      # For Clang linking to work correctly, we need to place Clang's runtime
+      # library for `wasm32-wasi` in the right location (i.e., the `mkdir` and
+      # `cp` below).
+      run: |
+        cd test
+        make download
+        export WASI_DIR=$(realpath $CLANG_DIR/../lib/clang/${{ matrix.clang_version }}/lib/wasi/)
+        mkdir -p $WASI_DIR
+        cp download/lib/wasi/libclang_rt.builtins-wasm32.a $WASI_DIR
+        make test
+      # The older version of Clang does not provide the expected symbol for the
+      # test entrypoints: `undefined symbol: __main_argc_argv`.
+      if: matrix.os == 'ubuntu-latest' && matrix.clang_version != '10.0.0'
 
     - uses: actions/upload-artifact@v1
       with:


test/Makefile
```diff
@@ -0,0 +1,175 @@
...
+WASMTIME_URL ?= https://github.com/bytecodealliance/wasmtime/releases/download/v3.0.0/wasmtime-v3.0.0-x86_64-linux.tar.xz
+WASMTIME = $(DOWNDIR)/$(shell basename $(WASMTIME_URL) .tar.xz)/wasmtime
+
+download: $(LIBC_TEST) $(LIBRT) $(WASMTIME)
+
+$(DOWNDIR):
+	mkdir -p download
+
+$(LIBC_TEST): | $(DOWNDIR)
+	git clone --depth 1 $(LIBC_TEST_URL) $@
```
sbc100, line 34: Why not use a git submodule instead?

loganek:
Just a general comment about that; we've recently started a [wasi-testsuite](https://github.com/WebAssembly/wasi-testsuite/) project that aims to provide unified tests and runtime environment that's easily integrated with any WASM VM. Whereas the goal of wasi-testsuite is to validate WASI implementation of various VMs and the goal of the tests you just added is to verify the correctness of libc implementation, I think there will be an overlap. I wonder if we can somehow share some of the tests?
One option I can think of is moving all the wasi-libc tests to wasi-testsuite and run them as part of the CI in this repository, but the disadvantage of that is that not all the tests actually validate WASI layer (although I don't think this is such a big of the concern?).

test/Makefile
abrown: There were some concerns about licensing; some parts (unused here) are [licensed under GPL](http://nsz.repo.hu/git/?p=libc-test;a=blob;f=COPYRIGHT;h=67acf495aa00c48cf1b9334b12a94a83a28b6b8a;hb=HEAD). In discussions with @sunfishcode, we felt that downloading the code on demand and never releasing any of the libc-test parts avoided license contamination with this repo. What do you think?

sbc100: I don't see how doing a `git clone` programmatically is any different to using a git submodule, but IANAL..

abrown:
> One option I can think of is moving all the wasi-libc tests to wasi-testsuite and run them as part of the CI in this repository, but the disadvantage of that is that not all the tests actually validate WASI layer (although I don't think this is such a big of the concern?).

@loganek, I actually am agnostic about where these tests should live and even if libc-test should be the only suite to use (@sbc100 also suggested the [Open POSIX Test Suite](https://posixtest.sourceforge.net/) which is used by Emscripten). I do think, though, that given the trickiness of getting these tests to run (I had to iterate a bit on 93606b7), I would prefer to get something merged soon to start checking Wasmtime's pthread implementation. I'm in favor of wasi-testsuite in general but I don't think it is at the same abstraction level as libc-test in this case: e.g., for threads, we could add tests to wasi-testsuite to check that wasi-threads works as expected (and that would be valuable!) but we still need to check that the mapping from pthreads to wasi-threads was done correctly (which is what eventually libc-test could do).

test/Makefile
abrown: Yeah, I see what you mean; I guess I had thought that cloning it at a different time made it more optional (i.e., only those who want to run tests). I set up a meeting to talk to someone here about it because... IAANAL.

loganek:
> @loganek, I actually am agnostic about where these tests should live and even if libc-test should be the only suite to use (@sbc100 also suggested the [Open POSIX Test Suite](https://posixtest.sourceforge.net/) which is used by Emscripten). I do think, though, that given the trickiness of getting these tests to run (I had to iterate a bit on [93606b7](https://github.com/WebAssembly/wasi-libc/commit/93606b79ec39cd2f0b5759cae262aae86f6d1ea5)), I would prefer to get something merged soon to start checking Wasmtime's pthread implementation. I'm in favor of wasi-testsuite in general but I don't think it is at the same abstraction level as libc-test in this case: e.g., for threads, we could add tests to wasi-testsuite to check that wasi-threads works as expected (and that would be valuable!) but we still need to check that the mapping from pthreads to wasi-threads was done correctly (which is what eventually libc-test could do).

@abrown I didn't mean to block the PR; let's have something working merged so we can run more tests (I also started working on the WAMR implementation and currently compile pthread tests by myself, so having that as part of wasi-libc would be very valuable).
Regarding the abstraction layer - wasi-testsuite is aiming to focus mainly on the WASI interfaces and not standard library implementation (that's why I think having those tests in wasi-libc is the right thing to do), but we also have tests that use libc implementation if that simplifies testing. So I'm just calling out that there will be an overlap, but how to share tests (or whether we should do that at all) can probably be discussed later on.

abrown:
@loganek, makes sense; thanks for the heads up!

abrown:
@loganek, @sunfishcode, @sbc100: I see two future-facing "problems" with this PR and I wanted to note them here. I still feel we should merge something like this to make progress toward some type of testing for wasi-libc but I wanted to share these concerns as we review it:
 - the `test/Makefile` is very Linux-specific and requires a weird "put this .a file over here" hack to make Clang compile the tests; I've heard comments to the effect of "it's fine for a starter PR" but I wanted to bring it up to see if there is something cleaner I could have done
 - I have only enabled a subset of tests (the ones that passed!); ideally we would enable more by resolving whatever problems we encounter. But I don't want to be the sole owner of "make all the tests work"--does the current `Makefile` and the `libc-test` organization seem clear enough for others to hack on to add more tests? Is anyone else interested in helping out enabling tests? I can create an issue to track this but wanted to gauge how much interest there is in the testing effort.

sunfishcode:
@abrown My sense here is that this PR is ok with those concerns. The changes here are relatively contained so people who aren't running the tests aren't affected, so I think we can look at it as, this PR adds some value for some people, even if it doesn't do everything for everyone.

loganek:
Agree with @sunfishcode ,  and I'm ok with the current implementation. We can iterate on that if we see opportunity to make it better and unblock other usecases.

sunfishcode (approved):

PR #349, opened 2022-11-30
Install libtinfo5 on ubuntu.

sunfishcode:
We download LLVM releases built for Ubuntu 18 because LLVM doesn't always have builds for different versions, but the builds we use depend on libtinfo5 which isn't installed on Ubuntu 20 by default. So install it.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,6 +15,13 @@ jobs:
       with:
         submodules: true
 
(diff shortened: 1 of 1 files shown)


sunfishcode:
I'm going to go ahead and merge this as it fixes errors we're seeing in other PRs. I'm happy to address any follow-up comments.

PR #350, opened 2022-12-02
threads: Retrieve default stack size from __heap_base/__data_end

loganek:
When compiling with `-z stack-size` flag, only the main thread's stack size is set to the specified value and other threads use musl's default value. That's inconsistent with LLD's `-Wl,-stack_size`.

I think we can make it similar to musl's behavior, where thread's stack size can be set via `PT_GNU_STACK` program header (via `-Wl,-z,stack-size` flag).

Configuring stack size through `pthread_attr_t` still works as expected and overrides the defaults ([pthread_create.c](https://github.com/WebAssembly/wasi-libc/blob/be1ffd6a9eba1704085987482557c2a32724227f/libc-top-half/musl/src/thread/pthread_create.c#L362)).


libc-top-half/musl/src/env/__init_tls.c
```diff
@@ -16,6 +16,20 @@
 volatile int __thread_list_lock;
 
 #ifndef __wasilibc_unmodified_upstream
+
+extern unsigned char __heap_base;
+extern unsigned char __data_end;
+
+static inline void setup_default_stack_size()
+{
+	ptrdiff_t stack_size = &__heap_base - &__data_end;
```
sbc100, line 25: Recent version of wasm-ld will define `__stack_high` and `__stack_low` so we can be more explicit here on recent llvm versions (https://github.com/llvm/llvm-project/blob/4fd0c14a17f91fb7b0d188b509eb3dbdfbfec01a/lld/wasm/Symbols.h#L522-L528.

See https://reviews.llvm.org/D135910.

Perhaps we could declare weak references to them and fall back to  __heap_base/__data_end when they are not found?

sbc100, line 25: For this legacy path, we need to also take into account the case where somebody linked with `--stack-first`.   In this case the SP grows down address zero and the static data is after the stack region.  Maybe we can detect this case by compare SP to `__global_base`?

loganek: Great, didn't know about that; I'll update the code.

loganek: Good point, will handle that case.

libc-top-half/musl/src/env/__init_tls.c
```diff
@@ -16,6 +16,36 @@
 volatile int __thread_list_lock;
 
 #ifndef __wasilibc_unmodified_upstream
+
+extern unsigned char __heap_base;
```
sbc100, line 20: Perhaps add a commend like `These symbols are generated by wasm-ld`  and perhaps mentions why two of them are weak and add a `TODO` to rely on them exclusively once we drop support for llvm < N?

libc-top-half/musl/src/env/__init_tls.c
```diff
@@ -16,6 +16,36 @@
...
+extern weak unsigned char __stack_high;
+extern weak unsigned char __stack_low;
+
+static inline void setup_default_stack_size()
+{
+	ptrdiff_t stack_size;
+
+	if (&__stack_high)
+		stack_size = &__stack_high - &__stack_low;
+	else
```
sbc100, line 32: IIIUC the coding style of musl puts opening braces for (non-function) blocks on the same line as the keyword.

abrown (commented):
This makes sense to me! I'll let @sbc100 be the final approver since he had some comments already.

yamt (commented):
lgtm

sbc100 (approved):

loganek:
I think it's ready for merge @sbc100 @abrown ?

abrown:
I was about to merge this but now it has a merge conflict; @loganek, can you resolve that?

loganek:
@abrown should be ready for merge now

PR #352, opened 2022-12-05
Fix logic errors in the zero-inode path.

sunfishcode:
I've now tested the zero-inode path on a Wasm engine specially-modified to have `fd_readdir` set inode numbers to zero. Fix two bugs this turned up:
 - Increment `buffer_processed`, as noticed by @yamt
 - Don't do an `fstatat` on "..", because that references a path outside of the directory, which gets a permission-denied error.


yamt (approved):
lgtm

PR #353, opened 2022-12-07
__wasi_thread_spawn: stop truncating the return value

yamt:
as __wasi_errno_t is uint16_t, with the current coding, __pthread_create will never see negative return values from wasi:thread_spawn.
eg. (int)(uint16_t)-1 == 65535.

### DIFF
--- a/libc-bottom-half/sources/__wasilibc_real.c
+++ b/libc-bottom-half/sources/__wasilibc_real.c
@@ -665,8 +665,7 @@ int32_t __imported_wasi_thread_spawn(int32_t arg0) __attribute__((
     __import_name__("thread_spawn")
 ));
 
-__wasi_errno_t __wasi_thread_spawn(void* start_arg) {
-	int32_t ret = __imported_wasi_thread_spawn((int32_t) start_arg);
-    return (uint16_t) ret;
+int32_t __wasi_thread_spawn(void* start_arg) {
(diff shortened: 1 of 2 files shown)


libc-bottom-half/sources/__wasilibc_real.c
```diff
@@ -665,8 +665,7 @@ int32_t __imported_wasi_thread_spawn(int32_t arg0) __attribute__((
     __import_name__("thread_spawn")
 ));
 
-__wasi_errno_t __wasi_thread_spawn(void* start_arg) {
-	int32_t ret = __imported_wasi_thread_spawn((int32_t) start_arg);
-    return (uint16_t) ret;
+int32_t __wasi_thread_spawn(void* start_arg) {
+	return __imported_wasi_thread_spawn((int32_t) start_arg);
```
sbc100, line 669: Should we match the indentation that is used in the rest of this file?

yamt:
> So negative values here correspond to -errno values here?

it's documented as `a negative error code.'

libc-bottom-half/sources/__wasilibc_real.c
yamt: oops, fixed

sbc100 (approved):

loganek (approved):
LGTM

abrown (approved):

PR #360, opened 2022-12-13
__init_tp: Initialize TID to non-zero value

yamt:
__init_tp: Initialize TID to non-zero value

This fixes TID-based locking used within libc.

Also, initialize detach_state.

cf. https://github.com/WebAssembly/wasi-threads/pull/16

### DIFF
--- a/libc-top-half/musl/src/env/__init_tls.c
+++ b/libc-top-half/musl/src/env/__init_tls.c
@@ -70,6 +70,18 @@ int __init_tp(void *p)
 	td->tid = __syscall(SYS_set_tid_address, &__thread_list_lock);
 #else
 	setup_default_stack_size();
+	td->detach_state = DT_JOINABLE;
+	/*
+	 * Initialize the TID to a value which doesn't conflict with
+	 * host-allocated TIDs, so that TID-based locks can work.
+	 *
+	 * Note:
+	 * - Host-allocated TIDs range from 1 to 0x1fffffff. (inclusive)
+	 * - __tl_lock and __lockfile uses TID 0 as "unlocked".
+	 * - __lockfile relies on the fact the most significant two bits
+	 *   of TIDs are 0.
+	 */
+	td->tid = 0x3fffffff;
 #endif
 	td->locale = &libc.global_locale;
 	td->robust_list.head = &td->robust_list.head;


libc-top-half/musl/src/env/__init_tls.c
```diff
@@ -31,6 +31,23 @@ int __init_tp(void *p)
...
+	 * For longer term, we need a better solution.
+	 * See: https://github.com/WebAssembly/wasi-threads/issues/15
+	 *
+	 * Note:
+	 * - positive values can conflict with host-allocated TIDs.
+	 * - __tl_lock and __lockfile uses TID 0 as "unlocked".
+	 * - __lockfile relies on the fact the most significant two bits
+	 *   of TIDs are 0.
+	 */
+	td->tid = 0x3fffffff;
```
loganek, line 50: > Initialize the TID to a large value, which is hopefully
> less likely to conflict with host-allocated TIDs

So, that assumption is correct only if the host assigns the ID incrementally, which might not be the case (host can decide the implementation).

yamt: > > Initialize the TID to a large value, which is hopefully
> > less likely to conflict with host-allocated TIDs
> 
> So, that assumption is correct only if the host assigns the ID incrementally, which might not be the case (host can decide the implementation).

right. this is a kludge at best.

i think a real fix involves spec changes. see https://github.com/WebAssembly/wasi-threads/issues/15

loganek: I'd suggest let's wait with this PR then until the https://github.com/WebAssembly/wasi-threads/issues/15 is resolved

loganek: I think we can update the PR based on the discussion in https://github.com/WebAssembly/wasi-threads/issues/15 @yamt ?

yamt:
i updated the comment after https://github.com/WebAssembly/wasi-threads/pull/16

libc-top-half/musl/src/env/__init_tls.c
yamt: > I think we can update the PR based on the discussion in [WebAssembly/wasi-threads#15](https://github.com/WebAssembly/wasi-threads/issues/15) @yamt ?

done

abrown (approved):
I think this is fine; I would want to link to the wasi-threads heading where this is discussed, but that's personal preference.

loganek (approved):

PR #362, opened 2022-12-13
Enable flockfile and friends

yamt:


yamt:
tested with https://github.com/WebAssembly/wasi-libc/pull/361 and https://github.com/WebAssembly/wasi-libc/pull/360

sbc100 (approved):

PR #364, opened 2022-12-16
Use wasi thread_exit instead of proc_exit where appropriate

yamt:
https://github.com/WebAssembly/wasi-threads/issues/7
https://github.com/WebAssembly/wasi-threads/pull/17


sbc100 (commented):
lgtm assuming we have consensus on https://github.com/WebAssembly/wasi-threads/issues/7

abrown (commented):
Yeah, similar comment from me: let's wait on https://github.com/WebAssembly/wasi-threads/issues/7

sbc100:
> Yeah, similar comment from me: let's wait on [WebAssembly/wasi-threads#7](https://github.com/WebAssembly/wasi-threads/issues/7)

In that case we should completely remove `pthread_exit` from wasi-libc until that issue is resolved.   Implementing pthread_exit using proc_exit (exit the whole process) is not correct, and we also want to know how many folks/app are going to just fail to build without `pthread_exit` existing at all.

yamt:
> Yeah, similar comment from me: let's wait on [WebAssembly/wasi-threads#7](https://github.com/WebAssembly/wasi-threads/issues/7)

yes, that's why i made this a draft.

PR #365, opened 2022-12-17
Reduce over-allocation of stack

yamt:


sbc100 (approved):
Perhaps should be titled something like "reduce over-allocation of stack"?

loganek (approved):

yamt:
> 


> Perhaps should be titled something like "reduce over-allocation of stack"?

sure. i changed the title of the PR.

PR #366, opened 2022-12-19
Disable pthread_exit for now

yamt:
The current wasi-threads has no thread-exit functionality. Thus it isn't straightforward to implement pthread_exit without leaking thread context. This commit simply disables pthread_exit for now.

Also, instead of abusing `wasi_proc_exit` for thread exit, make `wasi_thread_start` return.

Note: `wasi_proc_exit` is supposed to terminate all threads in the "process", not only the calling thread.

Note: Depending on the conclusion of the discussion about `wasi_thread_exit`, we might revisit this change later.

References:
https://github.com/WebAssembly/wasi-threads/issues/7 https://github.com/WebAssembly/wasi-threads/pull/17

### DIFF
--- a/expected/wasm32-wasi-pthread/defined-symbols.txt
+++ b/expected/wasm32-wasi-pthread/defined-symbols.txt
@@ -193,7 +193,6 @@ __progname
 __progname_full
 __pthread_cond_timedwait
 __pthread_create
-__pthread_exit
 __pthread_join
 __pthread_key_create
 __pthread_key_delete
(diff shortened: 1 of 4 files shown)


yamt:
an alternative to https://github.com/WebAssembly/wasi-libc/pull/364

abrown (approved):
LGTM as a temporary fix until https://github.com/WebAssembly/wasi-threads/issues/7 is resolved.

PR #369, opened 2022-12-20
test: add tests for pthread API

abrown:
This change runs the pthread tests available in `libc-test` when `THREAD_MODEL=posix`. This is currently a work-in-progress, since there are custom build steps necessary to get this all to (mostly) work. Currently this might look like:

```console
$ cd test
$ make run THREAD_MODEL=posix \
    ENGINE=~/Code/wasmtime/target/debug/wasmtime
    CC=~/Code/llvm-project/build/bin/clang
```

What are the current issues?
- this requires a special build of Wasmtime that includes the `wasi-threads` implementation; this is shown above as built from source
- under `THREAD_MODEL=posix`, this requires Clang to understand the `--export-memory` flag which, though merged in https://reviews.llvm.org/D135898, is not available for download yet (but can be built from latest source as shown above)
- not all of the tests yet pass and some do not build: this is to be expected in several cases (PI? robust?) but there is perhaps more that can be done here to enable more of the pthread API. Only the working tests are included in this commit.


test/Makefile
```diff
@@ -116,23 +128,54 @@ ifeq ($(origin CC), default)
...
 
+# Configure support for threads.
+ifeq ($(THREAD_MODEL), single)
+CFLAGS += --target=wasm32-wasi -mthread-model single
+endif
+ifeq ($(THREAD_MODEL), posix)
+# Specify the tls-model until LLVM 15 is released (which should contain
+# https://reviews.llvm.org/D130053).
+CFLAGS += --target=wasm32-wasi-pthread -mthread-model posix -pthread -ftls-model=local-exec
+# NOTE: `--export-memory` is invalid until https://reviews.llvm.org/D135898 is
```
sbc100, line 143: The change related to being able to export or import the memory with a specific name  `--export-memory` on its own (using the default name) I think has existed for a long time.

abrown: I was only able to see the `--export-memory` flag in a built-from-tree version of Clang. Should I be able to find it in some binary version of Clang?

sbc100: Oh maybe you are right.. exporting of the memory was always just the default.

Remind me why you need to both import and export the memory? .. emscripten doesn't do that today.

abrown: Uh, well, not by choice:
- we want to import the shared memory to avoid magic: the shared memory is created outside of the instance and passed in as an import
- then, Wasmtime wants me to also export the memory because this is how the Wiggle integration figures out where buffers are at when a WASI function is called with a buffer as a parameter.

abrown:
cc: @loganek, I had forgotten about https://github.com/WebAssembly/wasi-threads/pull/11 (sorry!) but perhaps some of the tests here can be adapted for use there in the spec? Also, note here how much more complex the "build and run" scripting must be to get working examples.

test/Makefile
sbc100: Hmm.. that seems like unique concern of wasmtime and/or wiggle and perhaps we should at least make a note of that.  Perhaps add `--import-memory` (which all runtimes will need) on the line above and then leave `--export-memory` on a line by itself with a comment?   Perhaps something like "with the instnace-per-thread model we always need to import the memory, and under wasmtime we also need re-export this imported memory.  The ability to both export and import a memory is only available in recent versions of wasm-ld".

yamt:
i needed https://github.com/WebAssembly/wasi-libc/pull/372 for tls tests to pass on toywasm.
isn't it the case for wasmtime?

test/Makefile
sbc100: Actually I think its the `--export-memory` that is wasmtime specific.

Any runtime implementing instance-per-thread will need to use imported memory I think.  i.e.  The embedder will need to create the shared memory and use it in module instantiation on each thread.

yamt: > Actually I think its the `--export-memory` that is wasmtime specific.
> 
> Any runtime implementing instance-per-thread will need to use imported memory I think. i.e. The embedder will need to create the shared memory and use it in module instantiation on each thread.

why?
an implementation can choose to do it without imported memory.
as far as i know, the current implementation of wamr or toywasm don't need it.

sbc100: If the memory is defined inside the module, then each instantiation of the module creates a new memory.  In order to share memory between the instances there would need to be just one memory, shared between many instances.

sbc100: The current/initial wasi-threads proposal is based on the instance-per-thread model, which means performing instantiation once per thread.   Unless we change the definition of instantiation this means that the shared memory  cannot be declared within the module.

yamt: wasi-threads introduced a new way to instantiate a module, for which shared memory is somehow inherited from the parent instance.
at least it's my reading of https://github.com/WebAssembly/wasi-threads#detailed-design-discussion and what i've implemented in toywasm.

sbc100: If that was true there would be no way to implement/polyfill wasi-thread on top of webworkers  or node workers.  My understanding is that wasi-threads tries to build on / match the worker model, and does not depend on modification to the instantiation process/definition of core wasm.

sbc100: I agree that point 2 of https://github.com/WebAssembly/wasi-threads#detailed-design-discussion could use some clarification.  I think it should refer explicitly to memories that are provided to the module as *imports*.

yamt: > If that was true there would be no way to implement/polyfill wasi-thread on top of webworkers or node workers. 

i agree and it's my understanding.

> My understanding is that wasi-threads tries to build on / match the worker model, and does not depend on modification to the instantiation process/definition of core wasm.

do you mean something like the following? if so, do you have answers to Q1-4?

1. detect if the wasm module is for wasi-threads (Q1 how?)
2. if it is, and it has a memory import, create a matching shared memory. (Q2 what properties to match and how? just use min/max of the import? Q3 assuming it's always have a fixed name like "env:memory"? Q4 what to do for multi memory?)
4. put the shared memory in the import-objects when instantiating the module
5. on `wasi_spawn_thread`, the same import-objects which contains the shared memory is used to instantiate the child instance.
6. once all threads including the main thread is gone, dispose the automatically created shared memory.

sbc100: That sounds roughly correct, although I'm sure we need to spec all those details.  I think some of them can be left up to the runtime. 

Q1.  I'm not sure this is really necessary, but I can think of a few possible ways to do it:
  1. The module imports a shared memory
  2. The module imports wasi_thread_start API
I think its up the runtime though.  For example, it could be a valid choice for the runtime to support threads only `--wasi-threads` was passed explicitly on the command line.

Q2: Can't this be left up to the runtime and the user?  As long as the set of imports provided to the module match the import spec that module can be loaded.  A simple/reasonable implementation might create a shared memory with the minimal size needed to satisfy the import specification.

Q3: I don't think the name has to matter for the memory import, although wasi does currently specify a special/reserved name for the memory export.  We could do the same for the memory import, but I'm not sure that it is necessary.

Q4: I don't think we've discussed this yet, but the runtime could simply create N new shared memories to satisfy the imports.

yamt: > That sounds roughly correct, although I'm sure we need to spec all those details. I think some of them can be left up to the runtime.
> 
> Q1. I'm not sure this is really necessary, but I can think of a few possible ways to do it:
> 
>     1. The module imports a shared memory
> 
>     2. The module imports wasi_thread_start API
>        I think its up the runtime though.  For example, it could be a valid choice for the runtime to support threads only `--wasi-threads` was passed explicitly on the command line.
> 
> 
> Q2: Can't this be left up to the runtime and the user? As long as the set of imports provided to the module match the import spec that module can be loaded. A simple/reasonable implementation might create a shared memory with the minimal size needed to satisfy the import specification.
> 
> Q3: I don't think the name has to matter for the memory import, although wasi does currently specify a special/reserved name for the memory export. We could do the same for the memory import, but I'm not sure that it is necessary.
> 
> Q4: I don't think we've discussed this yet, but the runtime could simply create N new shared memories to satisfy the imports.

are you assuming that the only reason to import shared memory is wasi-threads? i'm not sure if it's a good assumption.

sbc100: I don't think there is any need to make that assumption. no.  If the runtime wants a static signal it could use the import of `wasi_thread_create`.

I'm not sure a runtime necessarily needs to know statically that a program is mulithreaded, though.  A runtime could just implement `wasi_thread_create` and have it create a new thread without necessarily knowing up front couldn't it?   This wouldn't be very useful without an imported shared memory though.... so we might want to specify that a program that calls `wasi_thread_create` needs to import at least one shared memory?

yamt: well, "automatically create shared memory which satisfies imports" is a very wasi-threads specific behavior, isn't it?

abrown: Reading through this chain so far, I think @sbc100 has explained things well. We need to import a shared memory and in Wasmtime we will know to do so because of configuration that turns on Wasm threads and wasi-threads. My takeaway is that we could clarify the README a bit more in this regard but I do want to be a bit cautious about overspecifying to allow different runtimes to implement the configuration as they wish.

yamt: ok, let me see how i can amend my implementation. after that experiment, maybe i will submit a README patch.

yamt: done.

https://github.com/WebAssembly/wasi-threads/pull/19#discussion_r1056718468
https://github.com/yamt/toywasm/commit/4d81846e3aa6f01f97e6aef0a51238fd18f6ecf1
https://github.com/WebAssembly/wasi-threads/pull/20

test/Makefile
```diff
@@ -15,6 +15,10 @@ test: run
 # directory (keeping with the `wasi-libc` conventions).
 OBJDIR ?= $(CURDIR)/build
 DOWNDIR ?= $(CURDIR)/download
+# Like the top-level `wasi-libc` Makefile, here we can decide whether to test
+# with threads (`posix`) or without (`single`). IMPORTANT: the top-level include
```
loganek, line 19: Feels like yet another argument in favor of [cmake](https://github.com/WebAssembly/wasi-libc/pull/330)

abrown:
Note to self: this should get moved to wasi-sdk.

PR #371, opened 2022-12-22
wasi_thread_start: add a comment

yamt:


sbc100 (approved):
Hmm.. perhaps we should use a .S or .s file here?  lgtm to the comment though.

sbc100 (approved):

yamt:
> Hmm.. perhaps we should use a .S or .s file here?

probably yes.

yamt:
> > Hmm.. perhaps we should use a .S or .s file here?
> 
> probably yes.

once the api/abi got stable.

abrown (approved):

libc-top-half/musl/src/thread/pthread_create.c
```diff
@@ -277,6 +277,11 @@ static int start_c11(void *p)
 __attribute__((export_name("wasi_thread_start")))
 void wasi_thread_start(int tid, void *p)
 {
+	/*
+	 * Note: it's fragile to implement wasi_thread_start in C.
+	 * On entry, we don't even have C stack (__stack_pointer)
+	 * set up. Be careful when modifying this function.
+	 */
```
loganek, line 284: I actually wonder if the function, even in its current state, is safe and won't cause memory corruption? Whatever value is currently set to __stack_pointer is not correct; I'd assume the function will reference another thread's (main thread's?) stack (e.g. for local variables?) and overwrite the data. Am I missing something?

yamt: > I presume it depends on the compiler or even it's optimization flags.

right. it's what i meant by "fragile".

when i disassembled it, it seemed ok. so i didn't bother to rewrite it in asm for now.

loganek: Yeah, I guess this depends on the optimization level, if you exclude `-O2` it actually does include stack operations. I was going to fix that in some of the WAMR examples by implementing the method in asm but then I notice @sbc100 's comment and your followup PR. Thansk.

PR #372, opened 2022-12-22
Initialize TLS of the main thread

yamt:


yamt:
does anyone have any concerns on this?

sbc100:
If I remember correctly you don't need to call `__wasm_init_tls` on the main thread.  

Yup, the main thread's TLS region gets setup automatically during module instantiation:  https://github.com/llvm/llvm-project/blob/26f83b43b16a6efde8769d4bdf12779ffd8ef874/lld/wasm/Writer.cpp#L1207-L1209

yamt:
> If I remember correctly you don't need to call `__wasm_init_tls` on the main thread.
> 
> Yup, the main thread's TLS region gets setup automatically during module instantiation: https://github.com/llvm/llvm-project/blob/26f83b43b16a6efde8769d4bdf12779ffd8ef874/lld/wasm/Writer.cpp#L1207-L1209

ok. thank you.
another reason to require llvm >=15.

PR #376, opened 2023-01-06
Implement the critical part of wasi_thread_start in asm

yamt:
It's fragile to set up the critical part of C environment in C.

### DIFF
--- a/libc-top-half/musl/src/thread/pthread_create.c
+++ b/libc-top-half/musl/src/thread/pthread_create.c
@@ -238,6 +238,11 @@ struct start_args {
 	volatile int control;
 	unsigned long sig_mask[_NSIG/8/sizeof(long)];
 #else
+	/*
+	 * Note: the offset of the "stack" member in this structure is
+	 * hardcoded in wasi_thread_start.
+	 */
+	void *stack;
 	void *(*start_func)(void *);
 	void *start_arg;
 	void *tls_base;
@@ -274,14 +279,18 @@ static int start_c11(void *p)
 	return 0;
 }
 #else
-__attribute__((export_name("wasi_thread_start")))
-void wasi_thread_start(int tid, void *p)
+
+/*
+ * We want to ensure wasi_thread_start is linked whenever
+ * pthread_create is used. The following reference is to ensure that.
+ * Otherwise, the linker doesn't notice the dependency because
+ * wasi_thread_start is used indirectly via a wasm export.
+ */
+void wasi_thread_start(int tid, void *p);
+hidden void *__dummy_reference = wasi_thread_start;
+
+hidden void __wasi_thread_start_C(int tid, void *p)
 {
-	/*
-	 * Note: it's fragile to implement wasi_thread_start in C.
-	 * On entry, we don't even have C stack (__stack_pointer)
-	 * set up. Be careful when modifying this function.
-	 */
 	struct start_args *args = p;
   	__asm__(".globaltype __tls_base, i32\n"
 			"local.get %0\n"
@@ -293,11 +302,6 @@ void wasi_thread_start(int tid, void *p)
 	// whichever thread (parent or child) reaches this point first can proceed
 	// without waiting.
 	atomic_store((atomic_int *) &(self->tid), tid);
-	// Set the stack pointer.
-  	__asm__(".globaltype __stack_pointer, i32\n"
-			"local.get %0\n"
-			"global.set __stack_pointer\n"
-			:: "r"(self->stack));
 	// Execute the user's start function.
 	__pthread_exit(args->start_func(args->start_arg));
 }
@@ -501,6 +505,7 @@ int __pthread_create(pthread_t *restrict res, const pthread_attr_t *restrict att
 	/* Correct the stack size */
 	new->stack_size = stack - stack_limit;
 
+	args->stack = new->stack; /* just for convenience of asm trampoline */
 	args->start_func = entry;
 	args->start_arg = arg;
 	args->tls_base = (void*)new_tls_base;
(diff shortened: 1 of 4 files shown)


libc-top-half/musl/src/thread/pthread_create.c
```diff
@@ -274,14 +279,18 @@ static int start_c11(void *p)
...
+
+hidden void __wasi_thread_start_C(int tid, void *p)
 {
-	/*
-	 * Note: it's fragile to implement wasi_thread_start in C.
-	 * On entry, we don't even have C stack (__stack_pointer)
-	 * set up. Be careful when modifying this function.
-	 */
 	struct start_args *args = p;
   	__asm__(".globaltype __tls_base, i32\n"
```
sbc100, line 295: Would it make sense to move this into the asm file too and have `__wasi_thread_start_C` take tls_base as a third argument?

libc-top-half/musl/src/thread/wasm32/wasi_thread_start.s
```diff
@@ -0,0 +1,26 @@
+	.text
+
+	.export_name	wasi_thread_start, wasi_thread_start
+
+	.globaltype	__stack_pointer, i32
+	.functype	__wasi_thread_start_C (i32, i32) -> ()
+
+	.hidden	wasi_thread_start
```
sbc100, line 8: Logically (since this is an exported function) it maybe shouldn't be hidden?

libc-top-half/musl/src/thread/wasm32/wasi_thread_start.s
```diff
@@ -0,0 +1,26 @@
+	.text
+
+	.export_name	wasi_thread_start, wasi_thread_start
+
+	.globaltype	__stack_pointer, i32
```
sbc100, line 5: In the long run we might want to make this `.S` rather than a `.s` so we can using macros like `__wasm64__` to detect pointer size. e.g.: https://github.com/emscripten-core/emscripten/blob/0d7014e1b20ead9739f1fce54b316152413601c3/system/lib/compiler-rt/stack_ops.S#L6-L12



For now this seems fine though.

sbc100 (commented):
Nice!

libc-top-half/musl/src/thread/pthread_create.c
yamt: IMO it's better to make the asm minimum.
unlike stack pointer, it's simple to avoid using tls in C.

libc-top-half/musl/src/thread/wasm32/wasi_thread_start.s
yamt: as far as i know, symbol visibility here is a completely separate concept from wasm export.

libc-top-half/musl/src/thread/pthread_create.c
sbc100: But this *is* asm code being written inline inside of C doe, which just seems more complex than just doing it in the asm file.  e.g. it depends on this rather obscure gcc asm syntax (I personally have to look this up every time).

I probably wouldn't argue for this if it wasn't the case that we are adding a new asm already and it defines a function that is the caller of this function.  Given we have a logical place to move this inline asm it makes sense to me that it live there.

Its maybe just comes down to distaste, and perhaps mistrust of the `__asm__` syntax.  IIRC, the support for inline asm in for wasm in clang is pretty limited and not very well tested.  we *kind* of made it work enough for this kind of simple case, but anything more complex and I wouldn't bet on it.

yamt: > But this _is_ asm code being written inline inside of C code, which seems more complex than just doing it in the asm file. e.g. it depends on this rather obscure gcc asm syntax (I personally have to look this up every time).
> 
> I probably wouldn't argue for this if it wasn't the case that we are adding a new asm already and it defines a function that is the caller of this function. Given we have a logical place to move this inline asm it makes sense to me that it live there.
> 
> Its maybe just comes down to distaste, and perhaps mistrust of the `__asm__` syntax. IIRC, the support for inline asm in for wasm in clang is pretty limited and not very well tested. we _kind_ of made it work enough for this kind of simple case, but anything more complex and I wouldn't bet on it.

ok. i didn't know that the support of inline asm is limited. (do you have any references?)
i moved __tls_base stuff to asm.

sbc100: @sunfishcode might know more details as he did most of the work for make it work at all.  I don't think it ever got much very testing or attention.  IIUC until wasi-libc, emscripten was the only place it was used at all.   Basic things like this evidently work fine, which is great, but I wouldn't bet on much more.

libc-top-half/musl/src/thread/wasm32/wasi_thread_start.s
```diff
@@ -0,0 +1,31 @@
...
+
+	# Set up the minimum C environment.
+	# Note: offsetof(start_arg, stack) == 0
+	local.get   1  # start_arg
+	i32.load    0  # stack
+	global.set  __stack_pointer
+
+	local.get   1  # start_arg
+	i32.load    4  # tls_base
+	global.set  __tls_base
```
sbc100, line 24: Would it be simpler just to pass these as two extra args to __wasi_thread_start_C?  This would avoid using memory at all, be smaller in terms of code size, and be less brittle in terms of struct offsets.

yamt: > Would it be simpler just to pass these as two extra args to __wasi_thread_start_C? This would avoid using memory at all, be smaller in terms of code size, and be less brittle in terms of struct offsets.

i'm not sure what you mean.
do you mean to save a pthread_self() call in the C function?

sbc100 (approved):

loganek (approved):

PR #377, opened 2023-01-07
Don't use sbrk(0) to determine the initial heap size

alexcrichton:
This commit changes the `try_init_allocator` function as part of dlmalloc to not use `sbrk(0)` to determine the initial heap size. The purpose of this function is to use the extra memory at the end of linear memory for the initial allocation heap before `memory.grow` is used to allocate more memory. To learn the extent of this region the code previously would use `sbrk(0)` to find the current size of linear memory. This does not work, however, when other systems have called `memory.grow` before this function is called. For example if another allocator is used or if another component of a wasm binary grows memory for its own purposes then that memory will be incorrectly claimed to be owned by dlmalloc.

Instead this commit rounds up the `__heap_base` address to the nearest page size, since that must be allocatable. Otherwise anything above this rounded address is assumed to be used by something else, even if it's addressable.

### DIFF
--- a/dlmalloc/src/malloc.c
+++ b/dlmalloc/src/malloc.c
@@ -5222,6 +5222,9 @@ static void try_init_allocator(void) {
   /* Check that it is a first-time initialization. */
   assert(!is_initialized(gm));
 
+  /* Initialize mstate. */
+  ensure_initialization();
+
   char *base = &__heap_base;
   // Try to use the linker pseudo-symbol `__heap_end` for the initial size of
   // the heap, but if that's not defined due to LLVM being too old perhaps then
@@ -5247,9 +5250,6 @@ static void try_init_allocator(void) {
   if (initial_heap_size <= MIN_CHUNK_SIZE + TOP_FOOT_SIZE + MALLOC_ALIGNMENT) {
     return;
   }
-
-  /* Initialize mstate. */
-  ensure_initialization();
 
   /* Initialize the dlmalloc internal state. */
   gm->least_addr = base;
--- a/dlmalloc/src/malloc.c
+++ b/dlmalloc/src/malloc.c
@@ -5222,8 +5222,20 @@ static void try_init_allocator(void) {
   assert(!is_initialized(gm));
 
   char *base = (char *)&__heap_base;
-  /* Calls sbrk(0) that returns the initial memory position. */
-  char *init = (char *)CALL_MORECORE(0);
+  // Round up `base` to the nearest `PAGESIZE`. The initial size of linear
+  // memory will be at least the heap base to this page boundary, and it's then
+  // assumed that the initial linear memory image was truncated at that point.
+  // While this reflects the default behavior of `wasm-ld` it is also possible
+  // for users to craft larger linear memories by passing options to extend
+  // beyond this threshold. In this situation the memory will not be used for
+  // dlmalloc.
+  //
+  // Note that `sbrk(0)`, or in dlmalloc-ese `CALL_MORECORE(0)`, is specifically
+  // not used here. That captures the current size of the heap but is only
+  // correct if the we're the first to try to grow the heap. If the heap has
+  // grown elsewhere, such as a different allocator in place, then this would
+  // incorrectly claim such memroy as our own.
+  char *init = (char*) (((size_t) base + PAGESIZE - 1) & ~(PAGESIZE - 1));
   int initial_heap_size = init - base;
 
   /* Check that initial heap is long enough to serve a minimal allocation request. */
--- a/dlmalloc/src/malloc.c
+++ b/dlmalloc/src/malloc.c
@@ -5214,15 +5214,18 @@ static void internal_inspect_all(mstate m,
 /* ------------------ Exported try_init_allocator -------------------- */
 
 /* Symbol marking the end of data, bss and explicit stack, provided by wasm-ld. */
-extern unsigned char __heap_base;
+extern char __heap_base;
+extern char __heap_end __attribute__((__weak__));
 
 /* Initialize the initial state of dlmalloc to be able to use free memory between __heap_base and initial. */
 static void try_init_allocator(void) {
   /* Check that it is a first-time initialization. */
   assert(!is_initialized(gm));
 
-  char *base = (char *)&__heap_base;
-  // Round up `base` to the nearest `PAGESIZE`. The initial size of linear
+  char *base = &__heap_base;
+  // Try to use the linker pseudo-symbol `__heap_end` for the initial size of
+  // the heap, but if that's not defined due to LLVM being too old perhaps then
+  // round up `base` to the nearest `PAGESIZE`. The initial size of linear
   // memory will be at least the heap base to this page boundary, and it's then
   // assumed that the initial linear memory image was truncated at that point.
   // While this reflects the default behavior of `wasm-ld` it is also possible
@@ -5235,7 +5238,9 @@ static void try_init_allocator(void) {
   // correct if the we're the first to try to grow the heap. If the heap has
   // grown elsewhere, such as a different allocator in place, then this would
   // incorrectly claim such memroy as our own.
-  char *init = (char*) (((size_t) base + PAGESIZE - 1) & ~(PAGESIZE - 1));
+  char *init = &__heap_end;
+  if (init == NULL)
+    init = (char*) (((size_t) base + PAGESIZE - 1) & ~(PAGESIZE - 1));
   int initial_heap_size = init - base;
 
   /* Check that initial heap is long enough to serve a minimal allocation request. */
--- a/dlmalloc/src/malloc.c
+++ b/dlmalloc/src/malloc.c
@@ -5238,10 +5238,10 @@ static void try_init_allocator(void) {
   // correct if the we're the first to try to grow the heap. If the heap has
   // grown elsewhere, such as a different allocator in place, then this would
   // incorrectly claim such memroy as our own.
-  char *init = &__heap_end;
-  if (init == NULL)
-    init = (char*) (((size_t) base + PAGESIZE - 1) & ~(PAGESIZE - 1));
-  int initial_heap_size = init - base;
+  char *end = &__heap_end;
+  if (end == NULL)
+    end = (char*) page_align((size_t) base);
+  size_t initial_heap_size = end - base;
 
   /* Check that initial heap is long enough to serve a minimal allocation request. */
   if (initial_heap_size <= MIN_CHUNK_SIZE + TOP_FOOT_SIZE + MALLOC_ALIGNMENT) {


yamt:
it breaks the case where initial heap is more than 1 page, doesn't it?

alexcrichton:
This doesn't break that use case, but it does make it less efficient. The default output of LLD doesn't ever have more than one page at the end of memory (or at least not that I'm aware of), and I'm not sure if such a setup is commonly used in the wild. If it is then I think it would perhaps make sense to add a new pseudo-symbol to LLD such as `__heap_base` along the lines of `__heap_initial_end` or something like that, and then that can be used instead of the round-up-to-page-size calculation done here.

TerrorJack:
Note there's [`__heap_end`](https://reviews.llvm.org/D136110) added re discussion in https://github.com/WebAssembly/wasi-libc/pull/338, so that can be used to avoid the roudup calculation here.

Also, re avoiding collision with userland `sbrk` calls, I still believe making `try_init_allocator` a ctor is a simpler solution.

sbc100:
So is `__heap_end` the solution here?  Maybe can use a weak reference to it and call fall to the current implementation when it is missing (i.e. on older llvm versions)?

alexcrichton:
Yes I've updated this to prioritize using `__heap_end` through a weak symbol now.

Note @TerrorJack that using a ctor is also not going to work here since ctors are not run as part of the wasm `start` function (or at least not that I know of) which means that the initialization here would still run too late and embedders could, for example, grow memory before calling exports of the module.

TerrorJack:
> ctors are not run as part of the wasm start function

IIUIC `wasm-ld` takes care of ctors at link time and puts them in the `_initialize`/`_start` function per wasi reactor/command spec?

dlmalloc/src/malloc.c
```diff
@@ -5214,16 +5214,33 @@ static void internal_inspect_all(mstate m,
...
+  // for users to craft larger linear memories by passing options to extend
+  // beyond this threshold. In this situation the memory will not be used for
+  // dlmalloc.
+  //
+  // Note that `sbrk(0)`, or in dlmalloc-ese `CALL_MORECORE(0)`, is specifically
+  // not used here. That captures the current size of the heap but is only
+  // correct if the we're the first to try to grow the heap. If the heap has
+  // grown elsewhere, such as a different allocator in place, then this would
+  // incorrectly claim such memroy as our own.
+  char *init = &__heap_end;
```
sbc100, line 5241: How about calling this `end` to match the `__heap_end` name?

alexcrichton:
Yes that is distinct from the `start` function. When the wasm module is instantiated the `_start` or `_initialize` function isn't run, it's only when explicitly requested by an embedder that it's executed. In the use case that discovered this issue it was before the `_start` function was run that memory was grown, so a ctor would not solve this issue.

dlmalloc/src/malloc.c
```diff
@@ -5214,16 +5214,33 @@ static void internal_inspect_all(mstate m,
...
+  if (init == NULL)
+    init = (char*) (((size_t) base + PAGESIZE - 1) & ~(PAGESIZE - 1));
```
sbc100, line 5243: Looks like we already have `page_align` defined in this file.  Does that work here?

dlmalloc/src/malloc.c
```diff
@@ -5214,16 +5214,33 @@ static void internal_inspect_all(mstate m,
...
   int initial_heap_size = init - base;
```
sbc100, line 5244: Unrelated to this change but shouldn't the type of initial_heap_size be `size_t`?

TerrorJack:
That's fair. Maybe a bit offtopic but I do wonder why the ctors aren't put into the module `start` function instead.

> In the use case that discovered this issue it was before the _start function was run that memory was grown

Sounds like undefined behavior to call anything before `_start`/`_initialize` in the wasi land. But it's also reasonable to try working around this (perhaps common) user mistake, so in that case indeed ctor isn't a good solution.

alexcrichton:
Ah excellent suggestions @sbc100, updated to include those.

> Sounds like undefined behavior to call anything before _start/_initialize in the wasi land.

Personally I don't consider that a great stance for wasi-libc to take. That's a pretty onerous restriction to work around when relatively simple lazy initialization, such as what happens in dlmalloc right now, fixes the issue.

sbc100:
> That's fair. Maybe a bit offtopic but I do wonder why the ctors aren't put into the module `start` function instead.

There is a long discussion on this here: https://github.com/WebAssembly/design/issues/1160

TLDR: Calling host functions normally requires exports to have been received by the host (e.g. the host can do much until is has a handle to the memory export).   Exports are not received by the host until after the start function runs, therefore its not possible, in the general case to call host functions during start function.

sbc100:
> > That's fair. Maybe a bit offtopic but I do wonder why the ctors aren't put into the module `start` function instead.
> 
> There is a long discussion on this here: [WebAssembly/design#1160](https://github.com/WebAssembly/design/issues/1160)
> 
> TLDR: Calling host functions normally requires exports to have been received by the host (e.g. the host can do much until is has a handle to the memory export). Exports are not received by the host until after the start function runs, therefore its not possible, in the general case, to call host functions during start function.

For what its worth the wasm-ld linker will use the start function for some things if it knows that code is only doing internal stuff (i.e. loading memory segments, applying relocations).   its only if user code is involved that we must wait until after start has run.

sunfishcode:
The approach of using `__heap_end` as a weak symbol, and falling back to the page alignment logic looks good to me. `__heap_end` was added relatively recently so it isn't in LLVM 15, but when it is available, it's exactly what we need here.

sunfishcode:
> Shouldn't the same thing be done for [`emmalloc`](https://github.com/WebAssembly/wasi-libc/blob/a1c7c2c7a4b2813c6f67bd2ef6e0f430d31cebad/emmalloc/emmalloc.c#L548)?

Good point; I've submitted #378 for this.

yamt:
> This doesn't break that use case, but it does make it less efficient.

i feel you have a different definition of `break` from mine.
dropping a few pages is critical for some of my environment.

> The default output of LLD doesn't ever have more than one page at the end of memory (or at least not that I'm aware of), and I'm not sure if such a setup is commonly used in the wild. If it is then I think it would perhaps make sense to add a new pseudo-symbol to LLD such as `__heap_base` along the lines of `__heap_initial_end` or something like that, and then that can be used instead of the round-up-to-page-size calculation done here.

while it might not be too common setup, i occasionally use such a setup.
also, i suspect the use of `memory.grow` out of malloc isn't common either.
maybe you can explain your use case a bit more?

sunfishcode:
LLVM has added a `__heap_end` symbol which is the eventual path forward here. It unfortunately didn't make it into LLVM 15, but it will be in future versions. The patch is https://reviews.llvm.org/D136110.

Does wasm-ld have some option which enables extra patches to be allocated after `__heap_base`? If not, do you use a patched wasm-ld? If so, if you also add the patch from https://reviews.llvm.org/D136110, then this PR will work for you, because it'll use `__heap_end`.

yamt:
> LLVM has added a `__heap_end` symbol which is the eventual path forward here. It unfortunately didn't make it into LLVM 15, but it will be in future versions. The patch is https://reviews.llvm.org/D136110.
> 
> Does wasm-ld have some option which enables extra patches to be allocated after `__heap_base`? If not, do you use a patched wasm-ld? If so, if you also add the patch from https://reviews.llvm.org/D136110, then this PR will work for you, because it'll use `__heap_end`.

i'm aware of the situation about `__heap_end`. i have no problem with that part of this PR.

what i'm not very happy with is the "if __heap_end is not available" part of this PR.

i use `--init-memory` when i want to use an initial heap larger then default.

sbc100:
> > LLVM has added a `__heap_end` symbol which is the eventual path forward here. It unfortunately didn't make it into LLVM 15, but it will be in future versions. The patch is https://reviews.llvm.org/D136110.
> > Does wasm-ld have some option which enables extra patches to be allocated after `__heap_base`? If not, do you use a patched wasm-ld? If so, if you also add the patch from https://reviews.llvm.org/D136110, then this PR will work for you, because it'll use `__heap_end`.
> 
> i'm aware of the situation about `__heap_end`. i have no problem with that part of this PR.
> 
> what i'm not very happy with is the "if __heap_end is not available" part of this PR.

I tend to agree with this.  I think in the absence of __heap_end it seems reasonable to just assume that the heap goes from `__heap_base` to the end of the linear memory.    Indeed, this is kind of what wasm-ld and emscripten have assumed for a while now.

A very common use case (at least for emscripten) is to have a fixed size memory (non--growable) which specified via e.g.  `-Wl,--initial-memory=256mb`.   This patch seems to completely break that use case for LLVM < 15.

IIUC @alexcrichton has a use case where he wants to be able to grow the memory (over-allocate the initial memory), prior to instantiation and then have malloc skip over/ignore that chunk of memory.   Is that about right, alex?  

Given that with LLVM < 15 we can't make both use cases work, I think we need to agree that one of them will require LLVM 15 or above..   @alexcrichton's use case seems like the more exotic address space layout (and possibly incompatible with allocatators that assume a single contiguous malloc region)  so it would make sense to me to require LLVM 15 for that use case.   @alexcrichton does that seem reasonable to you? 

Another argument for keeping the status quo for LLVM < 15 is that all previous wasi-sdk releases have had this behaviour.

Perhaps that argument is academic because all future wasi-sdk versions will be built with LLVM >= 15 anyway?

> 
> i use `--init-memory` when i want to use an initial heap larger then default.

alterstep:
sorry -- https://reviews.llvm.org/D136110 => LLVM 15 or LLVM 16?

yamt:
> sorry -- https://reviews.llvm.org/D136110 => LLVM 15 or LLVM 16?

as it isn't included in `llvmorg-16-init`, i suspect it's even LLVM 17.

sunfishcode:
> allocatators that assume a single contiguous malloc region

This hasn't been discussed anywhere, but I don't think we should want to consider such allocators to be supported. Users and tools can call `sbrk` or even memory.grow themselves, both dlmalloc and emmalloc are ok with this, and have been since the beginning. It's a common way to allocate memory without interfering with the malloc heap.

> A very common use case (at least for emscripten) is to have a fixed size memory (non--growable) which specified via e.g. -Wl,--initial-memory=256mb. This patch seems to completely break that use case for LLVM < 15.

I was previously unaware people were using --initial-memory this way. I'm curious what the motivation for this is. What's the advantage of creating a larger initial memory versus just using memory.grow to grow the memory to the desired size? Is there a significant performance difference?

sunfishcode:
> as it isn't included in `llvmorg-16-init`, i suspect it's even LLVM 17.

llvmorg-16-init isn't the release branch; releases/16.x will be the release branch, which is expected to be created in a few weeks. I expect `__heap_end` will be in LLVM 16. I've now also requested a backport to the LLVM 15 (releases/15.x) branch.

`memory.grow` is an MVP instruction, and in default configurations, wasm-ld does default to creating growable memory. That said, if people are using --initial-memory in this way, this is a regression in any case. If we can get the patch backported to LLVM 15, it's tempting to fix this by just releasing a new wasi-sdk with the updated LLVM 15.

sbc100:
> `memory.grow` is an MVP instruction, and in default configurations, wasm-ld does default to creating growable memory. 

Ah my bad.  I misremembered that.   

I'm pretty sure that unless you specify a `--max-memory` wasm-ld will not create a growable memory.  The code for it is here https://github.com/llvm/llvm-project/blob/b5cb91fc1cb851dcc4d3dc82a19a9d56dd352350/lld/wasm/Writer.cpp#L386-L400.

Here the maxMemory config option defaults to zero, which means no max is specified which means that memory is not growable (I think?.. unless I'm misremembering that too :)

sunfishcode:
With a wasi-sdk clang and wasm-ld and no special args, this:
```c
#include <stdio.h>

int main() {
	unsigned long a = __builtin_wasm_memory_size(0);
	unsigned long b = __builtin_wasm_memory_grow(0, 1);
	unsigned long c = __builtin_wasm_memory_size(0);
	unsigned long d = __builtin_wasm_memory_grow(0, 1);
	unsigned long e = __builtin_wasm_memory_size(0);
	printf("%lu %lu %lu %lu %lu\n", a, b, c, d, e);
}
```
prints "2 2 3 3 4".

And it's good that it does, because malloc depends on `memory.grow` once its initial heap runs out, and we default to not giving it any extra pages in its initial heap :smile: .

sbc100:
I see, by default we get a memory that doesn't have a max at all.. I though that meant it was not growable, but I guess it means its has no upper limit(?).   Sorry for the confusion.

For the record, it looks like in emscripten sets both `--max-memory` and `--initial-memory` by default to the same value so its not affected by this "default-to-growable" behaviour.

yamt:
> I was previously unaware people were using --initial-memory this way. I'm curious what the motivation for this is. What's the advantage of creating a larger initial memory versus just using memory.grow to grow the memory to the desired size? Is there a significant performance difference?

realloc-based memory.grow implementation has non negligible chances to fail on resource-constraint environment.

yamt:
> > as it isn't included in `llvmorg-16-init`, i suspect it's even LLVM 17.
> 
> llvmorg-16-init isn't the release branch; releases/16.x will be the release branch, which is expected to be created in a few weeks. I expect `__heap_end` will be in LLVM 16. I've now also requested a backport to the LLVM 15 (releases/15.x) branch.

ok. i was assuming it was a branch point. thank you for correction.

sunfishcode:
The `__heap_end` patch has now landed on the [releases/15.x branch](https://github.com/llvm/llvm-project/tree/release/15.x). I'll do a wasi-sdk-19 release soon based on this.

I've added a note about this situation to the wasi-sdk-18 release notes, and I'll mention it in the wasi-sdk-19 release notes too.

yamt:
> The `__heap_end` patch has now landed on the [releases/15.x branch](https://github.com/llvm/llvm-project/tree/release/15.x). I'll do a wasi-sdk-19 release soon based on this.
> 
> I've added a note about this situation to the wasi-sdk-18 release notes, and I'll mention it in the wasi-sdk-19 release notes too.

thank you.

@alexcrichton can the controversial part of this PR be reverted with it?

alexcrichton:
Personally I don't think that the round-up-to-page-size-if-`__heap_end`-is-missing part of this PR should be reverted. Without this logic it's not possible to do any memory growth before malloc is first called. This differs from the fact, however, that malloc does correctly handle memory growth after it is called the first time. 

I personally feel that `wasi-libc` should appeal to a broad range of use cases and strive to maximally flexible, in this case not limiting what can be done with it, for example preventing use of `memory.grow` before the first call to `malloc` but then allowing it after the first call to `malloc`. In that sense I would personally like to see this code remain to ensure that wasi-libc remains compatible while the transition is happening waiting for `__heap_end` to make its way out to everyone.

I don't mean to discount the problems that happen when `__heap_end` is not defined. That's a very real problem and I think it's worth solving, but that's what `__heap_end` is for as a "root fix". Otherwise it's only about interim fixes and I would personally err on the side of caution for an interim fix which is to have what's here which is more conservative than aggressively claiming everything from `__heap_base` to the end of memory as malloc-able space.

sbc100:
> Personally I don't think that the round-up-to-page-size-if-`__heap_end`-is-missing part of this PR should be reverted. Without this logic it's not possible to do any memory growth before malloc is first called. This differs from the fact, however, that malloc does correctly handle memory growth after it is called the first time.

But hasn't wasi-libc and wasi-sdk historically never allowed this, and always claimed that memory between __heap_base the end of memory for its initial malloc region.

I think that emscripten binaries assume something similar.  They  assume that anything after `__heap_base` constitutes the sbrk region, and that this region grows contiguously from from `__heap_base`.  (see `system/lib/sbrk.c` in emscripten).

It seems like what you are asking for a new and different behaviour,  which of course is fine, and I agree it should work.  But why can't you gate this new behaviour on existence of `__heap_end`?   Why break the old behaviour?

alexcrichton:
I'll reiterate again that these are just my own personal thoughts, and I'm not really a core maintainer here so my thoughts I don't think should really hold all that much weight for maintaining wasi-libc. I don't think that the old behavior is correct, to me it's broken. To me that's enough reason to preserve the old behavior because I don't think that a bug should be preserved just because it's been there for awhile. 

Looking at the history this behavior was specifically added in https://github.com/WebAssembly/wasi-libc/pull/114 so while it's pretty old at this point it may not be so long as "wasi-libc never allowed this". Regardless though, in my opinion, how old it is I don't think should have a bearing on fixing the underlying bug.

I again don't want to appear like I'm diminishing or brushing away use cases here. My point is that the use case of someone else calling `memory.grow` before `malloc` is fundamentally broken today, and has been since #114 was merged. Fixing that to me, personally at least, takes a higher priority than preserving an alternative use case in the interm while LLVM is updated to specify `__heap_end` everywhere.

Again though I don't maintain wasi-libc, I just contribute on occasion. If this patch is reverted I'll find other ways to solve my problems, so I don't think it necessarily needs my own personal approval.

yamt:
my understanding of the situation is same.

my preference is the opposite though.
in this case, preserving a working use case outweighs fixing a use case which has been broken for long, especially a better fix (`__heap_end`) is on the horizon.

sbc100:
> My point is that the use case of someone else calling `memory.grow` before `malloc` is fundamentally broken today,

This is somewhat of a new use case to me.  I had always imagined that `__heap_base` onwards was considered the malloc heap.   Do you imagine that it should be OK to grab chunks of memory in this way after malloc has been called?  Or is it only OK to do this before malloc is called?

sunfishcode:
[LLVM 15.0.7](https://github.com/llvm/llvm-project/releases/tag/llvmorg-15.0.7) is now released, with the fix.

alexcrichton:
I believe the current status of wasi-libc is that:
 
* If `memory.grow` is called before the first `malloc`, then wasi-libc claims that memory for itself
* If `memory.grow` is called after the first `malloc`, then wasi-libc avoids using that memory

I personally think wasi-libc's malloc should not assume aggressive ownership of the entire address space. There might be multiple allocators in play or other custom embedding logic which uses memory and allocates via `memory.grow`. For example my use case is component-model related where state for a separate module than the "main module" using wasi-libc is allocated via `memory.grow` because the malloc function is not guaranteed to be exported.

sbc100:
Oh, maybe `MORECORE_CONTIGUOUS` does actually handle discontinuities:

```
MORECORE_CONTIGUOUS       default: 1 (true) if HAVE_MORECORE                     
  If true, take advantage of fact that consecutive calls to MORECORE             
  with positive arguments always return contiguous increasing                    
  addresses.  This is true of unix sbrk. It does not hurt too much to            
  set it true anyway, since malloc copes with non-contiguities.                  
  Setting it false when definitely non-contiguous saves time                        
  and possibly wasted space it would take to discover this though.  
```

alexcrichton:
Sorry, I should clarify I'm describing the pre-this-PR behavior since that's also what I think you're talking about.

As for libc's behavior, I determined this empirically:

```rust
use std::alloc;
use std::arch::wasm32::memory_grow;

const PAGESIZE: usize = 64 * 1024;

#[no_mangle]
extern "C" fn entry() {
    // 0x110000 - before first malloc
    println!("{:#x}", memory_grow::<0>(1) * PAGESIZE);

    // 0x102600 - note that this extends into the grown region
    println!("{:p}", allocate(PAGESIZE));

    // 0x120000 - grow after first malloc
    println!("{:#x}", memory_grow::<0>(1) * PAGESIZE);

    // 0x130010 - note that this does not conflict with second `memory.grow`
    println!("{:p}", allocate(PAGESIZE));
}

fn allocate(amt: usize) -> *mut u8 {
    unsafe { alloc::alloc(alloc::Layout::from_size_align(amt, 8).unwrap()) }
}
```

This prints:

```
$ rustc foo.rs --crate-type cdylib --target wasm32-wasi && wasmtime foo.wasm --invoke entry
0x110000
0x102600
0x120000
0x130010
```

Here the return value of `malloc` overlaps with the first `memory.grow`, but is disjoint from the second `memory.grow`.

I also saw `MORECORE_CONTIGUOUS` and I don't know why it doesn't have the expected effect. I did not dig deeply into `dlmalloc`'s implementation.

yamt:
> I believe the current status of wasi-libc is that:
> 
>     * If `memory.grow` is called before the first `malloc`, then wasi-libc claims that memory for itself
> 
>     * If `memory.grow` is called after the first `malloc`, then wasi-libc avoids using that memory
> 
> 
> I personally think wasi-libc's malloc should not assume aggressive ownership of the entire address space. There might be multiple allocators in play or other custom embedding logic which uses memory and allocates via `memory.grow`. For example my use case is component-model related where state for a separate module than the "main module" using wasi-libc is allocated via `memory.grow` because the malloc function is not guaranteed to be exported.

we are all fine with `__heap_end` and the remaining question is what to do without `__heap_end`, right?
i prefer to keep the old behavior for better compatibility.
can't you wait for `__heap_end` for some reasons? why not?

yamt:
i submitted a partial revert https://github.com/WebAssembly/wasi-libc/pull/386

PR #378, opened 2023-01-09
Port the fix from #377 to emmmalloc.

sunfishcode:
Avoid using sbrk(0) in emmalloc too.

### DIFF
--- a/emmalloc/emmalloc.c
+++ b/emmalloc/emmalloc.c
@@ -56,6 +56,7 @@
 
 // Defind by the linker to have the address of the start of the heap.
 extern unsigned char __heap_base;
+extern unsigned char __heap_end __attribute__((__weak__));
 
 // Behavior of right shifting a signed integer is compiler implementation defined.
 static_assert((((int32_t)0x80000000U) >> 31) == -1, "This malloc implementation requires that right-shifting a signed integer produces a sign-extending (arithmetic) shift!");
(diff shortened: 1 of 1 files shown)


jedisct1:
Unfortunately this is causing applications to trap with an out of bounds access.

sunfishcode:
Ah, that's unfortunate. Any help anyone might be able to provide in debugging this would be appreciated!

sbc100:
> Unfortunately this is causing applications to trap with an out of bounds access.

Is that with or without the `__heap_base` symbol?  i.e. which llvm version are you using?

jedisct1:
This is with LLVM 15.0.6.

cuviper:
(irrelevant nit: the commit message and title of this PR have too many 'm's)

emmalloc/emmalloc.c
```diff
@@ -56,6 +56,7 @@
 
 // Defind by the linker to have the address of the start of the heap.
 extern unsigned char __heap_base;
+extern unsigned char __heap_end __attribute__((__weak__));
```
cuviper, line 59: Since #394, this should no longer be weak.

In fact, building with `MALLOC_IMPL=emmalloc` is currently broken due to missing this symbol.

emmalloc/emmalloc.c
```diff
@@ -545,7 +546,10 @@ static bool claim_more_memory(size_t numBytes)
     // If this is the first time we're called, see if we can use
     // the initial heap memory set up by wasm-ld.
     if (!listOfAllRegions) {
-      unsigned char *heap_end = sbrk(0);
+      unsigned char *heap_end = &__heap_end;
+      if (heap_end == NULL)
+        heap_end = (unsigned char*) ((size_t) &__heap_base + (PAGE_SIZE-1) & -PAGE_SIZE);
```
cuviper, line 551: Once non-weak, the `NULL` fallback is not needed, but the dlmalloc code is trapping on `end < base`.

PR #379, opened 2023-01-10
Adjust Makefile for LLVM trunk (16) as of 2023-01-05

glandium:
https://github.com/llvm/llvm-project/commit/d227c3b68cf5c236902c9ff4fdf8b719c9a3dd26 added __GCC_HAVE_SYNC_COMPARE_AND_SWAP macros.


sbc100 (approved):
We should probably go through and remove some of the older filters here (assuming we aren't trying to support very old versions of llvm).

lgtm to this though

glandium:
I guess the main question would be: what would be the new baseline version? (CI tests LLVM 10, for instance, but the hacks go back to at least version 8).
Anyways, could you merge this?

sunfishcode:
We almost certainly don't support building wasi-libc with LLVM 8 or 9 anymore; if someone wanted to remove the hacks, that'd be great!

PR #381, opened 2023-01-11
threads: change `wasm32-wasi-pthread` to `wasm32-wasi-threads`

abrown:
After some thought, I think that we should rename the `THREAD_MODEL=posix` build to avoid confusion. Why? Though in this project the use of this target does involve pthreads, it will not be so in other standard libraries or languages (see, e.g., https://github.com/rust-lang/compiler-team/issues/574). I think it would be preferable to emphasize the "threads" Wasm-level proposal and the "wasi-threads" proposal rather than the specific details of which threading API is being exposed. I also think it would be good to settle on the name now before this gets exposed in wasi-sdk (https://github.com/WebAssembly/wasi-sdk/pull/287).


loganek (approved):
That makes sense to me.

abrown:
No specific guidance like "we reject this name" but you can see from this [comment](https://github.com/rust-lang/compiler-team/issues/574#issuecomment-1379318255) and this [discussion](https://rust-lang.zulipchat.com/#narrow/stream/233931-xxx/topic/New.20tier-2.20target.20for.20.60wasm32-wasi.60.20.2B.20thr.E2.80.A6.20compiler-team.23574) that the name could potentially be considered an issue. With this change I'm trying to avoid any such issues.

abrown:
@sbc100, @sunfishcode: any objections?

abrown:
cc: @penzn, @alexcrichton, @yamt

sbc100 (approved):

sunfishcode (approved):

PR #383, opened 2023-01-11
threads: enable more of `pthread_mutexattr_setprotocol`

abrown:
While using wasi-threads with wasi-libc, I found a use of this branch of the code. Adding in the WASI version of `futex_wait` seems to fix things.


yamt (changes requested):
it doesn't make sense as wasm doesn't have priority or its inheritance.

yamt:
when i use PI, i need it actually work.
i feel it's more user friendly to raise an error than silently ignoring it.

besides that, this implementation looks inappropriate even as a dummy implementation:

* __wasilibc_futex_wait doesn't implement FUTEX_LOCK_PI
* i believe the original FUTEX_LOCK_PI call here is to detect the existence of kernel functionality

abrown:
Sure, let's close this.

PR #387, opened 2023-01-20
Rename thread_spawn import

yamt:
Following the wit-defined ABI:
https://github.com/WebAssembly/wasi-threads/pull/26

### DIFF
--- a/libc-bottom-half/sources/__wasilibc_real.c
+++ b/libc-bottom-half/sources/__wasilibc_real.c
@@ -662,7 +662,7 @@ __wasi_errno_t __wasi_sock_shutdown(
 #ifdef _REENTRANT
 int32_t __imported_wasi_thread_spawn(int32_t arg0) __attribute__((
     __import_module__("wasi"),
-    __import_name__("thread_spawn")
+    __import_name__("thread-spawn")
 ));
 
 int32_t __wasi_thread_spawn(void* start_arg) {


abrown:
I don't think this is right; IIRC, WIT dashes are converted to underscores.

yamt:
> I don't think this is right; IIRC, WIT dashes are converted to underscores.

wit-bindgen seems to convert C identifiers. but not import names.

sbc100:
> > I don't think this is right; IIRC, WIT dashes are converted to underscores.
> 
> wit-bindgen seems to convert C identifiers. but not import names.

If that true, I hope its not to late to revisit.  I would hope we could choose a convention for import/export names that aligns the symbol names normally found in programming languages.   As programmers, we not used to seeing hyphens in the middle of identifiers.

yamt:
> > > I don't think this is right; IIRC, WIT dashes are converted to underscores.
> > 
> > 
> > wit-bindgen seems to convert C identifiers. but not import names.
> 
> If thats true, I hope its not to late to revisit. I would hope we could choose a convention for import/export names that aligns the symbol names normally found in programming languages. As programmers, we not used to seeing hyphens in the middle of identifiers.

as preview2 heavily uses names with hyphens, i thought it was the latest trend.
eg.
```
    (import "wasi-filesystem" "append-via-stream" (func $_ZN22wasi_snapshot_preview18bindings15wasi_filesystem17append_via_stream10wit_import17hd102ec87cb8cdfd2E (;2;) (type 2)))
```

wrt this specific change, i even have confirmed the intention explicitly: https://github.com/WebAssembly/wasi-threads/pull/26#discussion_r1067595779
at least it was @loganek 's intention to use hyphen.

abrown:
> wit-bindgen seems to convert C identifiers. but not import names.

Yeah, sorry to confuse things; I was mistaken. Indeed, like @yamt says, the import names will have hyphens. But the C generated from WIT would have the expected underscores. Since this is what users would typically interact with, I think it is fine for the lower level WIT names to use the "almost a convention" hyphens that WIT has transitioned to.

yamt:
> > > wit-bindgen seems to convert C identifiers. but not import names.
> > 
> > 
> > Yeah, sorry to confuse things; I was mistaken. Indeed, like @yamt says, the import names will have hyphens. But the C generated from WIT would have the expected underscores. Since this is what users would typically interact with, I think it is fine for the lower level WIT names to use the "almost a convention" hyphens that WIT has transitioned to.
> 
> I guess I missed that transition. Maybe it happened a repo I'm not tracking.

honestly speaking, i feel the transition a bit awkward and am curious about rationales.
but i guess it isn't something specific to this PR.

abrown:
> i guess it isn't something specific to this PR

Yeah, we should bring up the name convention change in the next WASI meeting. In the meantime, should I merge this?

sbc100:
> > i guess it isn't something specific to this PR
> 
> Yeah, we should bring up the name convention change in the next WASI meeting. In the meantime, should I merge this?

sgtm

abrown (approved):

PR #388, opened 2023-01-25
Add a check to reactor modules to ensure _initialize is only called once

TerrorJack:
Calling _initialize multiple times is undefined behavior, since the ctors are not guaranteed to be idempotent. We should have this safety check which is similar to #329.

### DIFF
--- a/libc-bottom-half/crt/crt1-reactor.c
+++ b/libc-bottom-half/crt/crt1-reactor.c
@@ -1,7 +1,27 @@
+#if defined(_REENTRANT)
+#include <stdatomic.h>
+extern void __wasi_init_tp(void);
+#endif
 extern void __wasm_call_ctors(void);
 
 __attribute__((export_name("_initialize")))
 void _initialize(void) {
+#if defined(_REENTRANT)
+    static volatile atomic_int initialized = 0;
+    int expected = 0;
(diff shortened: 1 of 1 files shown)


libc-bottom-half/crt/crt1-reactor.c
```diff
@@ -2,6 +2,12 @@ extern void __wasm_call_ctors(void);
 
 __attribute__((export_name("_initialize")))
 void _initialize(void) {
+    static volatile int started = 0;
```
sbc100, line 5: Call this `initialized` maybe?

TerrorJack: Sure.

sbc100 (commented):
Seems reasonable to me.  I'll wait @sunfishcode for chime in too though.

TerrorJack:
@sunfishcode 👀

sunfishcode:
Sounds like a good idea to me. We should also add code to do this with atomics when `ifdef REENTRANT`, similar to what crt-command.c now does.

TerrorJack:
@sunfishcode The check logic when `_REENTRANT` has been implemented.

sunfishcode:
Thanks!

PR #394, opened 2023-02-17
dlmalloc: require __heap_end

yamt:
This commit effectively drops the support of older wasm-ld. (LLVM <15.0.7)

We have two relevant use cases:

* `memory.grow` use outside of malloc (eg. used by polyfill preview1 binaries)

* `--init-memory` to somehow preallocate heap (eg. avoid dynamic allocations, especially on small environments)

While https://github.com/WebAssembly/wasi-libc/pull/377 fixed the former, it broke the latter if you are using an older LLVM, which doesn't provide the `__heap_end` symbol, to link your module.

As we couldn't come up with a solution which satisfies all parties, this commit simply makes it require new enough LLVM which provides `__heap_end`. After all, a link-time failure is more friendly to users than failing later in a subtle way.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,8 +8,26 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        # oldest and newest supported LLVM version
-        clang_version: [10.0.0, 14.0.0]
+        clang_version: [10.0.0]
+        # use different LLVM versions among oses because of the lack of
+        # official assets on github.
+        include:
+          - os: ubuntu-latest
+            clang_version: 10.0.0
+            llvm_asset_suffix: x86_64-linux-gnu-ubuntu-18.04
+          - os: macos-latest
+            clang_version: 10.0.0
+            llvm_asset_suffix: x86_64-apple-darwin
+          - os: windows-latest
+            clang_version: 10.0.0
+          - os: ubuntu-latest
+            clang_version: 16.0.0
+            llvm_asset_suffix: x86_64-linux-gnu-ubuntu-18.04
+          - os: macos-latest
+            clang_version: 15.0.7
+            llvm_asset_suffix: x86_64-apple-darwin21.0
+          - os: windows-latest
+            clang_version: 16.0.0
     steps:
     - uses: actions/checkout@v1
       with:
@@ -44,8 +62,8 @@ jobs:
     - name: Install LLVM tools (MacOS)
       shell: bash
       run: |
-        curl -sSfL https://github.com/llvm/llvm-project/releases/download/llvmorg-${{ matrix.clang_version }}/clang+llvm-${{ matrix.clang_version }}-x86_64-apple-darwin.tar.xz | tar xJf -
-        export CLANG_DIR=`pwd`/clang+llvm-${{ matrix.clang_version }}-x86_64-apple-darwin/bin
+        curl -sSfL https://github.com/llvm/llvm-project/releases/download/llvmorg-${{ matrix.clang_version }}/clang+llvm-${{ matrix.clang_version }}-${{ matrix.llvm_asset_suffix }}.tar.xz | tar xJf -
+        export CLANG_DIR=`pwd`/clang+llvm-${{ matrix.clang_version }}-${{ matrix.llvm_asset_suffix }}/bin
         echo "$CLANG_DIR" >> $GITHUB_PATH
         echo "CC=$CLANG_DIR/clang" >> $GITHUB_ENV
         echo "AR=$CLANG_DIR/llvm-ar" >> $GITHUB_ENV
@@ -55,8 +73,8 @@ jobs:
     - name: Install LLVM tools (Linux)
       shell: bash
       run: |
-        curl -sSfL https://github.com/llvm/llvm-project/releases/download/llvmorg-${{ matrix.clang_version }}/clang+llvm-${{ matrix.clang_version }}-x86_64-linux-gnu-ubuntu-18.04.tar.xz | tar xJf -
-        export CLANG_DIR=`pwd`/clang+llvm-${{ matrix.clang_version }}-x86_64-linux-gnu-ubuntu-18.04/bin
+        curl -sSfL https://github.com/llvm/llvm-project/releases/download/llvmorg-${{ matrix.clang_version }}/clang+llvm-${{ matrix.clang_version }}-${{ matrix.llvm_asset_suffix }}.tar.xz | tar xJf -
+        export CLANG_DIR=`pwd`/clang+llvm-${{ matrix.clang_version }}-${{ matrix.llvm_asset_suffix }}/bin
         echo "$CLANG_DIR" >> $GITHUB_PATH
         echo "CLANG_DIR=$CLANG_DIR" >> $GITHUB_ENV
         echo "CC=$CLANG_DIR/clang" >> $GITHUB_ENV
@@ -82,6 +100,8 @@ jobs:
         make test
       # The older version of Clang does not provide the expected symbol for the
       # test entrypoints: `undefined symbol: __main_argc_argv`.
+      # The older (<15.0.7) version of wasm-ld does not provide `__heap_end`,
+      # which is required by our malloc implementation.
       if: matrix.os == 'ubuntu-latest' && matrix.clang_version != '10.0.0'
 
     - uses: actions/upload-artifact@v1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -94,7 +94,7 @@ jobs:
       run: |
         cd test
         make download
-        export WASI_DIR=$(realpath $CLANG_DIR/../lib/clang/${{ matrix.clang_version }}/lib/wasi/)
+        export WASI_DIR=$(realpath $($CLANG_DIR/clang -print-resource-dir)/lib/wasi/)
         mkdir -p $WASI_DIR
         cp download/lib/wasi/libclang_rt.builtins-wasm32.a $WASI_DIR
         make test
--- a/dlmalloc/src/malloc.c
+++ b/dlmalloc/src/malloc.c
@@ -5233,6 +5233,8 @@ static void try_init_allocator(void) {
     // This can happen when 1. you are using an old wasm-ld which doesn't
     // provide `__heap_end` and 2. something (other libraries or maybe
     // your app?) provide a weak reference to `__heap_end`.
+    //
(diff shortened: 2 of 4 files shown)


dlmalloc/src/malloc.c
```diff
@@ -5227,23 +5227,14 @@ static void try_init_allocator(void) {
...
-  // not used here. That captures the current size of the heap but is only
-  // correct if the we're the first to try to grow the heap. If the heap has
-  // grown elsewhere, such as a different allocator in place, then this would
-  // incorrectly claim such memroy as our own.
+  // the heap.
   char *end = &__heap_end;
-  if (end == NULL)
-    end = (char*) page_align((size_t) base);
+  if (end == NULL) {
+    // This can happen when 1. you are using an old wasm-ld which doesn't
```
sbc100, line 5233: Wouldn't this just fail with `undefined symbol: __heap_end` on older versions of llvm?

If `foo` is not weak I'm pretty sure `&foo` can never be null.

.github/workflows/main.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         # oldest and newest supported LLVM version
-        clang_version: [10.0.0, 14.0.0]
+        clang_version: [10.0.0, 15.0.7]
```
sbc100, line 12: Wouldn't we want to update the min version here?  I think the idea is to test with the oldest supported version and also with a current/newer version.

yamt: even before this PR, we were not testing app link for 10.0.0.

dlmalloc/src/malloc.c
yamt: usually yes.
but see the comment in the code.

sbc100: But I don't see why `&__heap_end` could ever be null here.. that would be linker bug I think.    

I'm not sure why you mean by "something (other libraries or maybe provide a weak reference to `__heap_end`" ... do you mean that this symbol is weakly defined elsewhere?  If so then it should have a non-zero address still.

sbc100: In other words I don't think its possible for anyone to ever provide a symbol which has a null address.    

The only way its address can be null is if the reference here in this file is weak.

yamt:
i couldn't find any llvm >=15 releases which have assets for both of x86-64 ubuntu and x86-64 darwin.
what to do?

dlmalloc/src/malloc.c
yamt: > But I don't see why `&__heap_end` could ever be null here.. that would be linker bug I think.
> 
> I'm not sure why you mean by "something (other libraries or maybe provide a weak reference to `__heap_end`" ... do you mean that this symbol is weakly defined elsewhere?

yes.

eg. have `extern __attribute__((weak)) unsigned char __heap_end` in your app.

> If so then it should have a non-zero address still.

i observed NULL address when i experimented.
i don't know if it's a linker bug or not.

sbc100:
> i couldn't find any llvm >=15 releases which have assets for both of x86-64 ubuntu and x86-64 darwin. what to do?


15.0.0 looks like it has those binaries: https://github.com/llvm/llvm-project/releases/tag/llvmorg-15.0.0

And 15.0.2: https://github.com/llvm/llvm-project/releases/tag/llvmorg-15.0.2

Strange that some of the others don't those.  Might be worth opening an issue on the llvm tracker?

dlmalloc/src/malloc.c
sbc100: If having a second, weak, reference to the same symbol causes the strong reference to get a zero address that would certainly be a linker bug.  The behaviour should be that if the program contains any strong references to a missing symbol then the link should fail.

yamt:
> > i couldn't find any llvm >=15 releases which have assets for both of x86-64 ubuntu and x86-64 darwin. what to do?
> 
> 15.0.0 looks like it has those binaries: https://github.com/llvm/llvm-project/releases/tag/llvmorg-15.0.0
> 
> And 15.0.2: https://github.com/llvm/llvm-project/releases/tag/llvmorg-15.0.2

which assets are you talking about? can you give me the exact urls of assets?
i only see powerpc for ubuntu.

> 
> Strange that some of the others don't those. Might be worth opening an issue on the llvm tracker?

while i'm not familiar with llvm release process, i guess it's like volunteers building and uploading each binaries.
eg. https://discourse.llvm.org/t/llvm-15-0-5-has-been-tagged/66615

sbc100:
> > > i couldn't find any llvm >=15 releases which have assets for both of x86-64 ubuntu and x86-64 darwin. what to do?
> > 
> > 
> > 15.0.0 looks like it has those binaries: https://github.com/llvm/llvm-project/releases/tag/llvmorg-15.0.0

https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.0/clang+llvm-15.0.0-x86_64-apple-darwin.tar.xz
https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.0/clang+llvm-15.0.0-x86_64-linux-gnu-rhel-8.4.tar.xz

dlmalloc/src/malloc.c
yamt: ok.
i wrote a small demo. https://github.com/yamt/garbage/tree/master/c/weak

yamt: and filed a llvm issue: https://github.com/llvm/llvm-project/issues/60829

yamt:
> > > > i couldn't find any llvm >=15 releases which have assets for both of x86-64 ubuntu and x86-64 darwin. what to do?
> > > 
> > > 
> > > 15.0.0 looks like it has those binaries: https://github.com/llvm/llvm-project/releases/tag/llvmorg-15.0.0
> 
> https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.0/clang+llvm-15.0.0-x86_64-apple-darwin.tar.xz https://github.com/llvm/llvm-project/releases/download/llvmorg-15.0.0/clang+llvm-15.0.0-x86_64-linux-gnu-rhel-8.4.tar.xz

the rhel assert worked on ubuntu.
but it reminded me we actually need >=15.0.7.

yamt:
a. wait for llvm 16 and hope it will have necessary assets
b. use clang from wasi-sdk. feeling a bit chicken-and-egg
c. bulid clang on our ci

TerrorJack:
d. use https://apt.llvm.org/ on linux

yamt:
as a compromise, i made the ci to use different versions of llvm among platforms.

yamt:
i still want to merge this. any concerns?

dlmalloc/src/malloc.c
```diff
@@ -5227,23 +5227,16 @@ static void try_init_allocator(void) {
...
+    // provide `__heap_end` and 2. something (other libraries or maybe
+    // your app?) provide a weak reference to `__heap_end`.
+    //
+    // Note: This is a linker bug: https://github.com/llvm/llvm-project/issues/60829
+    return;
+  }
```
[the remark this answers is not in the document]
yamt: > its really inconceivable to me that somebody would provide a weak definition of this (internal) symbol

i actually have seen such a code. dump these values for diagnostic purposes.

> If you want to keep some kind of check then perhaps just assert(end); with note referring to the bug.

it's rare for users to use wasi-libc w/o NDEBUG.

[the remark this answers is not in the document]
sbc100: > > its really inconceivable to me that somebody would provide a weak definition of this (internal) symbol
> 
> i actually have seen such a code. dump these values for diagnostic purposes.

But I don't think we want to support that configuration, right?    Why would somebody weakly define a linker-internal symbol?  Can you point to the place where you have seen this?    Maybe just `if (!end) abort()` in that case, with a TODO to remove once we can depend on the llvm that contains that fix?

dlmalloc/src/malloc.c
```diff
@@ -5227,23 +5227,16 @@ static void try_init_allocator(void) {
...
-  // not used here. That captures the current size of the heap but is only
-  // correct if the we're the first to try to grow the heap. If the heap has
-  // grown elsewhere, such as a different allocator in place, then this would
-  // incorrectly claim such memroy as our own.
+  // the heap.
   char *end = &__heap_end;
-  if (end == NULL)
-    end = (char*) page_align((size_t) base);
+  if (end == NULL) {
+    // This can happen when 1. you are using an old wasm-ld which doesn't
```
sbc100, line 5233: This (1) case is not true anymore I think.  Older versions of llvm will simply cause link failure, not result in a zero-addressed version of `__heap_end`.

dlmalloc/src/malloc.c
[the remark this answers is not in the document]
yamt: > > > its really inconceivable to me that somebody would provide a weak definition of this (internal) symbol
> > 
> > 
> > i actually have seen such a code. dump these values for diagnostic purposes.
> 
> But I don't think we want to support that configuration, right? Why would somebody weakly define a linker-internal symbol?

diagnostic purposes.
it just dumps something potentially useful for later trouble shooting.

> Can you point to the place where you have seen this?

not open source.
but it's something similar to https://github.com/yamt/garbage/blob/03f9a1a1ee142194905818380f7f0e34287f9c64/wasm/hello/hello.c#L48

> Maybe just `if (!end) abort()` in that case, with a TODO to remove once we can depend on the llvm that contains that fix?

i feel it's too early to assume https://reviews.llvm.org/D144747
given that even the latest wasi-sdk doesn't have the fix yet.

dlmalloc/src/malloc.c
sbc100: I don't think we need to worry that much about it, since it will only effect people that are both (a) using the old linker and (b) are weakly referencing `__heap_end` elsewhere in their program.   I think this is highly unlikely isn't it?   

Maybe at least leave a TODO to completely remove this check once we drop support for older wasm-ld?

sbc100: Just to be clear, is https://reviews.llvm.org/D144747 actually required in order to guarantee that this symbol will be non-NULL?  Or is it simply enough to have 15.0.7?     IIUC if you have 15.0.7 then this can never be null, so this check is only in support of < 15.0.7?

yamt:
@sbc100 after thinking a bit, i changed my mind and turned it to a trap. the whole purpose of this PR is to force users to use newer linker. in that sense, it's better to make it fail loudly even if it's on runtime.

sbc100:
Is this change good to land now?

yamt:
> Is this change good to land now?

yes.
i'm not aware of any open issues.

PR #395, opened 2023-02-17
fix: disable bulk memory feature when `BULK_MEMORY_THRESHOLD` is `UINT32_MAX`

HerrCai0907:
I want to run c++ code in some platform which does not support bulk memory instruction.

If env `BULK_MEMORY_THRESHOLD` is `UINT32_MAX` which means we don't want to use bulk memory instruction. So maybe it is reasonable to disable this feature in this situation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -503,10 +503,12 @@ $(MUSL_PRINTSCAN_NO_FLOATING_POINT_OBJS): CFLAGS += \
 	    -D__wasilibc_printscan_no_floating_point \
 	    -D__wasilibc_printscan_floating_point_support_option="\"remove -lc-printscan-no-floating-point from the link command\""
 
+ifneq (${BULK_MEMORY_THRESHOLD},UINT32_MAX)
 # TODO: apply -mbulk-memory globally, once
 # https://github.com/llvm/llvm-project/issues/52618 is resolved
 $(BULK_MEMORY_OBJS): CFLAGS += \
         -mbulk-memory
+endif
 
 $(BULK_MEMORY_OBJS): CFLAGS += \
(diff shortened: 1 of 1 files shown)


HerrCai0907:
@sbc100 Could you kindly review this PR?

sbc100:
I'm not sure that setting BULK_MEMORY_THRESHOLD to UINT32_MAX is the best way to completely opt out bulk memory support/requirement.

After a quick review I don't see discussion of bulk memory dependency in #263.  @sunfishcode do you remember why we didn't add an opt out for this initially?

sunfishcode:
We didn't add it at the time because bulk-memory is in all the major Wasm engines. It can make sense to add options when people have specific needs, but it's a balance because every `ifdef` we add makes it a little harder to evolve the code to meet new needs.

dkegel-fastly:
I suspect "wasm MVP" is https://www.wasm.com.cn, and currently lacks bulk memory support; perhaps a chinese speaker can confirm.

radkomih:
My use case is such that it targets Wasm MVP (which does not include bulk memory operations).
I need to be able to execute Wasm in a host environment that does not support bulk memory operations, so +1 for being able to opt-out from memory bulk ops.

jedisct1:
@radkomih If you're using `zig cc`, the wasi libc gets compiled with the same options as the rest of the application.

The default target is compatible with Wasm MVP.

Additional optimizations for specific runtimes can be added with `-mcpu=...`, for example:

```sh
zig cc -s -Os -target wasm32-wasi -mcpu=generic+simd128+bulk_memory app.c
```

radkomih:
Thanks, but unfortunately, I can't take advantage of this option since I am not using Zig.

dkegel-fastly:
should still be able to use the same option change in whatever you use to build wasi-libc.

PR #396, opened 2023-02-21
setup_default_stack_size: set __default_stacksize unconditionally

yamt:
If a user specifies a small stack size for the main, maybe it's reasonable to use the same size for threads as well.


libc-top-half/musl/src/env/__init_tls.c
```diff
@@ -47,10 +47,9 @@ static inline void setup_default_stack_size()
 		stack_size = sp > &__global_base ? &__heap_base - &__data_end : (ptrdiff_t)&__global_base;
 	}
 
-	if (stack_size > __default_stacksize)
-		__default_stacksize =
-			stack_size < DEFAULT_STACK_MAX ?
-			stack_size : DEFAULT_STACK_MAX;
+    __default_stacksize =
+        stack_size < DEFAULT_STACK_MAX ?
+        stack_size : DEFAULT_STACK_MAX;
```
abrown, line 52: I think this change makes sense but should we bracket it with `#ifdef __wasilibc_unmodified_upstream`?

loganek: this whole thing is already guarded by `__wasilibc_unmodified_upstream` (line 18). Just a minor note - you use spaces, whereas the whole project uses tabs for indents, so better to keep it consistent (we might want to setup linters to do those checks automatically though).

yamt: i fixed space vs tab issue

loganek (approved):

abrown (approved):

PR #397, opened 2023-02-23
threads: enable `PTHREAD_{MUTEX,RWLOCK,COND}_INITIALIZER`

whitequark:
It looks like `defined(_REENTRANT)` was added [here](https://github.com/WebAssembly/wasi-libc/commit/e5f14be38362f1ab83302895a6e74b2ffd0e2302#diff-5fd85e51ddf267812e2334651dea03b419c9d7a981d33e0d43a9f72e64bb96db) and it was just to mark these as modifiied. (@sunfishcode would know for sure.)

With this change I can build libcxxabi, libcxx, and boost with threads enabled, and link a few binaries (that do not spawn threads, but use mutexes).

### DIFF
--- a/libc-top-half/musl/include/pthread.h
+++ b/libc-top-half/musl/include/pthread.h
@@ -55,15 +55,9 @@ extern "C" {
 #define PTHREAD_PROCESS_SHARED 1
 
 
-#if defined(__wasilibc_unmodified_upstream) || defined(_REENTRANT)
 #define PTHREAD_MUTEX_INITIALIZER {{{0}}}
 #define PTHREAD_RWLOCK_INITIALIZER {{{0}}}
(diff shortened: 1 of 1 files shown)


abrown (approved):
I think the idea made sense to me but initially it was a bit unclear why the comparison test for  `expected/wasm32-wasi/predefined-macros.txt` was not failing: if we unconditionally `#define` these `PTHREAD_*_INITIALIZER`s then I was expecting a diff error. But no: `pthread.h` is already omitted entirely when `THREAD_MODEL=single` in the top-level `Makefile` so these will only be defined when `THREAD_MODEL=posix`. And they will always be defined then, which I think is what we want. LGTM!

sunfishcode (approved):
Looks good to me!

Issue #398, opened 2023-02-28
libc.a is not built reproducibility/deterministically


PieroV:
When building `libc.a` in different machines with identical configuration, I get a different results.
It seems that the problem is the order in which the archive is created (I see some `wildcard` in the `Makefile`, but adding `$(sort )` to them is not enough).

As a result, a Firefox with WASI sandbox cannot be built in  a [reproducible way](https://reproducible-builds.org/).
Reproducibility matters to several projects, like Debian or Tor Browser (disclosure: I am a Tor Browser developer).

As a workaround, I created a Python script to re-create `libc.a`, but I wonder if something can be done to create `libc.a` deterministically in the first place (maybe some tweaks to the `Makefile` are enough).

Please notice that extracting `libc.a` with `llvm-ar x libc.a` and then creating it again with a sorted input list does not work, because the archive currently contains two `errno.o`, and also some object files without a name.

abrown:
I think it makes sense to attempt this; thanks for looking into it. Off the top of my head, I can't think of what the fix would be; you've tried the same standard things I would have tried. Perhaps posting more information about the differences would make things more clear? In any case, if you do figure out a fix, feel free to tag me in the review.

PieroV:
I have tried again to try to give you more information.

This time I have set `$(sort)` both in the wildcards and in the `%.a` inputs.
Let me open a PR with my changes.

PR #399, opened 2023-03-01
Use sorted inputs in the Makefile

PieroV:
This makes builds reproducible (see #398).

I have tried in two separate Debian testing machines, with different hardware specs (in particular, I think the number of cores can influence the output), the repository cloned in `/tmp` and LLVM/Clang provided by Debian (currently, 14.0.6).

I had this hash in both machines:

```
4490208a83e0da0d4e4cf88a1b15b3e3704e9f45ae79a1339d19603ce3f00ade  sysroot/lib/wasm32-wasi/libc.a
```

@abrown

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -488,13 +488,13 @@ $(SYSROOT_LIB)/libwasi-emulated-signal.a: $(LIBWASI_EMULATED_SIGNAL_OBJS) $(LIBW
 %.a:
 	@mkdir -p "$(@D)"
 	# On Windows, the commandline for the ar invocation got too long, so it needs to be split up.
-	$(AR) crs $@ $(wordlist 1, 199, $^)
-	$(AR) crs $@ $(wordlist 200, 399, $^)
-	$(AR) crs $@ $(wordlist 400, 599, $^)
-	$(AR) crs $@ $(wordlist 600, 799, $^)
+	$(AR) crs $@ $(wordlist 1, 199, $(sort $^))
+	$(AR) crs $@ $(wordlist 200, 399, $(sort $^))
+	$(AR) crs $@ $(wordlist 400, 599, $(sort $^))
+	$(AR) crs $@ $(wordlist 600, 799, $(sort $^))
 	# This might eventually overflow again, but at least it'll do so in a loud way instead of
 	# silently dropping the tail.
-	$(AR) crs $@ $(wordlist 800, 100000, $^)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -487,13 +487,13 @@ $(SYSROOT_LIB)/libwasi-emulated-signal.a: $(LIBWASI_EMULATED_SIGNAL_OBJS) $(LIBW
...
-	$(AR) crs $@ $(wordlist 400, 599, $^)
-	$(AR) crs $@ $(wordlist 600, 799, $^)
+	$(AR) crs $@ $(wordlist 1, 199, $(sort $^))
+	$(AR) crs $@ $(wordlist 200, 399, $(sort $^))
+	$(AR) crs $@ $(wordlist 400, 599, $(sort $^))
+	$(AR) crs $@ $(wordlist 600, 799, $(sort $^))
 	# This might eventually overflow again, but at least it'll do so in a loud way instead of
 	# silently dropping the tail.
-	$(AR) crs $@ $(wordlist 800, 100000, $^)
+	$(AR) crs $@ $(wordlist 800, 100000, $(sort $^))
```
sbc100, line 496: Isn't it enough to simply sort when calling AR here?  

I don't see why the sorting needs to happen up in `LIBC_TOP_HALF_MUSL_SOURCES`.   IIUC those all get built in parallel anyway so it won't dictate the order of the building of the object files?

Also we should probably just switch to using a response file here rather than using multiple calls to AR.. but thats a different issue.

sbc100: Sorting the inputs to AR makes total sense.   We can land that now.

I would like to know why the filelists need to be sorted though.

PieroV: > I would like to know why the filelists need to be sorted though.

Sorting the objects was my first approach, but it didn't work, for some reason (maybe I missed one of the `$^`).
So, I moved to sorting the filelists (I thought the object list was generated from them, I am not good at reading `Makefile`s), and that was not enough, too.
Eventually I have tried to sort both, and it worked, so I have opened the PR with this change.

Only after your prompt I have checked again the sort applied only to the object list.
So, I think that the filelist doesn't need to be sorted after all (it doesn't hurt either, though :slightly_smiling_face: ).

abrown (approved):
If this more minimal version works to make the builds reproducible, I'm a +1!

PieroV:
I've tried the latest revision again with our build system and it seems that yes, it's enough to get a reproducible wasi-libc and even a reproducible wasi-sdk :smile: .

PR #401, opened 2023-03-06
Use __builtin_ctz and __builtin_clz in dlmalloc

TerrorJack:
This patch ports https://github.com/emscripten-core/emscripten/pull/18186 to `wasi-libc` to get a bit of speed & size improvement for cheap :)


sunfishcode:
Nice find!

PR #407, opened 2023-04-05
Fix debug build's `predefined-macros.txt`

alexcrichton:
This commit fixes the ability to build `wasi-libc` with `-g` options and possibly without `-O2` options as well. I've found this useful when debugging issues as historically that the build fails when `-g` is passed or optimizations are removed due to the checks against these expectation files. This commit adds more filters to the list of macros to ensure that optimization/debug related ones are all removed from the expectation lists.


sunfishcode:
Yes, this makes sense. The main use for the predefined macros files is to monitor changes to the very large set of macros that libc is expected to define, and so it's ok to trim some of the boring macros out.

PR #408, opened 2023-04-12
Convert preopen initialization to be lazy.

sunfishcode:
Insteead of eagerly initializing the preopens in a static constructor, perform preopen initialization the first time it's needed, or before a close or a renumber which might disrupt the file descriptor space.

And, use a weak symbol with a stub function for use by `close` or `fd_renumber`, we that they can trigger preopen initialization only if it's actually needed.

This way, if a program doesn't contain any calls to any function that needs preopens, it can avoid linking in the preopen initialization code. And if it contains calls but doesn't execute them at runtime, it can avoid executing the preopen intiailization code.

A downside here is that this may cause problems for users that call `__wasi_fd_close` or `__wasi_fd_renumber` directly and close over overwrite some preopens before libc has a chance to scan them. To partially address this, this PR does add a declaration for `__wasilibc_populate_preopens` to <wasi/libc.h> so that users can call it manually if they need to.


alexcrichton (approved):

syrusakbary:
I think this defeats the propose of how preopens were designed in cloudabi? (for which WASI was designed upon)

Basically, if this gets merged, a malicious program can be designed to inspect what are the preopened dirs (and attack them with other vectors) using timing strategies.
Did I misunderstand that or can we assure there will no side-effects regarding security if the PR gets merged?

sbc100:
> I think this defeats the propose of how preopens were designed in cloudabi? (for which WASI was designed upon)
> 
> Basically, if this gets merged, a malicious program can be designed to inspect what are the preopened dirs (and attack them with other vectors) using timing strategies. Did I misunderstand that or can we assure there will no side-effects regarding security if the PR gets merged?

I don't think I understand.   Can you explain a bit more?   It seems like its up to the wasm module if and when it performs the pre-opens.. why would delaying (or even simply never performing) the pre-open syscalls make any difference?    In the threat model you are referring to where is the malicious code running?    And what is the vulnerable code?

libc-bottom-half/sources/preopens.c
```diff
@@ -219,13 +224,20 @@ int __wasilibc_find_abspath(const char *path,
...
-// libc-bottom-half/sources/environ.c.
-__attribute__((constructor(51)))
-static void __wasilibc_populate_preopens(void) {
+void __wasilibc_populate_preopens(void) {
+    // Fast path: If the preopens are already initialized, do nothing.
+    if (preopens_populated) {
+        return;
+    }
+
+    LOCK(lock);
```
yamt, line 233: does this work with `_REENTRANT`?
this function calls internal_register_preopened_fd which tries to acquire the same lock.

syrusakbary:
> Can you explain a bit more? It seems like its up to the wasm module if and when it performs the pre-opens.. why would delaying (or even simply never performing) the pre-open syscalls make any difference?

Basically, the main reason is [timing attacks](https://en.wikipedia.org/wiki/Timing_attack). That's why in secure places rather than comparing strings you usually compare hashes of them (because they are constant in time no matter the length of the string).
If you don't precompute the preopens it can become obvious to an attacker which are the preopened directories. That by itself is not an issue per se, but it can become one very quickly

sbc100:
> > Can you explain a bit more? It seems like its up to the wasm module if and when it performs the pre-opens.. why would delaying (or even simply never performing) the pre-open syscalls make any difference?
> 
> Basically, the main reason is [timing attacks](https://en.wikipedia.org/wiki/Timing_attack). That's why in secure places rather than comparing strings you usually compare hashes of them (because they are constant in time no matter the length of the string). If you don't precompute the preopens it can become obvious to an attacker which are the preopened directories. That by itself is not an issue per se, but it can become one very quickly

IIUC the pre-opening actually happens in on the host side, via command line flags such as `--dir=` in wasmtime.    Whether or when the wasm module decides to enumerate, or access those preopened files is completely up the userspace program, isn't it?    I don't think we can or should try to mandate if or when a program chooses to access them.

Some programs might choose never to access the re-opened files or directories at all.  IIUC wasi-libc already tries to completely avoid re-opening for programs that don't use file access at all.   Why include that that code at in programs that don't actually use files, right?  

Furthermore, I don't think that the design of the pre-open system had anything to do with timing attacks, its all about providing a way to pass the initial file handles to a program in system that doesn't support `open` directly (only `openat`, which requires at least an initial directory fd).   I could be wrong though, do you have any evidence to suggest that this was part of the design?

yamt:
> I think this defeats the propose of how preopens were designed in cloudabi? (for which WASI was designed upon)
> 
> Basically, if this gets merged, a malicious program can be designed to inspect what are the preopened dirs (and attack them with other vectors) using timing strategies. Did I misunderstand that or can we assure there will no side-effects regarding security if the PR gets merged?

what exactly does such a malicious program want to measure the timing of?
if it's wasi api like `fd_prestat_dir_name`, such a malicious program can happily call it by itself to measure, regardless of this PR, can't it?

libc-bottom-half/sources/preopens.c
sunfishcode: Indeed, good catch. I've now fixed that.

sunfishcode:
Indeed, this PR isn't doing anything that malicious code can't already do. Are there any further comments here?

yamt:
lgtm except that i don't understand syrusakbary's security concern.

sunfishcode:
If there are any further concerns, please file an [issue](https://github.com/WebAssembly/wasi-libc/issues).

achille-roussel:
Hello @sunfishcode !

I've been working on integrating WASI in Go recently, and I am wondering whether we should replicate this behavior in Go. My thinking is that the more guests behave similarly across languages, the more likely they are to avoid subtle bugs when interacting with the host.

Would you be able to share more details about the reasons that motivated this change?

sunfishcode:
Not all programs need preopens, so this change allows programs that don't use them at all to avoid linking them in, and programs that don't use them dynamically to avoid calling the preopen initialization functions.

It also potentially helps us with a broader question about how C++-style static constructors should work in Wasm, because previously the preopens were initialized with such a constructor, making it one of the more common reasons why many programs had a constructor in them. With this change, many programs can avoid having any constructors. Some still will, but if it's a small number, we may have more options.

achille-roussel:
Thanks for sharing the context!

PR #410, opened 2023-04-19
Acquire the global lock before initializing malloc

MoritzS:
In a multi-threaded execution we need to make sure that only exactly one thread initializes malloc. The function try_init_allocator() can't easily be made thread-safe, so just move the call to try_init_allocator() inside the block that holds the lock.


yamt (approved):
lgtm

abrown (approved):

PR #426, opened 2023-06-29
Add definitions for PF_INET, PF_INET6 and PF_UNSPEC

loganek:
Given there are already AF_* definitions, and they are (now) essentially synonyms, we add those definitions to enable compilation of code that already use PF_* macros.


abrown (commented):
Can we fix up the removed white spaces? I think with that this diff should be much smaller and I will merge it.

loganek:
> Can we fix up the removed white spaces? I think with that this diff should be much smaller and I will merge it.

done

abrown (approved):
Thanks!

PR #429, opened 2023-07-28
add shared library support

dicej:
This adds support for building WASI shared libraries per https://github.com/WebAssembly/tool-conventions/blob/main/DynamicLinking.md.

For the time being, the goal is to allow "pseudo-dynamic" linking using the Component Model per
https://github.com/WebAssembly/component-model/blob/main/design/mvp/examples/SharedEverythingDynamicLinking.md. This requires all libraries to be available when the component is created, but still allows runtime symbol resolution via `dlopen`/`dlsym` backed by a static lookup table.  This is sufficient to support Python native extensions, for example.  A complete demo using `wit-component` is available at https://github.com/dicej/component-linking-demo.

This commit adds support for building `libc.so`, `libc++.so`, and `libc++abi.so` alongside their static counterparts.

Notes:

- I had to refactor `errno` support a bit to avoid a spurious `_ZTH5errno` (AKA "thread-local initialization routine for errno") import in `libc++.so`.
- Long double print and scan are included by default in `libc.so` rather than in a separate library.
- `__main_argc_argv` is now a weak symbol since it's not relevant for reactors.
- `dlopen`/`dlsym` rely on a lookup table provided by the "dynamic" linker via `__wasm_set_libraries`.  Not all flags are supported yet, and unrecognized flags will result in an error.
- This requires https://reviews.llvm.org/D153293, which we will need to backport to LLVM 16 until 17 is released.  I'll open a `wasi-sdk` PR with that change and various Makefile tweaks to support shared libraries.
- `libc.so` is temporarily disabled for the `wasi-threads` build until someone can make `wasi_thread_start.s` position-independent.


Makefile
```diff
@@ -8,7 +8,7 @@ NM ?= $(patsubst %clang,%llvm-nm,$(filter-out ccache sccache,$(CC)))
 ifeq ($(origin AR), default)
 AR = $(patsubst %clang,%llvm-ar,$(filter-out ccache sccache,$(CC)))
 endif
-EXTRA_CFLAGS ?= -O2 -DNDEBUG
+EXTRA_CFLAGS ?= -O2 -DNDEBUG -fPIC -fvisibility=default
```
TerrorJack, line 11: The `-fPIC` overhead also applies to the original non-shared `libc.a` which is not ideal IMHO.

dicej: Yeah, that's a topic that's come up a few times.  @sunfishcode took a look at libc.a before and after `-fPIC` was added and concluded that it wasn't a big deal from his perspective.  The alternative is to build everything twice -- once for static and once for dynamic.  Either way is fine with me, as long as we have consensus.

loganek: Are there any disadvantages of building everything twice other than doubled build time? If not, I think that'd be my preference

dicej: Not that I'm aware of.  @sunfishcode or @alexcrichton do you have any concerns about building the libc and libc++ code twice (with and without `-fPIC`)?  (Actually four times, since we're already building twice for threads/no-threads)

yamt:
> I had to refactor errno support a bit to avoid a spurious _ZTH5errno (AKA "thread-local initialization routine for errno") import in libc++.so.

can you explain a bit?

with the patch reverted, i got:
```
wasm-ld: error: libcxx/src/CMakeFiles/cxx_shared.dir/atomic.cpp.o: relocation R_WASM_MEMORY_ADDR_TLS_SLEB cannot be used against an undefined symbol `errno`
```
is this what you meant?

> Long double print and scan are included by default in libc.so rather than in a separate library.

i'd suggest to drop this because it's an unrelated change.

> dlclose/dlsym rely on a lookup table provided by the "dynamic" linker via __wasm_set_libraries. Not all flags are supported yet, and unrecognized flags will result in an error.

i'd suggest to drop this as it seems like a kludge for the specific demo.

> libc.so is temporarily disabled for the wasi-threads build until someone can make wasi_thread_start.s position-independent.

it's fine as wasi-threads as of today is not compatible witt multiple module linking setup like this.

Makefile
```diff
@@ -472,6 +472,19 @@ endif
...
+	$(CC) -nostdlib -shared -o $@ -Wl,--whole-archive $< -Wl,--no-whole-archive $(BUILTINS_LIB)
+
+$(SYSROOT_LIB)/libc.so.a: \
+		$(filter-out $(MUSL_PRINTSCAN_OBJS),$(LIBC_OBJS)) \
+		$(MUSL_PRINTSCAN_LONG_DOUBLE_OBJS) \
+		$(LIBWASI_EMULATED_MMAN_OBJS) \
+		$(LIBWASI_EMULATED_PROCESS_CLOCKS_OBJS) \
+		$(LIBWASI_EMULATED_GETPID_OBJS) \
+		$(LIBWASI_EMULATED_SIGNAL_OBJS) \
+		$(LIBWASI_EMULATED_SIGNAL_MUSL_OBJS)
```
yamt, line 486: including EMULATED things in libc is an unrelated change. please drop.

yamt (changes requested):
also, please revert whitespace changes.

dicej:
> > I had to refactor errno support a bit to avoid a spurious _ZTH5errno (AKA "thread-local initialization routine for errno") import in libc++.so.
> 
> can you explain a bit?
> 
> with the patch reverted, i got:
> 
> ```
> wasm-ld: error: libcxx/src/CMakeFiles/cxx_shared.dir/atomic.cpp.o: relocation R_WASM_MEMORY_ADDR_TLS_SLEB cannot be used against an undefined symbol `errno`
> ```
> 
> is this what you meant?

Originally, I hit a different issue such that libc++.so would build without trouble, but import `_ZTH5errno`, leading to undefined symbol errors when _using_ libc++.so.  However, something must have changed since then because now I'm getting the same error you reported above.

> > Long double print and scan are included by default in libc.so rather than in a separate library.
> 
> i'd suggest to drop this because it's an unrelated change.

If we don't build libc.so with long double print/scan support, how will applications enable it?  They can't link against libc-printscan-long-double.a because that will result in duplicate definitions, and I believe we'd have the same problem if we built a libc-printscan-long-double.so since its definitions will also conflict with those of libc.so.

> > dlclose/dlsym rely on a lookup table provided by the "dynamic" linker via __wasm_set_libraries. Not all flags are supported yet, and unrecognized flags will result in an error.
> 
> i'd suggest to drop this as it seems like a kludge for the specific demo.

Would you please elaborate on why you feel it's a kludge?  Do you have a different approach in mind for supporting `dlopen`, `dlsym`, etc. that would work in a WASI runtime?

Makefile
dicej: I want to make sure application programmers can use these features in a shared library context if they need them.  Putting them in libc.so would be the most convenient option, but I'm open to alternatives.  Is your main concern that they make libc.so larger that it otherwise needs to be?  If so, would you prefer that they go in a separate library, e.g. libc-emulated.so?

dicej:
> also, please revert whitespace changes.

Done!

dicej:
Regarding the CI failures: stock LLVM versions prior to 17 simply can't build .so files, so ideally we'd just skip building libc.so in that case.  Probably the easiest option would be to not include libc.so in the default makefile target, i.e. only build it when explicitly requested.  Does that sound reasonable?

Makefile
sunfishcode: I don't have strong opinions either way. The -fPIC overhead from the builds I've seen was low. But also, I don't think building twice is a problem either, unless it makes something more complex.

TerrorJack: > The -fPIC overhead from the builds I've seen was low

imho it can only be low after a `wasm-opt` run that replaces all symbols with i32 literals again. it would really be nice if the static linking use case doesn't pay for what it doesn't use.

dicej: I just pushed an update which builds a separate set of `-fPIC` object files for libc.so in addition to the non-`-fPIC` ones for libc.a.

yamt:
> > > I had to refactor errno support a bit to avoid a spurious _ZTH5errno (AKA "thread-local initialization routine for errno") import in libc++.so.
> > 
> > 
> > can you explain a bit?
> > with the patch reverted, i got:
> > ```
> > wasm-ld: error: libcxx/src/CMakeFiles/cxx_shared.dir/atomic.cpp.o: relocation R_WASM_MEMORY_ADDR_TLS_SLEB cannot be used against an undefined symbol `errno`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is this what you meant?
> 
> Originally, I hit a different issue such that libc++.so would build without trouble, but import `_ZTH5errno`, leading to an undefined symbol error when _using_ libc++.so. However, something must have changed since then because now I'm getting the same error you reported above.

ok.

> 
> > > Long double print and scan are included by default in libc.so rather than in a separate library.
> > 
> > 
> > i'd suggest to drop this because it's an unrelated change.
> 
> If we don't build libc.so with long double print/scan support, how will applications enable it? They can't link against libc-printscan-long-double.a because that will result in duplicate definitions, and I believe we'd have the same problem if we built a libc-printscan-long-double.so since its definitions will also conflict with those of libc.so.

why?
it's common to have a shared library which override some symbols in other library. not specific to wasm.

> 
> > > dlclose/dlsym rely on a lookup table provided by the "dynamic" linker via __wasm_set_libraries. Not all flags are supported yet, and unrecognized flags will result in an error.
> > 
> > 
> > i'd suggest to drop this as it seems like a kludge for the specific demo.
> 
> Would you please elaborate on why you feel it's a kludge? Do you have a different approach in mind for supporting `dlopen`, `dlsym`, etc. that would work in a WASI runtime?

yes, different (and IMO more natural) approaches are possible.
eg. https://github.com/yamt/toywasm/tree/master/examples/libdl

Makefile
yamt: > I want to make sure application programmers can use these features in a shared library context if they need them. Putting them in libc.so would be the most convenient option, but I'm open to alternatives. Is your main concern that they make libc.so larger that it otherwise needs to be? If so, would you prefer that they go in a separate library, e.g. libwasi-emulated.so?

sort of.
i feel it's better to keep things consistent between .a and .so where reasonable.
yes, i prefer to have libwasi-emulated-process-clocks.so etc.

yamt:
> Regarding the CI failures: stock LLVM versions prior to 17 simply can't build .so files, so ideally we'd just skip building libc.so in that case. Probably the easiest option would be to not include libc.so in the default makefile target, i.e. only build it when explicitly requested. Does that sound reasonable?

making it a non-default target for now sounds reasonable to me.

dicej:
> > If we don't build libc.so with long double print/scan support, how will applications enable it? They can't link against libc-printscan-long-double.a because that will result in duplicate definitions, and I believe we'd have the same problem if we built a libc-printscan-long-double.so since its definitions will also conflict with those of libc.so.
> 
> why? it's common to have a shared library which override some symbols in other library. not specific to wasm.

They're currently not weak symbols, but yes, we could make them weak in libc.so and strong in libc-printscan-long-double.so.

> > Would you please elaborate on why you feel it's a kludge? Do you have a different approach in mind for supporting `dlopen`, `dlsym`, etc. that would work in a WASI runtime?
> 
> yes, different (and IMO more natural) approaches are possible. eg. https://github.com/yamt/toywasm/tree/master/examples/libdl

I'll move dl.c into a separate PR so we can discuss it separately.

sbc100:
> > They're currently not weak symbols, but yes, we could make them weak in libc.so and strong in libc-printscan-long-double.so.
> 
> Well, I tried this, but I can't seem to make it work -- at least not the way I had planned.
> 
> My expectation was that adding `__attribute__((__weak__))` to the relevant symbols would result in an `WASM_DYLINK_EXPORT_INFO` entry being added the `dylink.0` for each symbol, with a flags field that includes `WASM_SYM_BINDING_WEAK`. However, it seems that `wasm-ld` does not generate such entries for any weak definitions (i.e. exports) -- only for weak imports. So when the dynamic linker sees the same symbol exported by both libc.so and libc-printscan-long-double.so, it doesn't know which one takes precedence.
> 
> I'm not sure if that's a bug or intended behavior. @sbc100 do you know? Is the intention that the dynamic linker should use e.g. `WASM_DYLINK_NEEDED` to determine search order when more than one library exports the same symbol (i.e. rather than weak vs. strong symbols)?

Yes, this order in which libraries are loaded determines which one gets to define a given symbol.   At last this is how to works on linux and this is the behaviour we try to emulate in emscripten.

However, in think I I don't see why we wouldn't just have a single version of libc.so which has the printscan-long-double stuff.   Trying to make libc.so smaller seems a little pointless given that it is already a completely superset of the entire libc.  By definition isn't can't do any DCE to be tuned to a specific app.    Unless the difference between the size of the two builds of libc.so is really enormous I don't think it makes sense to split in the same way we do for the static build.

Anther way of putting it: The point of libc.so is that the size cost of libc.so itself is not paid by the individual application but instead is amortized over all libc consumers.

dicej:
Thanks for the insight, @sbc100.  I agree that putting everything in libc.so is a good place to start given its intended use.

For reference, the sizes of libc.so with and without the printscan-long-double and emulated stuff are 636222 and 632310 bytes, respectively, so the difference is quite minimal.

Makefile
alexcrichton: I apologize for not reviewing this earlier I've only gotten around to this just now! 

One point I would say in favor of only having one build is that by having two sysroots everyone is now presented with a choice of which one to use and it's possible to get it wrong. That can add downstream complexity to builds, configuration, etc. Having only one build solves this because you can't make the wrong choice and there's only one canonical version of libc.a
 
I see now though I'm late to the punch and you've already updated this to build two copies. I don't feel particularly strongly either way, but I think that this is something to watch out for when integrating this into other toolchains. If it's onerous/difficult to have multiple copies of libc then I think it's worth reconsidering. One example of this is that to use this from Rust there will have to be new target triples, and no preexisting target I think differentiates target triples purely on dynamic linking of the same C standard library.

TerrorJack: > because you can't make the wrong choice

maybe it is the compiler driver's responsibility to automatically pick between `.a` and `.so` from the same sysroot libdir, based on whether it is doing dynamic linking or not?

alexcrichton: While possible, that's forcing all compiler drivers, embedders, etc, to have duplicate copies and sysroots. For example the rust standard library would need to be compiled for both dynamic linking and not and store both sysroots, it's not purely a concern for WASI. Rust doesn't currently do this for any other target that I'm aware of, and if that's the case it's unlikely to happen for just the WASI target.

To me at least this cost is not worth the minor savings of code size in libc which optimizing compilers/engines will all strip away anyway.

TerrorJack: apologies in advance given my rust experience is limited and error-prone; afaik `rust-std` already ships both `libstd-*.so` / `libstd-*.rlib` for various targets, so it looks like a natural conclusion to ship both `-fPIC`/non-`-fPIC` variants of `libstd` for wasm, and for `wasm32-wasi`, ship two variants of `libc` as well (in the same sysroot libdir, to be chosen at link-time)

alexcrichton: Indeed! Internally though those two artifacts have the same machine code. The standard library is compiled once simultaneously with a dylib/rlib output which means that LLVM generates one copy of all the code for the standard library and then that object file is linked into two locations. This is enabled because all Rust platforms (AFAIK) generate position-independent code by default and eschew the minor size-related optimizations from generating static code. That's left for users to build their own toolchains if that size difference matters.

I'll also admit though I'm probably quite biased on this as I wrote the support for rlibs/dylibs and am the reason that the standard library ships two artifacts for example.

Makefile
```diff
@@ -85,90 +85,90 @@ LIBC_TOP_HALF_MUSL_SRC_DIR = $(LIBC_TOP_HALF_MUSL_DIR)/src
 LIBC_TOP_HALF_MUSL_INC = $(LIBC_TOP_HALF_MUSL_DIR)/include
 LIBC_TOP_HALF_MUSL_SOURCES = \
     $(addprefix $(LIBC_TOP_HALF_MUSL_SRC_DIR)/, \
-        misc/a64l.c \
```
yamt, line 88: this file still seems to contain a ton of whitespace changes.

Makefile
```diff
@@ -472,6 +471,31 @@ endif
...
+BULK_MEMORY_SO_OBJS = $(patsubst %.o,%.pic.o,$(BULK_MEMORY_OBJS))
+DLMALLOC_SO_OBJS = $(patsubst %.o,%.pic.o,$(DLMALLOC_OBJS))
+LIBC_BOTTOM_HALF_ALL_SO_OBJS = $(patsubst %.o,%.pic.o,$(LIBC_BOTTOM_HALF_ALL_OBJS))
+LIBC_TOP_HALF_ALL_SO_OBJS = $(patsubst %.o,%.pic.o,$(LIBC_TOP_HALF_ALL_OBJS))
+
+# TODO: specify SDK version, e.g. libc.so.wasi-sdk-21, as SO_NAME once `wasm-ld` supports it
+$(SYSROOT_LIB)/libc.so: $(SYSROOT_LIB)/libc.so.a $(BUILTINS_LIB)
+	$(CC) -nostdlib -shared -o $@ -Wl,--whole-archive $< -Wl,--no-whole-archive $(BUILTINS_LIB)
+
+$(SYSROOT_LIB)/libc.so.a: \
```
yamt, line 490: what's this "so.a"?

libc-bottom-half/sources/__main_void.c
```diff
@@ -3,6 +3,7 @@
 #include <sysexits.h>
 
 // The user's `main` function, expecting arguments.
+__attribute__((__weak__, nodebug))
```
yamt, line 6: why nodebug?

expected/wasm32-wasi-threads/predefined-macros.txt
```diff
@@ -1988,18 +1997,18 @@
 #define YESSTR 0x50002
 #define YXDOMAIN ns_r_yxdomain
 #define YXRRSET ns_r_yxrrset
-#define _ALLOCA_H 
+#define _ALLOCA_H
```
yamt, line 2000: ton of unrelated whitespace changes

yamt:
> Thanks for the insight, @sbc100. I agree that putting everything in libc.so is a good place to start given its intended use.
> 
> For reference, the sizes of libc.so with and without the printscan-long-double and emulated stuff are 636222 and 632310 bytes, respectively, so the difference is quite minimal.

well, my understanding is that the reason why the "emulated" stuff are separated is not size.
it's the fact they are emulations.

i have no strong opinions on long-double things.

yamt:
> > > If we don't build libc.so with long double print/scan support, how will applications enable it? They can't link against libc-printscan-long-double.a because that will result in duplicate definitions, and I believe we'd have the same problem if we built a libc-printscan-long-double.so since its definitions will also conflict with those of libc.so.
> > 
> > 
> > why? it's common to have a shared library which override some symbols in other library. not specific to wasm.
> 
> They're currently not weak symbols, but yes, we could make them weak in libc.so and strong in libc-printscan-long-double.so.

weak symbols are not related for this purpose. it's merely symbol search order.

> > > Would you please elaborate on why you feel it's a kludge? Do you have a different approach in mind for supporting `dlopen`, `dlsym`, etc. that would work in a WASI runtime?
> > 
> > 
> > yes, different (and IMO more natural) approaches are possible. eg. https://github.com/yamt/toywasm/tree/master/examples/libdl
> 
> I'll move dl.c into a separate PR so we can discuss it separately.

can you move dlfcn.h related changes too?

yamt:
> > > I had to refactor errno support a bit to avoid a spurious _ZTH5errno (AKA "thread-local initialization routine for errno") import in libc++.so.
> > 
> > 
> > can you explain a bit?
> > with the patch reverted, i got:
> > ```
> > wasm-ld: error: libcxx/src/CMakeFiles/cxx_shared.dir/atomic.cpp.o: relocation R_WASM_MEMORY_ADDR_TLS_SLEB cannot be used against an undefined symbol `errno`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is this what you meant?
> 
> Originally, I hit a different issue such that libc++.so would build without trouble, but import `_ZTH5errno`, leading to an undefined symbol error when _using_ libc++.so. However, something must have changed since then because now I'm getting the same error you reported above.

this is not relevant as far as we don't build shared+threads, right?
i'd suggest to drop this part too.

Makefile
dicej: Sorry, I thought I had fixed that; will try again.

libc-bottom-half/sources/__main_void.c
dicej: I probably copied the whole line from elsewhere; I'll remove that part.

Makefile
dicej: That's to avoid the Windows command line overflow issue by collecting the .o files into a .a file prior to making the .so:

```makefile
%.a:
	@mkdir -p "$(@D)"
	# On Windows, the commandline for the ar invocation got too long, so it needs to be split up.
	$(AR) crs $@ $(wordlist 1, 199, $(sort $^))
	$(AR) crs $@ $(wordlist 200, 399, $(sort $^))
	$(AR) crs $@ $(wordlist 400, 599, $(sort $^))
	$(AR) crs $@ $(wordlist 600, 799, $(sort $^))
	# This might eventually overflow again, but at least it'll do so in a loud way instead of
	# silently dropping the tail.
	$(AR) crs $@ $(wordlist 800, 100000, $(sort $^))
```

dicej:
> > Originally, I hit a different issue such that libc++.so would build without trouble, but import `_ZTH5errno`, leading to an undefined symbol error when _using_ libc++.so. However, something must have changed since then because now I'm getting the same error you reported above.
> 
> this is not relevant as far as we don't build shared+threads, right? i'd suggest to drop this part too.

I just tried that by updating the `wasi-sdk` Makefile to only build libc++.so and libc++abi.so for the non-threads build.  That got me past the "relocation R_WASM_MEMORY_ADDR_TLS_SLEB cannot be used" issue, but now I'm hitting the `_ZTH5errno` issue again.  So it seems it _is_ relevant, even when we avoid shared+threads.

I'm open to ideas on this, but so far it seems that the errno changes are necessary.

Makefile
dicej: I'll move it into `OBJDIR` so it isn't included in the distribution.

yamt:
> > > Originally, I hit a different issue such that libc++.so would build without trouble, but import `_ZTH5errno`, leading to an undefined symbol error when _using_ libc++.so. However, something must have changed since then because now I'm getting the same error you reported above.
> > 
> > 
> > this is not relevant as far as we don't build shared+threads, right? i'd suggest to drop this part too.
> 
> I just tried that by updating the `wasi-sdk` Makefile to only build libc++.so and libc++abi.so for the non-threads build. That got me past the "relocation R_WASM_MEMORY_ADDR_TLS_SLEB cannot be used" issue, but now I'm hitting the `_ZTH5errno` issue again. So it seems it _is_ relevant, even when we avoid shared+threads.
> 
> I'm open to ideas on this, but so far it seems that the errno changes are necessary.

ok.

`_ZTH5errno` is for dynamic initialization (ie. calling ctor) for C++ `thread_local`.
i wonder why clang generates these stuff even for `int`. maybe it's mandated by ABI? i dunno.

anyway, it seems that it's a bug for us to use C++ `thread_local` because it's incompatible with the C version which uses `_Thread_local`.

possible solutions i can think of right now are:
a. use `_Thread_local` in C++ too.
b. use `__errno_location` for C++. (follow https://github.com/WebAssembly/wasi-libc/pull/347)
c. use `__errno_location` for everything, including C.

for now, my suggestion is option a. or b. because:
* the option c. involves a bit overhead for C
* i'm conservative about changing libc ABI

yamt:
>  i wonder why clang generates these stuff even for int. maybe it's mandated by ABI? i dunno.

answering myself, as C++ allows more complex initializers than C, accessing `extern` thread-local vars need to prepare for them.

dicej:
> possible solutions i can think of right now are: a. use `_Thread_local` in C++ too. b. use `__errno_location` for C++. (follow #347) c. use `__errno_location` for everything, including C.

I went with solution (a) by changing

```c
#ifdef __cplusplus
extern thread_local int errno;
#else
extern _Thread_local int errno;
#endif
```
into just
```c
extern _Thread_local int errno;
```
That seems to work fine.  Does anyone have any concerns about that change?  Is there a deeper reason we've been using `thread_local` for C++ besides being idiomatic?

dicej:
I believe I've addressed all the feedback so far.  Does anyone have other concerns?

Makefile
```diff
@@ -472,6 +472,32 @@ endif
...
+
+$(OBJDIR)/libc.so.a: $(LIBC_SO_OBJS) $(MUSL_PRINTSCAN_LONG_DOUBLE_SO_OBJS)
+
+$(OBJDIR)/libwasi-emulated-mman.so.a: $(LIBWASI_EMULATED_MMAN_SO_OBJS)
+
+$(OBJDIR)/libwasi-emulated-process-clocks.so.a: $(LIBWASI_EMULATED_PROCESS_CLOCKS_SO_OBJS)
+
+$(OBJDIR)/libwasi-emulated-getpid.so.a: $(LIBWASI_EMULATED_GETPID_SO_OBJS)
+
+$(OBJDIR)/libwasi-emulated-signal.so.a: $(LIBWASI_EMULATED_SIGNAL_SO_OBJS) $(LIBWASI_EMULATED_SIGNAL_MUSL_SO_OBJS)
```
sbc100, line 499: Why produce these intermediate `.a` archives and then link them with `--whole-archive`?    i.e. Why not just add the objects to the link line?

Makefile
```diff
@@ -507,15 +533,23 @@ $(MUSL_PRINTSCAN_NO_FLOATING_POINT_OBJS): CFLAGS += \
...
+$(BULK_MEMORY_OBJS) $(BULK_MEMORY_SO_OBJS): CFLAGS += \
         -DBULK_MEMORY_THRESHOLD=$(BULK_MEMORY_THRESHOLD)
 
-$(LIBWASI_EMULATED_SIGNAL_MUSL_OBJS): CFLAGS += \
+$(LIBWASI_EMULATED_SIGNAL_MUSL_OBJS) $(LIBWASI_EMULATED_SIGNAL_MUSL_SO_OBJS): CFLAGS += \
 	    -D_WASI_EMULATED_SIGNAL
 
+$(OBJDIR)/%.long-double.pic.o: %.c include_dirs
+	@mkdir -p "$(@D)"
+	$(CC) $(CFLAGS) -fPIC -fvisibility=default -MD -MP -o $@ -c $<
```
sbc100, line 547: Is this chunk needed?

Makefile
```diff
@@ -596,6 +630,20 @@ startup_files: include_dirs $(LIBC_BOTTOM_HALF_CRT_OBJS)
...
+# TODO: As of this writing, wasi_thread_start.s uses non-position-independent
+# code, and I'm not sure how to make it position-independent.  Once we've done
+# that, we can enable libc.so for the wasi-threads build.
+ifneq ($(THREAD_MODEL), posix)
+LIBC_SO = \
+	$(SYSROOT_LIB)/libc.so \
+	$(SYSROOT_LIB)/libwasi-emulated-mman.so \
+	$(SYSROOT_LIB)/libwasi-emulated-process-clocks.so \
+	$(SYSROOT_LIB)/libwasi-emulated-getpid.so \
+	$(SYSROOT_LIB)/libwasi-emulated-signal.so
```
sbc100, line 642: I'm not sure having 5 separate shared libraries here is the right way to go.

Could we maybe start by simply not included these optional parts in the shared library build?   For now they could continue to be linked in statically if needed, and we can maybe come up with a more advanced solution later, as needed?

libc-bottom-half/sources/__main_void.c
```diff
@@ -3,6 +3,7 @@
 #include <sysexits.h>
 
 // The user's `main` function, expecting arguments.
+__attribute__((__weak__))
```
sbc100, line 6: Why is this needed in particular for the shared build? ?  Perhaps add a comment?

sbc100: ...  I wonder if this file should be logically part of crt1?

Makefile
dicej: See https://github.com/WebAssembly/wasi-libc/pull/429#discussion_r1289232216.

Makefile
dicej: We'd still need to build separate libraries using `-fPIC` (e.g. `libwasi-emulated-mman-pic.a`).  Or I guess if folks are okay with using `-fPIC` unconditionally for the relevant .o files, we could do that.

Makefile
dicej: Apparently not -- I'll remove it.

Makefile
sbc100: I was thinking the they would be linked into the main binary and wouldn't need `-fPIC`.   But that would limit their usefulness I guess.   This approach just seems a little overly complex, which is why I suggested punting on this for a while.  

An alternative approach, for example, would be to have a single libc-emulated.so or something like that.

dicej: FWIW, libraries sometimes need these features, too, so I wouldn't want to restrict them to the main binary.

I'd be in favor of a single libc-emulated.so.  @yamt would that be okay with you?

Makefile
sbc100: This is kind of sad.   Can you add a comment before the `--whole-archive` line to explain why we are doing this?   

I seems like there must be a better solution to this windows issue :(

dicej: I've added a comment.

Makefile
dicej: I went ahead and combined the libraries into a single `libwasi-emulated.so`.

Makefile
dicej: Oops, it _is_ needed.  If you do a clean build without it, you get this:
```
make[1]: *** No rule to make target `build/wasm32-wasi/libc-top-half/musl/src/internal/floatscan.long-double.pic.o', needed by `build/wasm32-wasi/libc.so.a'.  Stop.
```

libc-bottom-half/sources/__main_void.c
sunfishcode: wasi-libc doesn't use musl's crt1.c; it uses libc-bottom-half/crt/crt1-command.c and libc-bottom-half/crt/crt1-reactor.c. Putting the `__main_argc_argc` symbol there sounds reasonable to me.

dicej: I've moved both the `__main_void` definition and the `__main_argc_argv` declaration to crt1-command.c.

Makefile
yamt: > I'd be in favor of a single libc-emulated.so. @yamt would that be okay with you?

i'm not sure if it's a good idea.
it would make try-link style configure scripts detect all of the emulated stuff
once you link the single libc-emulated.so.
also, what to do about _WASI_EMULATED_xxx macros?
i feel it's conservative to mirror what we do for static libs at least for the first step.

Makefile
```diff
@@ -596,6 +635,17 @@ startup_files: include_dirs $(LIBC_BOTTOM_HALF_CRT_OBJS)
...
+# TODO: As of this writing, wasi_thread_start.s uses non-position-independent
+# code, and I'm not sure how to make it position-independent.  Once we've done
+# that, we can enable libc.so for the wasi-threads build.
+ifneq ($(THREAD_MODEL), posix)
+LIBC_SO = \
+	$(SYSROOT_LIB)/libc.so \
+	$(SYSROOT_LIB)/libwasi-emulated.so
+endif
+
+libc_so: include_dirs $(LIBC_SO)
```
yamt, line 647: is this the main target to build shared library?
which target builds pic version of crt1*.o?

libc-top-half/musl/src/internal/locale_impl.h
```diff
@@ -28,7 +28,7 @@ extern hidden const struct __locale_struct __c_dot_utf8_locale;
 hidden const struct __locale_map *__get_locale(int, const char *);
 hidden const char *__mo_lookup(const void *, size_t, const char *);
 hidden const char *__lctrans(const char *, const struct __locale_map *);
-hidden const char *__lctrans_cur(const char *);
+const char *__lctrans_cur(const char *);
```
yamt, line 31: is this intended?

dicej: Yes - libc-top-half/musl/src/string/strsignal.c (which is included in `LIBWASI_EMULATED_SIGNAL_MUSL_SOURCES`) uses `LCTRANS_CUR` (which is defined to be `__lctrans_cur`), which means libc.so must export it so libwasi-emulated.so can import it.

Makefile
dicej: > is this the main target to build shared library? 
Yes

> which target builds pic version of crt1*.o?
My understanding is that crt1.o is intended to be linked into the main module, which doesn't necessarily need to be PIC since it's the one exporting memory, not importing it.  I'll admit I've been focusing on reactor-style apps, though, so if there's something missing to support command-style apps I'm happy to add it.

yamt: > > which target builds pic version of crt1*.o?
> 
> My understanding is that crt1.o is intended to be linked into the main module, which doesn't necessarily need to be PIC since it's the one exporting memory, not importing it. I'll admit I've been focusing on reactor-style apps, though, so if there's something missing to support command-style apps I'm happy to add it.

i was assuming that `-pie` was used to produce a dynamically linked executable.
https://github.com/WebAssembly/tool-conventions/blob/main/DynamicLinking.md#llvm-implementation
but as you say, it might be possible to produce a non-pic dynamically linked executable.
i haven't tried.

btw, a reactor module is linked with crt1-reactor.o as well.
i guess you meant shared libraries, for which if crt1-reactor.o should be linked isn't obvious.

yamt: > but as you say, it might be possible to produce a non-pie dynamically linked executable.
i haven't tried.

it seems clang can somehow produce non-pie dynamically linked main module.
eg.
```
-fPIC
-nodefaultlibs
-Xlinker --unresolved-symbols=import-dynamic
-Xlinker --export-table
-Xlinker --growable-table
```

a few things are not obvious to me though:
* what to do for `__stack_pointer`. should the main module export it? or import it?
* how about `__heap_base` and `__heap_end`?

on the other hand, pie main module is more straightforward and known to work as it's what emscripten `-sMAIN_MODULE` uses.
to produce pie main modules, we need pic (or at least `dynamic-no-pic`) version of crt1.

how do you think?

dicej: Yes, I believe the main module must export `__stack_pointer`, `__heap_base`, and `__heap_end` (or at least _some_ module must export them, because shared libraries will import them in general).

I'd be fine with building crt1-*.o with `-fPIC` unconditionally -- it's a very small amount of code, so I doubt there will be a noticeable performance penalty.  Does anyone have concerns about that?

sbc100: As a data point, emscripten currently always builds the main module as `-pie` and then the runtime provides things like `__stack_pointer`, `__heap_end`, etc.  I've been trying for a while to convert emscripten to use statically linked main modules but I have not had any success yet.  See https://github.com/emscripten-core/emscripten/issues/12682

dicej: FWIW, `wasm-tools component link` synthesizes a main module which exports `_stack_pointer` etc. based on a set of input shared libraries.  In other words, all the application code and its dependencies are built as shared libraries and "pseudo-dynamically" linked together into a component, so I don't believe crt1.o plays any role in the process.  I realize that's not the _only_ way WASI shared libraries might be used, but that's the use case I've been focusing on so far.

sbc100: I think `crt1.o` (or something like it) will always be needed in order to define the `_initialize` function (who primary job, at least in the llvm world is to call `__wasm_call_ctors`).

dicej: > I think `crt1.o` (or something like it) will always be needed in order to define the `_initialize` function (who's primary job, at least in the llvm world is to call `__wasm_call_ctors`).

`wasm-tools component link` also synthesizes an "init" module which contains a `start` function that calls `__wasm_apply_data_relocs`, `__wasm_call_ctors`, initializes `GOT` globals, etc.

Makefile
dicej: To sum up the conversation so far:

- Initially, I bundled everything into libc.so for simplicity
- @yamt prefers we follow the pattern used for the static libraries and build a separate libwasi-emulated-*.so corresponding to each libwasi-emulated-*.a file
- @sbc100 is concerned that building so many .so files is overly complex and would prefer a single libwasi-emulated.so

Personally, I just want to get this PR merged, so I'm fine with any of the above.  If the maintainers of this project can come to a decision on this, I'm happy to implement any changes required.

Makefile
sbc100: I think in the long run we should use something like `_initialize` exported rather than `__wasm_call_ctors` (which I see as an llvm-specific thing).

`__wasm_apply_data_relocs` is a tricky one because (at least in cyclic module graphs) we need to call that on the whole module graph before we call `__wasm_call_ctors` on the whole module graph.. so perhaps we need to standardize on some kind of phase 1 and phase 2 `_initialize` functions.

Makefile
sbc100: I don't feel strongly about the multiple `.so` files and in fact I can see value in mirroring the `.a` names.   Sorry for the noise.  

I wish we didn't need all those special `.so` files but I guess since we decided to make the emulation optional we don't have much choice.

dicej: Thanks, @sbc100.  I've reverted the change the combined the `libwasi-emulated-*.so` into a single library, so we're back to four separate ones.

Makefile
dicej: I've pushed an update to build `crt1-*.o` with `-fPIC`.

dicej:
Okay, so CI's failing again, and I think it's because of my most recent change to build `crt1-*.o` with `-fPIC`.  It seems that LLVM 10 generates different code than the patched version of LLVM 16 that I've been using to test.  Specifically, the latter produces code that references `__memory_base` while the former does not, hence the `undefined-symbols.txt` mismatch.

This wasn't an issue earlier because the `check-symbols` target only looked at .a and .o files that were built without `-fPIC`, but now we're building `crt1-*.o` with `-fPIC`, leading to a slightly different list of undefined symbols depending on the compiler used.

Any suggestions to address this?

sbc100:
> Okay, so CI's failing again, and I think it's because of my most recent change to build `crt1-*.o` with `-fPIC`. It seems that LLVM 10 generates different code than the patched version of LLVM 16 that I've been using to test. Specifically, the latter produces code that references `__memory_base` while the former does not, hence the `undefined-symbols.txt` mismatch.
> 
> This wasn't an issue earlier because the `check-symbols` target only looked at .a and .o files that were built without `-fPIC`, but now we're building `crt1-*.o` with `-fPIC`, leading to a slightly different list of undefined symbols depending on the compiler used.
> 
> Any suggestions to address this?

We normally solve these kinds of problems by excluded certain symbols from the comparison.  

See the comment down in the check-symbols target: https://github.com/WebAssembly/wasi-libc/blob/9f51a7102085ec6a6ced5778f0864c9af9f50000/Makefile#L641-L642

libc-bottom-half/crt/crt1-command.c
```diff
@@ -3,10 +3,63 @@
...
+    }
+
+    // Add 1 for the NULL pointer to mark the end, and check for overflow.
+    size_t num_ptrs = argc + 1;
+    if (num_ptrs == 0) {
+        _Exit(EX_SOFTWARE);
+    }
+
+    // Allocate memory for storing the argument chars.
+    char *argv_buf = malloc(argv_buf_size);
```
yamt, line 37: does this mean to rely on `--gc-sections`? i'm not sure if it's a good idea.

dicej: Sorry, I'm not sure I understand what you mean.  My understanding is that a command-style app will need to define a either `main` function that takes no arguments or one that takes `argc` and `argv` arguments.  If the former, this `__main_void` will not be linked in at all (since it's a weak symbol).  If the latter, this function will (and should) be used.  Under what circumstances would `--gc-sections` affect the result?

sbc100: I'm not sure I understand either, but regardless, `--gc-sections` is the default in wasm-ld.  

Furthermore, `-ffunctions-sections` and `-fdata-sections` are also enabled by default in the llvm backend, and actually there is no way to disable `-ffunctions-sections`.

yamt: i meant it could break `--no-gc-sections`.

dicej: @yamt Can you provide an example of a case where this would break `--no-gc-sections`?  I'm still having trouble seeing the problem.

sbc100: I'm also curious if you have a use case for `-Wl,--no-gc-sections`.  I'm not saying we should keep it working, but I wonder if you have a use case in mind for it.  I'm struggling to think of one.

libc-bottom-half/crt/crt1-command.c
```diff
@@ -3,10 +3,63 @@
...
+        free(argv_buf);
+        _Exit(EX_SOFTWARE);
+    }
+
+    // Fill the argument chars, and the argv array with pointers into those chars.
+    // TODO: Remove the casts on `argv_ptrs` and `argv_buf` once the witx is updated with char8 support.
+    err = __wasi_args_get((uint8_t **)argv, (uint8_t *)argv_buf);
+    if (err != __WASI_ERRNO_SUCCESS) {
+        free(argv_buf);
+        free(argv);
```
sbc100, line 55: Unrelated to this change but I wonder why we don't use `alloca` here to avoid pulling in malloc?  

The emscripten version does this: https://github.com/emscripten-core/emscripten/blob/5ab750c822e47b0878e939f32d0f297edf632cf0/system/lib/standalone/__main_void.c#L41

libc-bottom-half/crt/crt1-command.c
yamt: > @yamt Can you provide an example of a case where this would break `--no-gc-sections`? I'm still having trouble seeing the problem.

just link a module with `--no-gc-sections`?
```
spacetanuki% cat a.c
int main() {return 0;}
spacetanuki% /opt/wasi-sdk/bin/clang --sysroot=${HOME}/git/wasi-libc/sysroot -Xlinker --no-gc-sections a.c
wasm-ld: error: /Users/yamamoto/git/wasi-libc/sysroot/lib/wasm32-wasi/crt1-command.o: undefined symbol: main
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
spacetanuki%
```

yamt: > I'm also curious if you have a use case for `-Wl,--no-gc-sections`. I'm not saying we should keep it working, but I wonder if you have a use case in mind for it. I'm struggling to think of one.

i don't have concrete use cases either.
but i can imagine there can be "creative" users. like embedding a magic signature in a module, which is otherwise unused. especially for apps which are ported from non-wasm environment.

dicej: > just link a module with `--no-gc-sections`?

Thanks.  I still don't understand _why_ `--no-gc-sections` makes a difference here, but clearly it does.

sbc100: Can you explain @yamt ?  What made you suspect this would be an issue with `--no-gc-sections`?   Or as it simply via experimentation that you discovered this issue?  

Normally `--no-gc-sections` should not effect whether a linker succeeds or not.. so this could be rare case.. or a linker bug.   Nothing jumps out at me here as to why this would cause that undefined symbol error, but I have yet to try to repro this locally.

sbc100: Confirmed.. this is behaviour also occurs with GNU ld, so we can't have this symbol live in the same compilation unit as `_start`

sbc100: Maybe we can put this symbol into compiler-rt, which I assume will remain statically linked.   For gcc it looks like the compiler rt is split into a shared part (libgcc_s) and a non-shared part (libgcc)

dicej: Moving it to compiler-rt would require an LLVM change, right?  If so, I'm assuming it's too late to get that into LLVM 17, and I'd prefer not to wait another 6 months to get everything aligned.  Perhaps we should just move it back to main_void.c (where it was originally) for the time being?  Alternatively, we could patch LLVM in wasi-sdk while we're waiting for LLVM 18.

sbc100: In wasi-sdk we can choose what we package into our compiler-rt library.. I don't see why we can't add to it.

Either that or we could create some other object that clang can link in (e.g. crt0.o or crtend.o).. but that would also require a clang driver change.

dicej: @sbc100 Please check my understanding of your first suggestion above:

1. We build `libclang_rt.builtins-wasm32.a` as usual in `wasi-sdk`
2. We also build `__main_void.o` from `__main_void.c` (which would contain the `__main_void` definition we've been discussing).
3. We add `__main_void.o` to `libclang_rt.builtins-wasm32.a`

Does that sound right?  If so, any objection to putting `__main_void.c` back where it was in `libc-bottom-half/sources` but omitting it from libc, using it only as described above?

sbc100: That is certainly an option... I can see an argument though for wanting to keep libclang_rt.builtins-wasm32.a pure.  Are there any other examples of files that we don't want to in the shared library but make more sense linking statically?

dicej: > That is certainly an option... I can see an argument though for wanting to keep libclang_rt.builtins-wasm32.a pure. Are there any other examples of files that we don't want to in the shared library but make more sense linking statically?

Nothing else has come up to my knowledge.

BTW, having `__main_void` in the shared library is not really a _problem_ -- just slightly awkward since it won't be used for reactors.  As long as `__main_argc_argv` is declared weak (which is what started this whole discussion), there's no burden left to the application developer.  So we always have an option of going back to where we started and adding a comment explaining the situation per https://github.com/WebAssembly/wasi-libc/pull/429#discussion_r1290434731.

yamt: i guess that putting back __main_void to libc involves the least effort for now.

dicej: @sbc100 How would you feel if I put `__main_void` back in libc for the time being?  I can add a comment about why the `__main_argc_argv` declaration is weak and that our long-term goal is to move `__main_void` to a crt*.o file that is linked in as appropriate in a `--no-gc-sections`-friendly way.

sbc100: Sure, that sounds fine, given that moving proved problematic.

libc-bottom-half/crt/crt1-command.c
TerrorJack: iirc the default stack size set by `wasm-ld` is 65536 which is exceptionally small, so using that for `argv_buf` sounds pretty risky.

sbc100: I don't think the fact that we have a small default stack size should be used as an argument for not using the stack in places where is makes sense (like this one).

Folks who want to accept long argv lists can easily link with `-Wl,-zstack-size=XX`.    I think the downsize of linking in a malloc implementation into all programs high enough to warrant making this change.

sbc100: Are you suggesting some kind of auto-detection of the need to real malloc vs a bump allocator?   Or some kind of opt-in for folks who think they only need a bump allocator?  (e.g. `-lfake-malloc`?)

sbc100: I guess we could do it with weak symbols.

Code like this could call `maybe_malloc` and `maybe_free`, which are provides weakly by `bumpmalloc.c` but strongly as aliases of `malloc/free` in `dlmalloc.c`?   That might work... but would require those aliases to be added to all malloc/free providers (dlmalloc being one of many)

sunfishcode: I was thinking more like just `-lfake-malloc`.

sbc100: I think that would be unfortunate since it would result is a "bloated-by-default" outcome for small programs.

sbc100: On the other hand many such small programs (.e.g hello world) are not that important since they don't represent real world uses?

sunfishcode: I expect most real-world programs call `malloc` somewhere anyway. The other side of it is, if we had more people maintaining wasi-libc, more things would become possible.

dicej:
Any other concerns about this PR?  I believe I've addressed everything raised so far.

libc-top-half/musl/src/internal/locale_impl.h
sbc100: I think this warrents and `#ifdef  __wasilibc_unmodified_upstream` along with a comment.

This is another reason why splitting libc into multiple libraries is maybe not the best idea.

sbc100 (commented):
lgtm % a couple of final comments

That extra complexity we carry from splitting up libc in to chucks for different `EMULATED_XX` parts makes me a little sad but that is the directions we have chosen for now. 

I do wonder if we should reconsider that and just emulate everything in a single libc.a/libc.so.   How much value are we getting from bring stingy with emulation support by default?  (I don't think we need to let that questions/discussion block that landing of this PR)

xphoenix:
What is the status of this work?

dicej:
> What is the status of this work?

It's ready to go AFAICT; just waiting for an approval from a maintainer.  @sunfishcode and/or any other maintainers: any further concerns before this can be merged?

sbc100 (approved):

PR #435, opened 2023-08-30
Update thread id validation returned by `__wasi_thread_spawn`

loganek:
According to the documentation: https://github.com/WebAssembly/wasi-threads#design-choice-thread-ids, TID should be in the range <1, 0x1FFFFFFF>

### DIFF
--- a/libc-top-half/musl/src/thread/pthread_create.c
+++ b/libc-top-half/musl/src/thread/pthread_create.c
@@ -13,6 +13,7 @@
 #endif
 
 #include <stdalign.h>
+#include <assert.h>
 
 static void dummy_0()
 {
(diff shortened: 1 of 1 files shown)


libc-top-half/musl/src/thread/pthread_create.c
```diff
@@ -558,14 +558,18 @@ int __pthread_create(pthread_t *restrict res, const pthread_attr_t *restrict att
 			__wait(&args->control, 0, 3, 0);
 	}
 #else
+#define WASI_THREADS_MIN_TID 1
+#define WASI_THREADS_MAX_TID 0x1FFFFFFF
 	/* `wasi_thread_spawn` will either return a host-provided thread ID (TID)
-	 * (`>= 0`) or an error code (`< 0`). As in the unmodified version, all
+	 * (`<1, 0x1FFFFFFF>`) or an error code (`< 0`). Please note that `0` is
+	 * reserved for compatibility reasons and must not be returned by the runtime
+	 * (if that happens, EAGAIN is returned). As in the unmodified version, all
```
yamt, line 566: why EAGAIN?
an assertion is more appropriate.

yamt (changes requested):

libc-top-half/musl/src/thread/pthread_create.c
abrown: Yeah, it might be better to crash to force people to go fix the engine.

loganek: fair enough, updated.

abrown (approved):

PR #443, opened 2023-10-20
add stubs for dlopen, dlsym, etc.

dicej:
This adds weak exports for the POSIX `dlopen`, `dlsym`, `dlclose`, and `dlerror` functions, allowing code which uses those features to compile.  The implementations are stubs which always fail since there is currently no official standard for runtime dynamic linking.

Since the symbols are weak, they can be overriden with useful, runtime-specific implementations, e.g. based on host functions or statically-generated tables (see https://github.com/dicej/component-linking-demo for an example of the latter).

### DIFF
--- a/expected/wasm32-wasi-threads/predefined-macros.txt
+++ b/expected/wasm32-wasi-threads/predefined-macros.txt
@@ -1500,14 +1500,13 @@
 #define RRQ 01
 #define RS_HIPRI 0x01
 #define RTLD_DEFAULT ((void *)0)
-#define RTLD_DI_LINKMAP 2
-#define RTLD_GLOBAL 256
-#define RTLD_LAZY 1
-#define RTLD_LOCAL 0
+#define RTLD_GLOBAL 0x8
+#define RTLD_LAZY 0x1
+#define RTLD_LOCAL 0x4
 #define RTLD_NEXT ((void *)-1)
-#define RTLD_NODELETE 4096
-#define RTLD_NOLOAD 4
-#define RTLD_NOW 2
+#define RTLD_NODELETE 0x80
+#define RTLD_NOLOAD 0x10
+#define RTLD_NOW 0x2
 #define RUSAGE_CHILDREN 2
 #define RUSAGE_SELF 1
 #define R_OK (4)
--- a/expected/wasm32-wasi-threads/predefined-macros.txt
+++ b/expected/wasm32-wasi-threads/predefined-macros.txt
@@ -1500,13 +1500,13 @@
 #define RRQ 01
 #define RS_HIPRI 0x01
 #define RTLD_DEFAULT ((void *)0)
-#define RTLD_GLOBAL 0x8
-#define RTLD_LAZY 0x1
-#define RTLD_LOCAL 0x4
+#define RTLD_GLOBAL 256
+#define RTLD_LAZY 1
+#define RTLD_LOCAL 8
 #define RTLD_NEXT ((void *)-1)
-#define RTLD_NODELETE 0x80
-#define RTLD_NOLOAD 0x10
-#define RTLD_NOW 0x2
+#define RTLD_NODELETE 4096
+#define RTLD_NOLOAD 4
+#define RTLD_NOW 2
 #define RUSAGE_CHILDREN 2
 #define RUSAGE_SELF 1
 #define R_OK (4)
--- a/expected/wasm32-wasi-threads/predefined-macros.txt
+++ b/expected/wasm32-wasi-threads/predefined-macros.txt
@@ -1499,6 +1499,15 @@
 #define RRFIXEDSZ NS_RRFIXEDSZ
 #define RRQ 01
 #define RS_HIPRI 0x01
+#define RTLD_DEFAULT ((void *)0)
+#define RTLD_DI_LINKMAP 2
+#define RTLD_GLOBAL 256
+#define RTLD_LAZY 1
+#define RTLD_LOCAL 0
+#define RTLD_NEXT ((void *)-1)
+#define RTLD_NODELETE 4096
+#define RTLD_NOLOAD 4
+#define RTLD_NOW 2
 #define RUSAGE_CHILDREN 2
 #define RUSAGE_SELF 1
 #define R_OK (4)
@@ -2044,6 +2053,7 @@
 #define _Complex_I (0.0f+1.0fi)
 #define _DIRENT_H 
 #define _DIRENT_HAVE_D_TYPE 
+#define _DLFCN_H
 #define _ENDIAN_H 
 #define _ERRNO_H 
 #define _ERR_H 
--- a/libc-top-half/musl/include/dlfcn.h
+++ b/libc-top-half/musl/include/dlfcn.h
@@ -7,24 +7,40 @@ extern "C" {
 
 #include <features.h>
 
+#ifdef __wasilibc_unmodified_upstream
 #define RTLD_LAZY   1
 #define RTLD_NOW    2
 #define RTLD_NOLOAD 4
 #define RTLD_NODELETE 4096
 #define RTLD_GLOBAL 256
 #define RTLD_LOCAL  0
+#else
+/* For WASI, we use flag values which match MacOS rather than musl.  This gives
+ * `RTLD_LOCAL` a non-zero value, avoiding ambiguity and allowing us to defer
+ * the decision of whether `RTLD_LOCAL` or `RTLD_GLOBAL` should be the default
+ * when neither is specified.
+ */
+#define RTLD_LAZY     0x1
+#define RTLD_NOW      0x2
+#define RTLD_LOCAL    0x4
+#define RTLD_GLOBAL   0x8
+#define RTLD_NOLOAD   0x10
+#define RTLD_NODELETE 0x80
+#endif
 
 #define RTLD_NEXT    ((void *)-1)
 #define RTLD_DEFAULT ((void *)0)
 
+#ifdef __wasilibc_unmodified_upstream
 #define RTLD_DI_LINKMAP 2
+#endif
 
 int    dlclose(void *);
 char  *dlerror(void);
 void  *dlopen(const char *, int);
 void  *dlsym(void *__restrict, const char *__restrict);
 
-#if defined(_GNU_SOURCE) || defined(_BSD_SOURCE)
+#if defined(__wasilibc_unmodified_upstream) && (defined(_GNU_SOURCE) || defined(_BSD_SOURCE))
 typedef struct {
 	const char *dli_fname;
 	void *dli_fbase;
--- a/libc-top-half/musl/include/dlfcn.h
+++ b/libc-top-half/musl/include/dlfcn.h
@@ -7,25 +7,19 @@ extern "C" {
 
 #include <features.h>
 
-#ifdef __wasilibc_unmodified_upstream
 #define RTLD_LAZY   1
 #define RTLD_NOW    2
 #define RTLD_NOLOAD 4
 #define RTLD_NODELETE 4096
 #define RTLD_GLOBAL 256
+#ifdef __wasilibc_unmodified_upstream
 #define RTLD_LOCAL  0
 #else
-/* For WASI, we use flag values which match MacOS rather than musl.  This gives
- * `RTLD_LOCAL` a non-zero value, avoiding ambiguity and allowing us to defer
- * the decision of whether `RTLD_LOCAL` or `RTLD_GLOBAL` should be the default
- * when neither is specified.
+/* For WASI, we give `RTLD_LOCAL` a non-zero value, avoiding ambiguity and
+ * allowing us to defer the decision of whether `RTLD_LOCAL` or `RTLD_GLOBAL`
+ * should be the default when neither is specified.
  */
-#define RTLD_LAZY     0x1
-#define RTLD_NOW      0x2
-#define RTLD_LOCAL    0x4
-#define RTLD_GLOBAL   0x8
-#define RTLD_NOLOAD   0x10
-#define RTLD_NODELETE 0x80
+#define RTLD_LOCAL  8
 #endif
 
 #define RTLD_NEXT    ((void *)-1)
--- a/libc-top-half/musl/src/misc/dl.c
+++ b/libc-top-half/musl/src/misc/dl.c
@@ -14,9 +14,10 @@
  * library resolution cannot or should not be supported (and the application can
  * handle this situation gracefully). */
 
+#include <stddef.h>
 #include <dlfcn.h>
 
-static const char *error = 0;
+static const char *error = NULL;
 
 weak int dlclose(void *library)
 {
@@ -27,18 +28,18 @@ weak int dlclose(void *library)
 weak char *dlerror(void)
 {
 	const char *var = error;
-	error = 0;
+	error = NULL;
 	return (char*) var;
 }
 
 weak void *dlopen(const char *name, int flags)
 {
 	error = "dlopen not implemented";
-	return 0;
+	return NULL;
 }
 
 weak void *dlsym(void *library, const char *name)
 {
 	error = "dlsym not implemented";
-	return 0;
+	return NULL;
 }
--- a/libc-top-half/musl/src/misc/dl.c
+++ b/libc-top-half/musl/src/misc/dl.c
@@ -1,31 +1,44 @@
-#include <stdint.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
-#include <dlfcn.h>
+/* This file is used to build libdl.so with stub versions of `dlopen`, `dlsym`,
+ * etc.  The intention is that this stubbed libdl.so can be used to build
+ * libraries and applications which use `dlopen` without committing to a
+ * specific runtime implementation.  Later, it can be replaced with a real,
+ * working libdl.so (e.g. at runtime or component composition time).
+ * 
+ * For example, the `wasm-tools component link` subcommand can be used to create
+ * a component that bundles any `dlopen`-able libraries in such a way that their
+ * function exports can be resolved symbolically at runtime using an
+ * implementation of libdl.so designed for that purpose.  In other cases, a
+ * runtime might provide Emscripten-style dynamic linking via URLs or else a
+ * more traditional, filesystem-based implementation.  Finally, even this
+ * stubbed version of libdl.so can be used at runtime in cases where dynamic
+ * library resolution cannot or should not be supported (and the application can
+ * handle this situation gracefully). */
 
-#define EXPORT __attribute__((visibility("default"))) __attribute__((used))
-#define WEAK __attribute__((__weak__))
+#include <dlfcn.h>
 
 static const char *error = 0;
 
-EXPORT WEAK int dlclose(void *library) {
-    error = "dlclose not implemented";
-    return -1;
+weak int dlclose(void *library)
+{
+	error = "dlclose not implemented";
+	return -1;
 }
 
-EXPORT WEAK char *dlerror(void) {
-    const char *var = error;
-    error = 0;
-    return (char*) var;
+weak char *dlerror(void)
+{
+	const char *var = error;
+	error = 0;
+	return (char*) var;
 }
 
-EXPORT WEAK void *dlopen(const char *name, int flags) {
-    error = "dlopen not implemented";
-    return 0;
+weak void *dlopen(const char *name, int flags)
+{
+	error = "dlopen not implemented";
+	return 0;
 }
 
-EXPORT WEAK void *dlsym(void *library, const char *name) {
-    error = "dlsym not implemented";
-    return 0;
+weak void *dlsym(void *library, const char *name)
+{
+	error = "dlsym not implemented";
+	return 0;
 }
--- a/libc-top-half/musl/src/misc/dl.c
+++ b/libc-top-half/musl/src/misc/dl.c
@@ -1,0 +1,31 @@
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <dlfcn.h>
+
+#define EXPORT __attribute__((visibility("default"))) __attribute__((used))
+#define WEAK __attribute__((__weak__))
+
+static const char *error = 0;
+
+EXPORT WEAK int dlclose(void *library) {
+    error = "dlclose not implemented";
+    return -1;
+}
+
+EXPORT WEAK char *dlerror(void) {
+    const char *var = error;
+    error = 0;
+    return (char*) var;
+}
+
+EXPORT WEAK void *dlopen(const char *name, int flags) {
+    error = "dlopen not implemented";
+    return 0;
+}
+
+EXPORT WEAK void *dlsym(void *library, const char *name) {
+    error = "dlsym not implemented";
+    return 0;
+}
--- a/Makefile
+++ b/Makefile
@@ -80,6 +80,7 @@ LIBWASI_EMULATED_SIGNAL_SOURCES = \
 LIBWASI_EMULATED_SIGNAL_MUSL_SOURCES = \
     $(LIBC_TOP_HALF_MUSL_SRC_DIR)/signal/psignal.c \
     $(LIBC_TOP_HALF_MUSL_SRC_DIR)/string/strsignal.c
+LIBDL_SOURCES = $(LIBC_TOP_HALF_MUSL_SRC_DIR)/misc/dl.c
 LIBC_BOTTOM_HALF_CRT_SOURCES = $(wildcard $(LIBC_BOTTOM_HALF_DIR)/crt/*.c)
 LIBC_TOP_HALF_DIR = libc-top-half
 LIBC_TOP_HALF_MUSL_DIR = $(LIBC_TOP_HALF_DIR)/musl
@@ -90,7 +91,6 @@ LIBC_TOP_HALF_MUSL_SOURCES = \
         misc/a64l.c \
         misc/basename.c \
         misc/dirname.c \
-        misc/dl.c \
         misc/ffs.c \
         misc/ffsl.c \
         misc/ffsll.c \
@@ -380,6 +380,7 @@ LIBWASI_EMULATED_PROCESS_CLOCKS_OBJS = $(call objs,$(LIBWASI_EMULATED_PROCESS_CL
 LIBWASI_EMULATED_GETPID_OBJS = $(call objs,$(LIBWASI_EMULATED_GETPID_SOURCES))
 LIBWASI_EMULATED_SIGNAL_OBJS = $(call objs,$(LIBWASI_EMULATED_SIGNAL_SOURCES))
 LIBWASI_EMULATED_SIGNAL_MUSL_OBJS = $(call objs,$(LIBWASI_EMULATED_SIGNAL_MUSL_SOURCES))
+LIBDL_OBJS = $(call objs,$(LIBDL_SOURCES))
 LIBC_BOTTOM_HALF_CRT_OBJS = $(call objs,$(LIBC_BOTTOM_HALF_CRT_SOURCES))
 
 # These variables describe the locations of various files and
@@ -481,6 +482,7 @@ LIBWASI_EMULATED_PROCESS_CLOCKS_SO_OBJS = $(patsubst %.o,%.pic.o,$(LIBWASI_EMULA
 LIBWASI_EMULATED_GETPID_SO_OBJS = $(patsubst %.o,%.pic.o,$(LIBWASI_EMULATED_GETPID_OBJS))
 LIBWASI_EMULATED_SIGNAL_SO_OBJS = $(patsubst %.o,%.pic.o,$(LIBWASI_EMULATED_SIGNAL_OBJS))
 LIBWASI_EMULATED_SIGNAL_MUSL_SO_OBJS = $(patsubst %.o,%.pic.o,$(LIBWASI_EMULATED_SIGNAL_MUSL_OBJS))
+LIBDL_SO_OBJS = $(patsubst %.o,%.pic.o,$(LIBDL_OBJS))
 BULK_MEMORY_SO_OBJS = $(patsubst %.o,%.pic.o,$(BULK_MEMORY_OBJS))
 DLMALLOC_SO_OBJS = $(patsubst %.o,%.pic.o,$(DLMALLOC_OBJS))
 LIBC_BOTTOM_HALF_ALL_SO_OBJS = $(patsubst %.o,%.pic.o,$(LIBC_BOTTOM_HALF_ALL_OBJS))
@@ -521,6 +523,8 @@ $(OBJDIR)/libwasi-emulated-getpid.so.a: $(LIBWASI_EMULATED_GETPID_SO_OBJS)
 
 $(OBJDIR)/libwasi-emulated-signal.so.a: $(LIBWASI_EMULATED_SIGNAL_SO_OBJS) $(LIBWASI_EMULATED_SIGNAL_MUSL_SO_OBJS)
 
+$(OBJDIR)/libdl.so.a: $(LIBDL_SO_OBJS)
+
 $(SYSROOT_LIB)/libc.a: $(LIBC_OBJS)
 
 $(SYSROOT_LIB)/libc-printscan-long-double.a: $(MUSL_PRINTSCAN_LONG_DOUBLE_OBJS)
@@ -534,6 +538,8 @@ $(SYSROOT_LIB)/libwasi-emulated-process-clocks.a: $(LIBWASI_EMULATED_PROCESS_CLO
 $(SYSROOT_LIB)/libwasi-emulated-getpid.a: $(LIBWASI_EMULATED_GETPID_OBJS)
 
 $(SYSROOT_LIB)/libwasi-emulated-signal.a: $(LIBWASI_EMULATED_SIGNAL_OBJS) $(LIBWASI_EMULATED_SIGNAL_MUSL_OBJS)
+
+$(SYSROOT_LIB)/libdl.a: $(LIBDL_OBJS)
 
 %.a:
 	@mkdir -p "$(@D)"
@@ -603,7 +609,7 @@ startup_files $(LIBC_BOTTOM_HALF_ALL_OBJS) $(LIBC_BOTTOM_HALF_ALL_SO_OBJS): CFLA
     -I$(LIBC_TOP_HALF_MUSL_SRC_DIR)/include \
     -I$(LIBC_TOP_HALF_MUSL_SRC_DIR)/internal
 
-$(LIBC_TOP_HALF_ALL_OBJS) $(LIBC_TOP_HALF_ALL_SO_OBJS) $(MUSL_PRINTSCAN_LONG_DOUBLE_OBJS) $(MUSL_PRINTSCAN_LONG_DOUBLE_SO_OBJS) $(MUSL_PRINTSCAN_NO_FLOATING_POINT_OBJS) $(LIBWASI_EMULATED_SIGNAL_MUSL_OBJS) $(LIBWASI_EMULATED_SIGNAL_MUSL_SO_OBJS): CFLAGS += \
+$(LIBC_TOP_HALF_ALL_OBJS) $(LIBC_TOP_HALF_ALL_SO_OBJS) $(MUSL_PRINTSCAN_LONG_DOUBLE_OBJS) $(MUSL_PRINTSCAN_LONG_DOUBLE_SO_OBJS) $(MUSL_PRINTSCAN_NO_FLOATING_POINT_OBJS) $(LIBWASI_EMULATED_SIGNAL_MUSL_OBJS) $(LIBWASI_EMULATED_SIGNAL_MUSL_SO_OBJS) $(LIBDL_OBJS) $(LIBDL_SO_OBJS): CFLAGS += \
     -I$(LIBC_TOP_HALF_MUSL_SRC_DIR)/include \
     -I$(LIBC_TOP_HALF_MUSL_SRC_DIR)/internal \
     -I$(LIBC_TOP_HALF_MUSL_DIR)/arch/wasm32 \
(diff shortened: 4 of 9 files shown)


abrown (commented):
I'm also fine merging this as-is, but wanted to get some feedback on the following: it feels like this pattern, "provide some maybe-working, weak symbols," is what the `libwasi-emulated-*.a` libraries are for. @sunfishcode, @sbc100: any opinions on this?

sbc100:
In order to override the weak symbols one would need to provide strong symbols, which already works today doesn't it (i.e. without this change).  What is the advantage of having these failing weak symbols?

dicej:
> In order to override the weak symbols one would need to provide strong symbols, which already works today doesn't it (i.e. without this change). What is the advantage of having these failing weak symbols?

(Sorry for being slow to reply; busy week)

The main advantage is being able to compile projects that use `dlopen`, `dlsym`, etc. using `wasi-sdk` without adding CFLAGS or LDFLAGS to the build to point to additional header files and libraries beyond with `wasi-libc` provides.

sbc100:
> > In order to override the weak symbols one would need to provide strong symbols, which already works today doesn't it (i.e. without this change). What is the advantage of having these failing weak symbols?
> 
> (Sorry for being slow to reply; busy week)
> 
> The main advantage is being able to compile projects that use `dlopen`, `dlsym`, etc. using `wasi-sdk` without adding CFLAGS or LDFLAGS to the build to point to additional header files and libraries beyond what `wasi-libc` provides.

I would imagine that most projects that use `dlopen`/`dlsym` have a code path that disables them?  e.g. `HAVE_DLOPEN`.  Wouldn't it be better to use that than to link against stub implementations? 

Do you have a examples of a codebases that would benefit from these stubs?  i.e. Do not have a way to disable the `dlopen` code path yet gracefully handle the failure of `dlopen`?

dicej:
> > The main advantage is being able to compile projects that use `dlopen`, `dlsym`, etc. using `wasi-sdk` without adding CFLAGS or LDFLAGS to the build to point to additional header files and libraries beyond what `wasi-libc` provides.
> 
> I would imagine that most projects that use `dlopen`/`dlsym` have a code path that disables them? e.g. `HAVE_DLOPEN`. Wouldn't it be better to use that than to link against stub implementations?
> 
> Do you have a examples of a codebases that would benefit from these stubs? i.e. Do not have a way to disable the `dlopen` code path yet gracefully handle the failure of `dlopen`?

Ah, but I don't want to disable them.  I've actually forked CPython to enable `dlopen` for WASI, and I'd like to upstream that change.  I very much want CPython to use `dlopen`, and in fact it's the whole reason for the `wasi-sdk` and `wasi-libc` PRs I've opened to date.  Also, it doesn't need to handle the failure of `dlopen` because it won't fail -- I'll provide the correct, working implementation when linking with `wasm-tools link` and the libc version will never actually be used.

Beyond CPython, I want to be able to compile any library that uses `dlopen` for WASI and use `wasm-tools link` to provide the appropriate implementation so it works at runtime.  Alternatively, yamt mentioned an implementation that uses host functions to handle linking: https://github.com/yamt/toywasm/tree/master/examples/libdl.

Granted, if you don't have a working `dlopen` implementation in mind (such as one of the above), you probably _do_ want to disable its use at build time, and of course you'll still be able to do that if the library in question supports it.

sbc100:
> > > The main advantage is being able to compile projects that use `dlopen`, `dlsym`, etc. using `wasi-sdk` without adding CFLAGS or LDFLAGS to the build to point to additional header files and libraries beyond what `wasi-libc` provides.
> > 
> > 
> > I would imagine that most projects that use `dlopen`/`dlsym` have a code path that disables them? e.g. `HAVE_DLOPEN`. Wouldn't it be better to use that than to link against stub implementations?
> > Do you have a examples of a codebases that would benefit from these stubs? i.e. Do not have a way to disable the `dlopen` code path yet gracefully handle the failure of `dlopen`?
> 
> Ah, but I don't want to disable them. I've actually forked CPython to enable `dlopen` for WASI, and I'd like to upstream that change. I very much want CPython to use `dlopen`, and in fact it's the whole reason for the `wasi-sdk` and `wasi-libc` PRs I've opened to date. Also, it doesn't need to handle the failure of `dlopen` because it won't fail -- I'll provide the correct, working implementation when linking with `wasm-tools component link` and the libc version will never actually be used.

Sorry I must be misunderstanding something.   If you are going to provide a working implementation at link time I don't see why this change is useful.  Can't you already do that without this change?

sbc100:
> > > > The main advantage is being able to compile projects that use `dlopen`, `dlsym`, etc. using `wasi-sdk` without adding CFLAGS or LDFLAGS to the build to point to additional header files and libraries beyond what `wasi-libc` provides.
> > > 
> > > 
> > > I would imagine that most projects that use `dlopen`/`dlsym` have a code path that disables them? e.g. `HAVE_DLOPEN`. Wouldn't it be better to use that than to link against stub implementations?
> > > Do you have a examples of a codebases that would benefit from these stubs? i.e. Do not have a way to disable the `dlopen` code path yet gracefully handle the failure of `dlopen`?
> > 
> > 
> > Ah, but I don't want to disable them. I've actually forked CPython to enable `dlopen` for WASI, and I'd like to upstream that change. I very much want CPython to use `dlopen`, and in fact it's the whole reason for the `wasi-sdk` and `wasi-libc` PRs I've opened to date. Also, it doesn't need to handle the failure of `dlopen` because it won't fail -- I'll provide the correct, working implementation when linking with `wasm-tools component link` and the libc version will never actually be used.
> 
> Sorry I must be misunderstanding something. If you are going to provide a working implementation at link time I don't see why this change is useful. Can't you already do that without this change?

I think I see now.. you are talking about that case where libc itself is a dynamic libraryl.. and you wan to link against one version of libc (that doesn't have dlopen) but then run with version that *does* have it?   I.e. you are going to run against a different version of libc.so to the one that you linked against.

Presumably that also means that this patch is really only useful in the case the libc is dynamically linked?

I'm still not 100% convinced this is that right way forward.. but I think I'm seeing you use case now.  What do others thing about this approach?  @sunfishcode ?

dicej:
Yes, that's correct.  And I'm happy to restrict this to the dynamically linked version of libc if that's feasible.

dicej:
Oops, I should have read more carefully.  I'm actually going to use the exact same libc.so in both cases, but also link in my special libdl-for-shared-everything-linking.so so that its version of dlopen etc. is used instead of libc.so's version.

dicej:
Here's an excerpt from the Makefile I'm using that does this, in case it's helpful:
```makefile
$(BUILD_DIR)/bar.wasm: \
		$(BUILD_DIR)/libbar.so \
		$(BUILD_DIR)/libfoo.so \
		$(BUILD_DIR)/libdl.so \
		$(BUILD_DIR)/libpython3.11.so \
		$(LIBC) \
		$(LIBCXX) \
		$(LIBCXXABI) \
		$(NUMPY_LIBRARIES) \
		$(WASI_ADAPTER)
	$(WASM_TOOLS_CLI) component link \
		--adapt wasi_snapshot_preview1=$(WASI_ADAPTER) \
		--dl-openable $(BUILD_DIR)/libfoo.so \
		$(BUILD_DIR)/libbar.so \
		$(BUILD_DIR)/libdl.so \
		$(BUILD_DIR)/libpython3.11.so \
		$(LIBC) \
		$(LIBCXX) \
		$(LIBCXXABI) \
		$(shell echo $(NUMPY_LIBRARIES) | \
			sed 's_$(BUILD_DIR)/\([^ ]*\)_--dl-openable /build/\1=$(BUILD_DIR)/\1_g') \
		-o $@
```
In this case, libdl.so is customized for how `wasm-tools component link` supports "dynamic linking" to other modules in the component I'm producing.  Note that `libdl.so` comes before `$(LIBC)` in the linking order, so it's version of `dlopen` etc. takes precedence.

expected/wasm32-wasi-threads/predefined-macros.txt
```diff
@@ -1499,6 +1499,15 @@
 #define RRFIXEDSZ NS_RRFIXEDSZ
 #define RRQ 01
 #define RS_HIPRI 0x01
+#define RTLD_DEFAULT ((void *)0)
+#define RTLD_DI_LINKMAP 2
+#define RTLD_GLOBAL 256
+#define RTLD_LAZY 1
+#define RTLD_LOCAL 0
```
yamt, line 1506: isn't it straightforward to use a non-zero value?
an implementation can still choose to ignore it.

expected/wasm32-wasi-threads/predefined-macros.txt
```diff
@@ -1499,6 +1499,15 @@
 #define RRFIXEDSZ NS_RRFIXEDSZ
 #define RRQ 01
 #define RS_HIPRI 0x01
+#define RTLD_DEFAULT ((void *)0)
+#define RTLD_DI_LINKMAP 2
```
yamt, line 1503: does this make sense w/o dlinfo?

yamt:
> Here's an excerpt from the Makefile I'm using that does this, in case it's helpful:
> 
> ```makefile
> $(BUILD_DIR)/bar.wasm: \
> 		$(BUILD_DIR)/libbar.so \
> 		$(BUILD_DIR)/libfoo.so \
> 		$(BUILD_DIR)/libdl.so \
> 		$(BUILD_DIR)/libpython3.11.so \
> 		$(LIBC) \
> 		$(LIBCXX) \
> 		$(LIBCXXABI) \
> 		$(NUMPY_LIBRARIES) \
> 		$(WASI_ADAPTER)
> 	$(WASM_TOOLS_CLI) component link \
> 		--adapt wasi_snapshot_preview1=$(WASI_ADAPTER) \
> 		--dl-openable $(BUILD_DIR)/libfoo.so \
> 		$(BUILD_DIR)/libbar.so \
> 		$(BUILD_DIR)/libdl.so \
> 		$(BUILD_DIR)/libpython3.11.so \
> 		$(LIBC) \
> 		$(LIBCXX) \
> 		$(LIBCXXABI) \
> 		$(shell echo $(NUMPY_LIBRARIES) | \
> 			sed 's_$(BUILD_DIR)/\([^ ]*\)_--dl-openable /build/\1=$(BUILD_DIR)/\1_g') \
> 		-o $@
> ```
> 
> In this case, libdl.so is customized for how `wasm-tools component link` supports "dynamic linking" to other modules in the component I'm producing. Note that `libdl.so` comes before `$(LIBC)` in the linking order, so it's version of `dlopen` etc. takes precedence.

i'm a bit confused. do you mean you don't really need these stubs?

sbc100:
> > Here's an excerpt from the Makefile I'm using that does this, in case it's helpful:
> > ```makefile
> > $(BUILD_DIR)/bar.wasm: \
> > 		$(BUILD_DIR)/libbar.so \
> > 		$(BUILD_DIR)/libfoo.so \
> > 		$(BUILD_DIR)/libdl.so \
> > 		$(BUILD_DIR)/libpython3.11.so \
> > 		$(LIBC) \
> > 		$(LIBCXX) \
> > 		$(LIBCXXABI) \
> > 		$(NUMPY_LIBRARIES) \
> > 		$(WASI_ADAPTER)
> > 	$(WASM_TOOLS_CLI) component link \
> > 		--adapt wasi_snapshot_preview1=$(WASI_ADAPTER) \
> > 		--dl-openable $(BUILD_DIR)/libfoo.so \
> > 		$(BUILD_DIR)/libbar.so \
> > 		$(BUILD_DIR)/libdl.so \
> > 		$(BUILD_DIR)/libpython3.11.so \
> > 		$(LIBC) \
> > 		$(LIBCXX) \
> > 		$(LIBCXXABI) \
> > 		$(shell echo $(NUMPY_LIBRARIES) | \
> > 			sed 's_$(BUILD_DIR)/\([^ ]*\)_--dl-openable /build/\1=$(BUILD_DIR)/\1_g') \
> > 		-o $@
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In this case, libdl.so is customized for how `wasm-tools component link` supports "dynamic linking" to other modules in the component I'm producing. Note that `libdl.so` comes before `$(LIBC)` in the linking order, so it's version of `dlopen` etc. takes precedence.
> 
> i'm a bit confused. do you mean you don't really need these stubs?

IIRC that use case if dynamic linking.  They want link (at build time) against a `.so` contain the stubs but then at runtime link against a library that contains real implementations.

dicej:
It might be helpful to step back and look at the big picture here.  Many (most?) popular high-level languages have some way to do FFI calls to C ABI native code loaded via `dlopen`/`dlsym` or the equivalent.  Java has JNI, Ruby has extension libraries, .NET has DllImport, Python has native extensions, etc.

Python is perhaps the most notable of these since many popular packages rely heavily on native extensions.  Those packages are distributed as platform-specific "wheels", each of which includes the Python code plus one or more shared libraries built for that specific platform.  The wheels are usually published to pypi.org, the default package index and repository used by `pip`, a popular package manager for Python.

The immediate motivation behind https://github.com/WebAssembly/wasi-libc/pull/429 and https://github.com/WebAssembly/wasi-sdk/pull/338 was to lay the foundation for introducing a new WASI platform for Python wheels alongside the existing ones such as MacOS/ARM64 and Windows/x64.  Such wheels would contain shared libraries built using a future release of `wasi-sdk` and be compatible with any build of CPython built using the same version of `wasi-sdk` (which is the best we can currently offer given that `wasi-libc` does not yet make any backwards compatibility guarantees).

In order to actually _use_ the shared libraries in these wheels, CPython needs some way to load them, and the easiest way is to build it with `dlopen`/`dlsym` enabled.  The idea behind this PR is to make that easy to do.  The dlfcn.h header and dl.c stubs would allow a `dlopen`-enabled WASI CPython to be built with a stock, unmodified `wasi-sdk` and `wasi-libc`.  The alternative would be for CPython to provide its own, built-in stubs for these functions.  Likewise, other language runtimes, applications, or libraries which want to use `dlopen`/`dlsym` when targeting WASI would be responsible for providing stubs to compile against.  In the case of a library that doesn't have any special WASI support, the user of that library would be responsible for providing the stubs and injecting them via the appropriate CFLAGS and LDFLAGS when building the library.  I'd like to avoid that putting that burden on third party developers and users, hence this PR.

As I mentioned earlier, the intention here is that the stubs will be replaced (either at runtime or component composition time) with a working implementation appropriate to the environment in which they're used.  One such implementation might load libraries Emscripten-style, via a URL.  Another might search the filesystem.  `wasm-tools component link` can do shared-everything linking to bundle everything into a self-contained component, with optional support for symbol lookup via `dlopen`.  Eventually `wasm-tools component link` will also support creating components that import modules (e.g. from a registry) as an alternative to bundling.

To summarize: my goal is to make it easy to create _and_ use WASI shared libraries -- including via `dlopen` -- in all contexts where that's useful, and Python happens to be one of the more urgent cases.  If there are better ways to achieve that which provides a comparable level of convenience to `wasi-sdk` users, I'm happy to consider them.

yamt:
my impression is that it's a bit tricky to replace only a part of a shared library. (libc.so)
on the other hand, i can understand it's useful to provide a stub implementation here.

how about providing a stub libdl.so as a separate library from libc?
that way, an executable using dlopen can be run more naturally as it has an explicit record in needed_dynlibs. (otherwise you would need an LD_PRELOAD equivalent functionality to load your version of dlopen implementation explicitly.)
as requiring `-ldl` for dlopen is not too uncommon, i guess it isn't a problem for applications like python.

dicej:
> how about providing a stub libdl.so as a separate library from libc? 

Sounds good to me.  Ideally I'd like to include it as part of `wasi-sdk` (i.e. as part of `wasi-libc` or some other dependency of `wasi-sdk`).  Any objections to me adding a `libdl.so` to the `wasi-libc` makefile alongside `libc.so` and `libwasi-emulated-*.so`?  @sbc100 does that sound okay to you?

sbc100:
> > how about providing a stub libdl.so as a separate library from libc?
> 
> Sounds good to me. Ideally I'd like to include it as part of `wasi-sdk` (i.e. as part of `wasi-libc` or some other dependency of `wasi-sdk`). Any objections to me adding a `libdl.so` to the `wasi-libc` makefile alongside `libc.so` and `libwasi-emulated-*.so`? @sbc100 does that sound okay to you?

That does sounds little more logical I guess, since then you replacing the entire libdl.so and not just part of libc.so.

I guess we could even make this a dynamic-only library since there would not be any use the static one.  But that also begs the question, why not just supply your working libdl.so and link time and link against that, which would take wasi-libc out of the picture completely.   I guess one argument would be that there could be other users who could benefit from a shared libdl.so.. not just python.. but I'm not sure we are that stage yet.

I don't feel strongly either way.

dicej:
> I guess we could even make this a dynamic-only library since there would not be any use the static one. But that also begs the question, why not just supply your working libdl.so and link time and link against that, which would take wasi-libc out of the picture completely. I guess one argument would be that there could be other users who could benefit from a shared libdl.so.. not just python.. but I'm not sure we are that stage yet.

Yes, there are three options here with respect to CPython:

1. I can maintain a fork of CPython indefinitely, patched to enable `dlopen`, and either build it with `CFLAGS=-I/path-to-dlfcn-h/` and `LDFLAGS="-L/path-to-libdl-so/ -ldl` or bake those flags into the build system.  Anyone who wants to use this version would need to either build my fork from source or download a pre-built artifact from my GitHub releases.
2. If they're willing, the CPython folks can enable `dlopen` for WASI in the official repo using a built-in stub libdl.so.  Then people would be able to build from the official source or use the official builds (once WASI becomes a tier 2 supported platform, which is what Brett Cannon is working towards).
3. We merge this PR, and the CPython folks adopt the next `wasi-sdk` release and enable `dlopen` for WASI.  In this case the stub is part of `wasi-libc` so CPython need not provide it.  Furthermore, any runtime, app, or library that uses `dlopen` will benefit -- not just CPython.  I'm personally planning to work on bringing Java/JNI and Ruby with extension libraries to WASI, and I'd prefer to avoid copying a stub library around and explaining to the upstream projects why it is necessary.

sbc100:
> 2\. If they're willing, the CPython folks can enable `dlopen` for WASI in the official repo using a built-in stub libdl.so.  Then people would be able to build from the official source or use the official builds (once WASI becomes a tier 2 supported platform, which is what Brett Cannon is working towards).

Does CPython gracefully handle failure of `dlopen` at runtime like this?  i.e. You are telling cpython that dlopen exists but in at runtime it might still be a stub (unless that user happens to interpose a working version like you will).   I wonder if its better to declare/detect dlopen missing at build time for some projects (maybe CPython is one of them?). 

Or are you imagining that all users of CPython+wasi is inject a working version of dlopen at runtime?  (Presumably those who build without dyanmic linking have no way to do this).

dicej:
> Does CPython gracefully handle failure of `dlopen` at runtime like this? i.e. You are telling cpython that dlopen exists but in at runtime it might still be a stub (unless that user happens to interpose a working version like you will). I wonder if its better to declare/detect dlopen missing at build time for some projects (maybe CPython is one of them?).

Yes, it does handle it gracefully -- the same way it would if it couldn't find the native extension library on the filesystem.  I.e. it will behave the same way it does today: trying to load a native extension will fail because it is running in an environment where loading native extensions is not possible.

BTW, keep in mind that 99% of users will *not* be building CPython from source -- they will be using an official, pre-built release, in which case build-time feedback isn't relevant to them.

> Or are you imagining that all users of CPython+wasi is inject a working version of dlopen at runtime? (Presumably those who build without dyanmic linking have no way to do this).

All CPython users who want to use native extensions with CPython+wasi will inject a working version of dlopen, yes (though not necessarily at runtime -- possibly at composition time).  `componentize-py` does this automatically today, and it can be done manually using `wasm-tools component link`.  Other implementations might do it at runtime with an empscripten-style trampoline or with a host function like `toywasm` does.

sbc100:
> > Does CPython gracefully handle failure of `dlopen` at runtime like this? i.e. You are telling cpython that dlopen exists but in at runtime it might still be a stub (unless that user happens to interpose a working version like you will). I wonder if its better to declare/detect dlopen missing at build time for some projects (maybe CPython is one of them?).
> 
> Yes, it does handle it gracefully -- the same way it would if it couldn't find the native extension library on the filesystem. I.e. it will behave the same way it does today: trying to load a native extension will fail because it is running in an environment where loading native extensions is not possible.
> 
> BTW, keep in mind that 99% of users will _not_ be building CPython from source -- they will be using an official, pre-built release, in which case build-time feedback isn't relevant to them.
> 
> > Or are you imagining that all users of CPython+wasi is inject a working version of dlopen at runtime? (Presumably those who build without dyanmic linking have no way to do this).
> 
> All CPython users who want to use native extensions with CPython+wasi will inject a working version of dlopen, yes (though not necessarily at runtime -- possibly at composition time). `componentize-py` does this automatically today, and it can be done manually using `wasm-tools component link`. Other implementations might do it at runtime with an empscripten-style trampoline or with a host function like `toywasm` does.

SGTM.. sounds like a good plan.

Out of interest when you talk about injecting a working version of "dlopen" at "composition time" is that just the "libdl" functions that get linked in, or are you also talking about the shared libraries that are opened by "dlopen" ?  Are they somehow determined ahead of time and kind of pre-linked?  i.e. is your dlopen actually able to load stuff at runtime off the wire?

sbc100:
> > > Does CPython gracefully handle failure of `dlopen` at runtime like this? i.e. You are telling cpython that dlopen exists but in at runtime it might still be a stub (unless that user happens to interpose a working version like you will). I wonder if its better to declare/detect dlopen missing at build time for some projects (maybe CPython is one of them?).
> > 
> > 
> > Yes, it does handle it gracefully -- the same way it would if it couldn't find the native extension library on the filesystem. I.e. it will behave the same way it does today: trying to load a native extension will fail because it is running in an environment where loading native extensions is not possible.
> > BTW, keep in mind that 99% of users will _not_ be building CPython from source -- they will be using an official, pre-built release, in which case build-time feedback isn't relevant to them.
> > > Or are you imagining that all users of CPython+wasi is inject a working version of dlopen at runtime? (Presumably those who build without dyanmic linking have no way to do this).
> > 
> > 
> > All CPython users who want to use native extensions with CPython+wasi will inject a working version of dlopen, yes (though not necessarily at runtime -- possibly at composition time). `componentize-py` does this automatically today, and it can be done manually using `wasm-tools component link`. Other implementations might do it at runtime with an empscripten-style trampoline or with a host function like `toywasm` does.
> 
> SGTM.. sounds like a good plan.
> 
> Out of interest when you talk about injecting a working version of "dlopen" at "composition time" is that just the "libdl" functions that get linked in, or are you also talking about the shared libraries that are opened by "dlopen" ? Are they somehow determined ahead of time and kind of pre-linked? i.e. is your dlopen actually able to load stuff at runtime off the wire?

Off topic now: How about I found out more about `componentize-py` and how it works?

dicej:
> Off topic now: How about I found out more about `componentize-py` and how it works? How does this work relate to the pyodide project?

Sounds great -- here's a talk I did about `componentize-py` at WasmCon: https://wasmcon2023.sched.com/event/1PCMP/introducing-componentize-py-a-tool-for-packaging-python-apps-as-components-joel-dice-fermyon-technologies (video and slides both available on that page).  Happy to do a Zoom call, chat on Discord, or whatever as well.

I've chatted with Hood Chatham in detail about Pyodide, and it's been an inspiration for a lot of what I've done with `componentize-py`.  The main differences are WASI instead of Emscripten, no JS or browser required, and heavy use of the Component Model.

dicej:
> Out of interest when you talk about injecting a working version of "dlopen" at "composition time" is that just the "libdl" functions that get linked in, or are you also talking about the shared libraries that are opened by "dlopen" ? Are they somehow determined ahead of time and kind of pre-linked? i.e. is your dlopen actually able to load stuff at runtime off the wire?

dicej:
> Out of interest when you talk about injecting a working version of "dlopen" at "composition time" is that just the "libdl" functions that get linked in, or are you also talking about the shared libraries that are opened by "dlopen" ? Are they somehow determined ahead of time and kind of pre-linked? i.e. is your dlopen actually able to load stuff at runtime off the wire?

The `wasm-tools component link` step (which `componentize-py` does automatically after searching the dependency tree for native extension libraries) bundles `libdl.so`, `libc.so`, `libc++.so` (if needed by one or more native extensions), `libpython3.11.so`, and all the native extension libraries into a component.  Of those, only the native extensions are made available via `dlopen` (using the `--dl-openable` flag).  That involves statically generating lookup tables which `dlopen` and `dlsym` use at runtime to convert each (library, symbol) name pair into a function pointer.

This implementation of `dlopen` cannot load libraries at runtime -- they all have to be accounted for when the component is created.  That's why I sometimes call it "pseudo-dynamic linking".  However, the Component Model does allow components to import modules, so the actual module used for a given native extension could be resolved immediately prior to instantiating and running the component if desired.  True runtime linking is not in scope at the moment, but it's certainly something we'd like to support eventually.

sbc100:
> > Out of interest when you talk about injecting a working version of "dlopen" at "composition time" is that just the "libdl" functions that get linked in, or are you also talking about the shared libraries that are opened by "dlopen" ? Are they somehow determined ahead of time and kind of pre-linked? i.e. is your dlopen actually able to load stuff at runtime off the wire?
> 
> The `wasm-tools component link` step (which `componentize-py` does automatically after searching the dependency tree for native extension libraries) bundles `libdl.so`, `libc.so`, `libc++.so` (if needed by one or more native extensions), `libpython3.11.so`, and all the native extension libraries into a component. Of those, only the native extensions are made available via `dlopen` (using the `--dl-openable` flag). That involves statically generating lookup tables which `dlopen` and `dlsym` use at runtime to convert each (library, symbol) name pair into a function pointer.
> 
> This implementation of `dlopen` cannot load libraries at runtime -- they all have to be accounted for when the component is created. That's why I sometimes call it "pseudo-dynamic linking". However, the Component Model does allow components to import modules, so the actual module used for a given native extension could be resolved immediately prior to instantiating and running the component if desired. True runtime linking is not in scope at the moment, but it's certainly something we'd like to support eventually.

Ah, yes that is what I think I remembered about this use case.   It seems like that "pseudo-dynamic linking" in this case could probably be optimized by doing that link statically use wasm-ld and building things as `.a` rather than `.so`?   Since it seems like building things `.so` at all here has no advantage over `.a`?   Maybe that can be done later as an optimization?

dicej:
> Ah, yes that is what I think I remembered about this use case. It seems like that "pseudo-dynamic linking" in this case could probably be optimized by doing that link statically use wasm-ld and building things as .a rather than .so? Since it seems like building things .so at all here has no advantage over .a? Maybe that can be done later as an optimization?

.so files have a number of advantages over .a files:

- They can be published as part of packages to e.g. pypi.org using the same build steps and infrastructure that package maintainers are already using to publish e.g. Windows, Mac, and Linux builds.
- The ability to import .so files into components allows you to deploy lightweight components which _reference_ libc.so, libpython3.11.so, any native extension libraries, etc. rather than _include_ them.  E.g. a FaaS host needs only one copy of libpython3.11.so rather than 50 statically-linked copies for 50 apps.
- They don't require a linker or any other native build tooling to use -- `pip install componentize-py` is all you need (although if someone were to package `wasm-ld` as a Python package maybe this wouldn't be a big deal).

sbc100:
> > Ah, yes that is what I think I remembered about this use case. It seems like that "pseudo-dynamic linking" in this case could probably be optimized by doing that link statically use wasm-ld and building things as .a rather than .so? Since it seems like building things .so at all here has no advantage over .a? Maybe that can be done later as an optimization?
> 
> .so files have a number of advantages over .a files:
> 
> * They can be published as part of packages to e.g. pypi.org using the same build steps and infrastructure that package maintainers are already using to publish e.g. Windows, Mac, and Linux builds.
> * The ability to import .so files into components allows you to deploy lightweight components which _reference_ libc.so, libpython3.11.so, any native extension libraries, etc. rather than _include_ them.  E.g. a FaaS host needs only one copy of libpython3.11.so rather than 50 statically-linked copies for 50 apps.  Likewise for `lapack_lite.cpython-311-wasm32-wasi.so` or any of the other dozen or so NumPy libraries.
> * They don't require a linker or any other native build tooling to use -- `pip install componentize-py` is all you need (although if someone were to package `wasm-ld` as a Python package maybe this wouldn't be a big deal).

Good points!    It still might be interesting as to measure how much using `.so` files costs though.  i.e. how much perf is being left on the table, especially given things like LTO.   I don't have a good intuition for how much it might be .. but I would guess is a fair bit.

tschneidereit:
Hey @sbc100 👋🏻 I agree that for use cases where having `.so` files isn't necessary, `.a` files would be preferable, and would probably improve performance quite a bit, at least for very chatty interfaces.

The use cases we're currently looking at unfortunately fundamentally require `.so` files, so we don't have a choice here. Besides the requirements for installed toolchains, as Joel mentions being able to share libraries among a potentially huge number of different applications deployed on the same server is a key reason why static linking is unfortunately just not an option.

Given that, do you think we could get this approach in? If anyone in the future has use cases that lend themselves to static linking, we'd be happy to help facilitate that work, e.g. by providing reviews and support.

libc-top-half/musl/src/misc/dl.c
```diff
@@ -0,0 +1,31 @@
...
+#include <string.h>
+#include <dlfcn.h>
+
+#define EXPORT __attribute__((visibility("default"))) __attribute__((used))
+#define WEAK __attribute__((__weak__))
+
+static const char *error = 0;
+
+EXPORT WEAK int dlclose(void *library) {
+    error = "dlclose not implemented";
```
sbc100, line 13: I guess we should follow musl conventions here which are (for better or worse!) to use tabs for indentation, and to put the opening brace of the function on a line by istelf.

libc-top-half/musl/src/misc/dl.c
```diff
@@ -0,0 +1,31 @@
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <dlfcn.h>
+
+#define EXPORT __attribute__((visibility("default"))) __attribute__((used))
+#define WEAK __attribute__((__weak__))
```
sbc100, line 8: I believe musl already defines the lowercase `weak` macro for this purpose.

libc-top-half/musl/src/misc/dl.c
```diff
@@ -0,0 +1,31 @@
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <dlfcn.h>
+
+#define EXPORT __attribute__((visibility("default"))) __attribute__((used))
```
sbc100, line 7: Why is this needed here but not for other libc symbols?  Maybe its not needed?

libc-top-half/musl/src/misc/dl.c
```diff
@@ -0,0 +1,31 @@
+#include <stdint.h>
```
sbc100, line 1: Perhaps add a comment here about why this file exists?  

Mention that it used to build libdl.so and that the intent is to then link with non-stub versions at runtime?

sbc100:
> Hey @sbc100 👋🏻 I agree that for use cases where having `.so` files isn't necessary, `.a` files would be preferable, and would probably improve performance quite a bit, at least for very chatty interfaces.
> 
> The use cases we're currently looking at unfortunately fundamentally require `.so` files, so we don't have a choice here. Besides the requirements for installed toolchains, as Joel mentions being able to share libraries among a potentially huge number of different applications deployed on the same server is a key reason why static linking is unfortunately just not an option.
> 
> Given that, do you think we could get this approach in? If anyone in the future has use cases that lend themselves to static linking, we'd be happy to help facilitate that work, e.g. by providing reviews and support.

Sure thing, I just have couple more nits.

I think mostly agreed to go with above proposal to putting these stubs in libdl.so, so that still needs to happen.

libc-top-half/musl/src/misc/dl.c
dicej: That's left over from earlier experiments.  I've removed it and also removed the unneeded `#include`s.

sbc100 (approved):

dicej:
@sbc100 thanks for the detailed feedback.  I believe I've addressed everything.

BTW, I noticed that `wasi-libc` was already creating an empty libdl.a, which confused `CPython`'s build system when building libpython3.11.a, so I replaced the empty libdl.a with one that has the same stubs that libdl.so has.  That seemed easier than trying to make CPython understand that `dlopen` and friends are available for the shared library build but not for the static library build.

expected/wasm32-wasi-threads/predefined-macros.txt
yamt: @dicej ping

expected/wasm32-wasi-threads/predefined-macros.txt
yamt: @dicej ping

expected/wasm32-wasi-threads/predefined-macros.txt
dicej: Sorry, I didn't notice these comments earlier.  The values in these files come straight from libc-top-half/musl/include/dlfcn.h, which appears to remain unchanged since it was imported from musl libc.  `glibc` seems to use the same values for these constants also.  I'm not necessarily opposed to changing the values, but sticking with the musl values seems like a safe default here.

expected/wasm32-wasi-threads/predefined-macros.txt
dicej: Yeah, I don't know why this is defined unconditionally in dlfcn.h.  We could move it into the `#if defined(_GNU_SOURCE) || defined(_BSD_SOURCE)` block in that file if desired.

expected/wasm32-wasi-threads/predefined-macros.txt
dicej: Today I learned that `RTLD_GLOBAL` is the default on MacOS, but `RTLD_LOCAL` is the default on Linux (which I guess is why `RTLD_LOCAL` is zero on musl and glibc but nonzero on MacOS).  I guess we ought to decide that the default should be for WASI.

yamt: unless you want to decide the default right now, we can postpone the decision by making it a non-zero value.

dicej: I just pushed an update changing it to a non-zero value.

expected/wasm32-wasi-threads/predefined-macros.txt
dicej: I just pushed an update to address this.

expected/wasm32-wasi-threads/predefined-macros.txt
sbc100: I'm not sure I understand how we can make both of these non-zero.   This is a binary choice between global or local symbol binding, its not possible to choose neither.   The question is what the default should be when neither is specified and the answer has to be one or the other.

I suggest we go with that musl's defaults for now given that we are largely a port of musl here.

libc-top-half/musl/include/dlfcn.h
```diff
@@ -7,24 +7,40 @@ extern "C" {
...
+ * `RTLD_LOCAL` a non-zero value, avoiding ambiguity and allowing us to defer
+ * the decision of whether `RTLD_LOCAL` or `RTLD_GLOBAL` should be the default
+ * when neither is specified.
+ */
+#define RTLD_LAZY     0x1
+#define RTLD_NOW      0x2
+#define RTLD_LOCAL    0x4
+#define RTLD_GLOBAL   0x8
+#define RTLD_NOLOAD   0x10
+#define RTLD_NODELETE 0x80
```
sbc100, line 28: Why re-define all of these just because you want to make RTLD_LOCAL non-zero?

dicej: We don't have to.  I just figured we should try to consistently match _some_ existing system that is already widely used, and MacOS was the first one I found that had a non-zero RTLD_LOCAL.  I don't have any strong feelings about it, though.  Would you prefer I change only RTLD_LOCAL?

sbc100: I think we should always prefer the minimal set of changes from musl... unless there is a good reason not to.

expected/wasm32-wasi-threads/predefined-macros.txt
sbc100: Oh wait, I understand now.  Please ignore my last comment.  Having `RTLD_LOCAL` and `RTLD_GLOBAL` both be non-zero allows the implementation to detect the default case and choose one or the other at runtime.     SGTM

libc-top-half/musl/include/dlfcn.h
dicej: I just pushed another update that uses the musl values except for RTLD_LOCAL.

libc-top-half/musl/src/misc/dl.c
```diff
@@ -0,0 +1,44 @@
...
+ * implementation of libdl.so designed for that purpose.  In other cases, a
+ * runtime might provide Emscripten-style dynamic linking via URLs or else a
+ * more traditional, filesystem-based implementation.  Finally, even this
+ * stubbed version of libdl.so can be used at runtime in cases where dynamic
+ * library resolution cannot or should not be supported (and the application can
+ * handle this situation gracefully). */
+
+#include <dlfcn.h>
+
+static const char *error = 0;
```
yamt, line 19: it's better to use NULL

libc-top-half/musl/src/misc/dl.c
```diff
@@ -0,0 +1,44 @@
...
+weak int dlclose(void *library)
+{
+	error = "dlclose not implemented";
+	return -1;
+}
+
+weak char *dlerror(void)
+{
+	const char *var = error;
+	error = 0;
```
yamt, line 30: ditto

yamt (approved):

abrown (approved):
It looks to me like all the concerns are addressed; I was generally in favor of this from the beginning but it looks like the discussion improved some parts. Thanks @dicej for working through all of this and giving clear explanations! Since there seems to be consensus on the main idea, I'll go ahead and merge this and we can work out any details in future PRs.

PR #445, opened 2023-11-13
ifdef out pthread_cancel

yamt:
as we don't actually support thread cancellation.

note: currently we don't build pthread_cancel.c either. this commit just disables it in our header too to make users notice that it isn't provided a bit earlier.

should we disable other cancellation related functions like pthread_testcancel? maybe. but they are harmless to ignore.

### DIFF
--- a/libc-top-half/musl/include/pthread.h
+++ b/libc-top-half/musl/include/pthread.h
@@ -97,7 +97,9 @@ int pthread_equal(pthread_t, pthread_t);
 int pthread_setcancelstate(int, int *);
 int pthread_setcanceltype(int, int *);
 void pthread_testcancel(void);
+#ifdef __wasilibc_unmodified_upstream /* WASI has no cancellation support. */
 int pthread_cancel(pthread_t);
+#endif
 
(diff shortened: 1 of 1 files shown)


libc-top-half/musl/include/pthread.h
```diff
@@ -97,7 +97,9 @@ int pthread_equal(pthread_t, pthread_t);
 int pthread_setcancelstate(int, int *);
 int pthread_setcanceltype(int, int *);
 void pthread_testcancel(void);
+#ifdef __wasilibc_unmodified_upstream /* WASI has no cancellation support. */
 int pthread_cancel(pthread_t);
+#endif
 
 #ifdef __wasilibc_unmodified_upstream /* WASI has no CPU scheduling support. */
```
sbc100, line 104: Maybe just combine these `#ifdef` block since they are contiguous anyway?

sbc100 (approved):

libc-top-half/musl/include/pthread.h
yamt: i separated them because they have different comments.

abrown (approved):

PR #450, opened 2023-12-13
fix libdl.so symbol visibility

yamt:


yamt:
@dicej

dicej (approved):
Good catch!

abrown (approved):
I was just working on releasing wasi-sdk-21... should this fix be in that release?

yamt:
> I was just working on releasing wasi-sdk-21... should this fix be in that release?

i guess so

PR #460, opened 2024-01-04
add WASI Preview 2 bindings

dicej:
This adds C bindings generated from the `wasi:cli/imports@0.2.0-rc-2023-12-05` world, plus a makefile target to regenerate them from the WIT source files.

We'll use these bindings to call Preview 2 host functions when building for the `wasm32-wasi-preview2` target.


libc-bottom-half/cloudlibc/src/libc/sys/wasi/preview2_component_type.o
sbc100, line 1: What is this object file?

libc-bottom-half/headers/public/wasi/preview2.h
```diff
@@ -0,0 +1,2481 @@
+// Generated by `wit-bindgen` 0.16.0. DO NOT EDIT!
+#ifndef __BINDINGS_IMPORTS_H
+#define __BINDINGS_IMPORTS_H
```
sbc100, line 3: Maybe it would be good if the tool could be passed the name `preview2` so that these lines would make sense, and then we could also drop the `sed` and rename logic in the Makefile?

libc-bottom-half/headers/public/wasi/preview2.h
```diff
@@ -0,0 +1,2481 @@
...
+// until all of these operations are complete, or an error occurs.
+// 
+// This is a convenience wrapper around the use of `check-write`,
+// `subscribe`, `write`, and `flush`, and is implemented with the
+// following pseudo-code:
+// 
+// ```text
+// let pollable = this.subscribe();
+// while !contents.is_empty() {
+  // // Wait for the stream to become writable
```
sbc100, line 1214: The indentation looks like it gets confused here..

libc-bottom-half/headers/public/wasi/preview2.h
```diff
@@ -0,0 +1,2481 @@
...
+    // 2. calling `read` on the `input-stream` with the smaller of the
+    // `check-write` permitted length and the `len` provided to `splice`
+    // 3. calling `write` on the `output-stream` with that read data.
+    // 
+    // Any error reported by the call to `check-write`, `read`, or
+    // `write` ends the splice and reports that error.
+    // 
+    // This function returns the number of bytes transferred; it may be less
+    // than `len`.
+    extern bool streams_method_output_stream_splice(streams_borrow_output_stream_t self, streams_borrow_input_stream_t src, uint64_t len, uint64_t *ret, streams_stream_error_t *err);
```
sbc100, line 1299: I assume all this stuff is supposed to be in column zero?

libc-bottom-half/headers/public/wasi/preview2.h
dicej: Yeah, `wit-bindgen` has a very primitive indentation algorithm which is easily fooled, unfortunately.   Perhaps running the output through `clang-format` would be enough to fix the damage.

libc-bottom-half/cloudlibc/src/libc/sys/wasi/preview2_component_type.o
dicej: It contains the custom section which `wit-component` uses to reconstruct the high-level used by the application.  It's essentially a binary representation of the high-level WIT types which are lifted from and lowered to core Wasm types as part of the component ABI.

libc-bottom-half/headers/public/wasi/preview2.h
dicej: Agreed; I'll look into that.

libc-bottom-half/headers/public/wasi/preview2.h
dicej: https://github.com/bytecodealliance/wit-bindgen/pull/804

libc-bottom-half/headers/public/wasi/preview2.h
dicej: https://github.com/bytecodealliance/wit-bindgen/pull/805

libc-bottom-half/cloudlibc/src/libc/sys/wasi/preview2.c
```diff
@@ -0,0 +1,4448 @@
+// Generated by `wit-bindgen` 0.16.0. DO NOT EDIT!
```
sunfishcode, line 1: I'd like to avoid adding new code within the `cloudlibc` directory, so that `cloudlibc` just contains code derived from cloudlibc. Could you move the bindings to a directory like libc-bottom-half/sources/preview2, or even just libc-bottom-half/preview2?

sunfishcode (commented):
Overall this looks good to me. Just one relatively minor detail:

libc-bottom-half/cloudlibc/src/libc/sys/wasi/preview2.c
dicej: Sure, no problem.  We also added various new files under the `cloudlibc` directory in https://github.com/WebAssembly/wasi-libc/pull/449/files, while also adding code to other existing files.  Should we move all the new files elsewhere, and is there anything special we should do when modifying the existing files?

sunfishcode: Yeah, ideally we should move those new files out of `cloudlibc`. I don't think we need to do anything else special though.

We already only use a few small parts of the full `cloudlibc`, and the parts we do use are pretty specific to preview1 which was derived from cloudabi, so as preview2 progresses, we may even stop using `cloudlibc` altogether.

dicej: I've moved the .c and .o files to libc-bottom-half/sources

sunfishcode (approved):
Looks great!

PR #462, opened 2024-01-10
emmalloc: use __heap_end instead of sbrk(0)

cuviper:
This synchronizes emmalloc with the dlmalloc changes in #377 and #394.


cuviper:
This is an alternative to #378, and more current since `__heap_end` was made mandatory in #394. I had suggested changes in that other PR a few months ago, but never heard back.

sunfishcode (approved):
Thanks!

PR #463, opened 2024-01-10
add stubs for `statvfs`, `chmod`, etc.

dicej:
Per https://github.com/WebAssembly/wasi-sdk/issues/373, LLVM's libc++ no longer allows us to enable `<fstream>` and `<filesystem>` separately -- it's both or neither.  Consequently, we either need to patch libc++ to not use `statvfs`, `chmod`, etc. or add stub functions for those features to `wasi-libc`.  Since we're planning to eventually support those features with WASI Preview 2 and beyond, it makes sense to do the latter.

Note that since libc++ uses `DT_SOCK`, I've added a definition for it -- even though WASI Preview 1 does not define it.  No Preview 1 file will ever have that type, so code that handles that type will never be reached, but defining it allows us to avoid WASI-specific patches to libc++.

Related to `DT_SOCK`, I had to change the `S_IFIFO` value so it does not conflict with `S_IFSOCK`, thereby avoiding ambiguity in `__wasilibc_iftodt`.

### DIFF
--- a/libc-top-half/musl/include/stdlib.h
+++ b/libc-top-half/musl/include/stdlib.h
@@ -128,9 +128,7 @@ int rand_r (unsigned *);
 
 #if defined(_XOPEN_SOURCE) || defined(_GNU_SOURCE) \
  || defined(_BSD_SOURCE)
-#ifdef __wasilibc_unmodified_upstream /* WASI has no absolute paths */
 char *realpath (const char *__restrict, char *__restrict);
-#endif
 long int random (void);
 void srandom (unsigned int);
 char *initstate (unsigned int, char *, size_t);
--- a/Makefile
+++ b/Makefile
@@ -425,7 +425,6 @@ MUSL_OMIT_HEADERS += \
     "bits/shm.h" "bits/msg.h" "bits/ipc.h" "bits/ptrace.h" \
     "bits/statfs.h" \
     "sys/vfs.h" \
-    "sys/statvfs.h" \
     "syslog.h" "sys/syslog.h" \
     "wait.h" "sys/wait.h" \
     "ucontext.h" "sys/ucontext.h" \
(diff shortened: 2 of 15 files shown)


sunfishcode (approved):
Looks good to me.

libc-top-half/musl/include/stdlib.h
```diff
@@ -128,9 +128,7 @@ int rand_r (unsigned *);
 
 #if defined(_XOPEN_SOURCE) || defined(_GNU_SOURCE) \
  || defined(_BSD_SOURCE)
-#ifdef __wasilibc_unmodified_upstream /* WASI has no absolute paths */
 char *realpath (const char *__restrict, char *__restrict);
-#endif
```
yamt, line 133: is this intended?

dicej: Yes -- `realpath` is one of the functions libcxx's `<filesystem>` relies on, although I just realized I neglected to add `realpath.c` to `LIBC_TOP_HALF_MUSL_SOURCES` in the Makefile, and I'm not sure if that implementation is what we want (vs. a stub), so I'll need to take a closer look.

yamt: it was intentionally disabled in the header. https://github.com/WebAssembly/wasi-libc/pull/82
if you are going to add a realpath implementation anytime soon, it's great. please go ahead.
otherwise, it's probably better to revert this part for now, until the implementation gets ready.

dicej: Reverting this part causes the libcxx build to fail, so I don't think that's an option (unless we want to patch libcxx).  I've added a stub in https://github.com/WebAssembly/wasi-libc/pull/473, which I believe is the best we can do right now short of patching libcxx.

dicej: Update: the stub wasn't needed -- turns out musl's `realpath` implementation works fine.

PR #464, opened 2024-01-11
add descriptor table for mapping fds to handles

dicej:
This introduces `descriptor_table.h` and `descriptor_table.c`, providing a global hashtable for tracking `wasi-libc`-managed file descriptors.

WASI Preview 2 has no notion of file descriptors and instead uses unforgeable resource handles (which are currently represented as integers at the ABI level, used as indices into per-component tables managed by the host).  Moreover, there's not necessarily a one-to-one correspondence between POSIX file descriptors and resource handles (e.g. a TCP connection may require separate handles for reading, writing, and polling the same connection). We use this table to map each POSIX descriptor to a set of one or more handles and any extra state which libc needs to track.

Note that we've added `descriptor_table.h` to the
libc-bottom-half/headers/public/wasi directory, making it part of the public API.  The intention is to give applications access to the mapping, enabling them to convert descriptors to handles and vice-versa should they need to interoperate with both libc and WASI directly.


libc-bottom-half/headers/public/wasi/descriptor_table.h
```diff
@@ -0,0 +1,121 @@
+#ifndef DESCRIPTOR_TABLE_H
+#define DESCRIPTOR_TABLE_H
+
+#include <wasi/preview2.h>
+
+typedef struct {
+} tcp_socket_state_unbound_t;
```
yamt, line 7: empty struct is not valid in C. if i were you, i'd add `int dummy` field.

libc-bottom-half/headers/public/wasi/descriptor_table.h
```diff
@@ -0,0 +1,121 @@
+#ifndef DESCRIPTOR_TABLE_H
+#define DESCRIPTOR_TABLE_H
```
sbc100, line 2: Are we sure we want to make this into a public API?   So far WASI-libc has not really added new APIs.. and more of the code has been an internal implementation detail.  

If we add new APIs and make them public we then have to start supporting those APIs going forward.  If we can get away with making this internal (at least initially) that would be been IMHO.

libc-bottom-half/headers/public/wasi/descriptor_table.h
```diff
@@ -0,0 +1,121 @@
...
+		tcp_socket_state_bound_t bound;
+		tcp_socket_state_connecting_t connecting;
+		tcp_socket_state_connected_t connected;
+		tcp_socket_state_connect_failed_t connect_failed;
+		tcp_socket_state_listening_t listening;
+	};
+} tcp_socket_state_t;
+
+typedef struct {
+	tcp_own_tcp_socket_t socket;
```
sbc100, line 47: Are these lower level types also still integers at the wasm module boundary?  i.e. are we still mapping integer "file descriptors" to interpreter "resource handles" here?    Or are we actually using unforgable externrefs at the core wasm level now?

dicej: They're still integers, but they're unforgable in the sense that they're indices into host-managed, per-component tables, which means that if a component makes a call to another component, passing a resource handle, the host will interpose itself and look up the resource in the caller's table for that index, find (or create) the corresponding index in the callee's table, and pass that index to the callee.  Which I assume is also how externrefs work at some level, but I'm still learning about those :)

@alexcrichton can perhaps chime in here if my understanding is incorrect.

sbc100: Thanks that makes sense.  I know we are going to be moving to externref at some point but I've not been following too closely recently.

It might be worth mention somewhere (perhaps in the PR description) that the current "unforgeable resource handles" as still just integers and the core wasm/C/wasi-libc level, just like file descriptors.

libc-bottom-half/headers/public/wasi/descriptor_table.h
dicej: Yes, I agree that it's a big commitment and not to be taken lightly, and I'm open to postponing making this API public.

I do expect applications will want it, though.  For example, I've updated Rust's `mio` library to support WASI Preview 2's `wasi:io/poll` interface directly rather than go through e.g. POSIX's `poll` or `select`, which reduces the layers of abstraction and their attendant performance and debugging overhead.  Without some way to translate handles into descriptors and vice-versa, that wouldn't be practical, since `mio` and the libraries and apps built on top of it need to interoperate with Rust's stdlib, which is in turn built on libc.

So it's not a "must-have", but it's already quite useful, and I expect it will become even more useful over time.  For example, we expect that WASI 0.2.x will have a `forward` call which tells the host to forward a `wasi:io/streams/output-stream` to a `wasi:io/streams/input-stream`, offloading that responsibility from the guest to the host and allowing it to proceed even after the guest is uninstantiated (optionally reinstantiating it if necessary to report failure).  Being able to e.g. forward the contents of a large file to an HTTP response body using that mechanism would be quite useful, and part of making it ergonomic is to be able to convert a libc file descriptor to a WASI `input-stream`.

sunfishcode: We should probably document that the WASI 0.2 ABI isn't stable for now. And it is worth thinking about whether we should make these table entries opaque to avoid exposing them in the ABI. That said, I do think we'll want to have public API for working with the table, to support interop between code that uses handles directly and code that uses libc fds.

dicej: In the interest of keeping this whole effort moving forward, I'll go ahead and move this file to `headers/private` for now, and we can revisit later.

PR #470, opened 2024-02-08
Add support for pthread_getattr_np

Milek7:
Add pthread_getattr_np for getting stack bounds.

### DIFF
--- a/libc-top-half/musl/src/env/__init_tls.c
+++ b/libc-top-half/musl/src/env/__init_tls.c
@@ -31,25 +31,35 @@ extern unsigned char __global_base;
 extern weak unsigned char __stack_high;
 extern weak unsigned char __stack_low;
 
-static inline void setup_default_stack_size()
-{
-	ptrdiff_t stack_size;
+struct stack_bounds {
+	void *base;
+	size_t size;
+};
 
-	if (&__stack_high)
-		stack_size = &__stack_high - &__stack_low;
-	else {
+static inline struct stack_bounds get_stack_bounds()
+{
+	struct stack_bounds bounds;
+
+	if (&__stack_high) {
+		bounds.base = &__stack_high;
+		bounds.size = &__stack_high - &__stack_low;
+	} else {
 		unsigned char *sp;
 		__asm__(
 			".globaltype __stack_pointer, i32\n"
 			"global.get __stack_pointer\n"
 			"local.set %0\n"
 			: "=r"(sp));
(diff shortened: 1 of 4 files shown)


libc-top-half/musl/src/env/__init_tls.c
```diff
@@ -31,25 +31,35 @@ extern unsigned char __global_base;
...
 			"local.set %0\n"
 			: "=r"(sp));
-		stack_size = sp > &__global_base ? &__heap_base - &__data_end : (ptrdiff_t)&__global_base;
+		if (sp > &__global_base) {
+			bounds.base = &__heap_base;
+			bounds.size = &__heap_base - &__data_end;
+		} else {
+			bounds.base = &__global_base;
+			bounds.size = (size_t)&__global_base;
+		}
```
sbc100, line 59: This code only calculates the stack found of the main thread (i.e. the singular stack that the linker produces).  Is that intended?

libc-top-half/musl/src/thread/pthread_getattr_np.c
```diff
@@ -20,5 +23,9 @@ int pthread_getattr_np(pthread_t t, pthread_attr_t *a)
 			l += PAGE_SIZE;
 		a->_a_stacksize = l;
 	}
+#else
```
sbc100, line 26: Perhaps we can leave the `if / else` above and just replace the else block. 

It looks like that `if` block covers secondary threads and the `else` block covers the main thread?  We could use the `__stack_high` / `__stack_low` globals for the else case I think?

libc-top-half/musl/src/env/__init_tls.c
Milek7: This fills data for main thread, spawned threads have these fields filled in pthread_create.

libc-top-half/musl/src/thread/pthread_getattr_np.c
Milek7: The idea was to fill these fields in main thread struct and handle everything the same way. Otherwise this would need to duplicate the logic for getting stack bounds, fallback to __heap_base, etc...

Milek7: Replaced only `else` block.

sbc100 (approved):

abrown (approved):
This looks fine to me as well. @yamt, @loganek: does any one want to test this out?

yamt (approved):
looks reasonable to me

abrown:
@Milek7, can you rebase to fix the conflicts and then I can merge?

PR #471, opened 2024-02-12
Update bindings and dependencies to WASI 0.2.0

dicej:
Now that WASI 0.2.0 has been released, we can update from the release candidate we were previously using.  That includes updating various URLs in the Makefiles to point to newer deps; this is hopefully the last time we have to do that for a while.  Updates include:

- Upgrade to the v0.2.0 release of `wasi-cli`
- Upgrade to wit-bindgen 0.17.0 instead of the temporary pre-release build we had been using
- Refreshed bindings using the above
- Upgrade to Wasmtime 17 for testing

In addition, this PR makes a couple of tweaks to the Makefiles to make them more robust:
- Escape "." character in `sed` regex
- Split the `component new` step into its own rule so that we never confuse core modules with components


test/Makefile
```diff
@@ -151,9 +151,12 @@ build: download $(WASMS)
 
 $(WASMS): | $(OBJDIRS)
 $(OBJDIR)/%.wasm: $(OBJDIR)/%.wasm.o $(INFRA_WASM_OBJS)
-	$(CC) $(CFLAGS) $(LDFLAGS) -o $@ $^
+	$(CC) $(CFLAGS) $(LDFLAGS) -o tmp.wasm $^
```
sbc100, line 154: Its kind of a shame that we don't have a different file extension for components and core modules.. then we could just use two different rules here.  e.g. one rule for linking core modules from object files, and second rule that transforms wasm modules into components (based only file extension).

Is there a plan for a separate extension?    Should we just invent one here?   Perhaps we could use `foo.core.wasm` for the core module?

dicej: I'm not aware of a plan for a separate extension.  Yes, we could invent one.  I guess splitting this rule into two implies that we'd also need two variants of the `%.wasm.err: %.wasm` rule below, one for when we want to run the component and the other for when we want to run the core module?

sbc100: That sounds cleaner to me yes.. especially since the command we use to invoke core modules vs components could looks very different.

sbc100: Was chatting with @sunfishcode and he suggested that using `.core.wasm` here some reasonable.   (Or we could invert it and give components a special name such as `.cmp.wasm` or `.wasm.cmp`).

dicej: OK, I just pushed an update.  In order to test it properly, I had to update the Wasmtime URL from the (constantly updating) `dev` release to the `v17.0.0` release, which in turn required updating the bindings to the final WASI `0.2.0` release, so the PR has ballooned.  Apologies for that.

sbc100: Do you want to update the PR title and description to include all this new stuff?  Or make a separate PR?

dicej: I've updated the title and description.

dicej:
Any more feedback on this one?  I'd like to get it merged soon-ish, since it will unblock further work on `wasi-sockets`.

sbc100 (commented):
I don't have any more feedback.  Perhaps @sunfishcode could give final lgtm though?

PR #472, opened 2024-02-12
add preview2_component_type.o to libc.a and libc.so

dicej:
This file adds a custom section to each core module linked with wasi-libc.  That custom section contains component type information needed by e.g. `wasm-tools component new` to generate a component from the module.  It will be required once we start using any part of WASI 0.2.0 directly (vs. via a Preview 1 adapter).  In addition, it allows developers to `#include <wasi/preview2.h>` in their code and make use of those APIs directly even if wasi-libc is not using them yet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -383,6 +383,9 @@ DLMALLOC_OBJS = $(call objs,$(DLMALLOC_SOURCES))
 EMMALLOC_OBJS = $(call objs,$(EMMALLOC_SOURCES))
 LIBC_BOTTOM_HALF_ALL_OBJS = $(call objs,$(LIBC_BOTTOM_HALF_ALL_SOURCES))
 LIBC_TOP_HALF_ALL_OBJS = $(call asmobjs,$(call objs,$(LIBC_TOP_HALF_ALL_SOURCES)))
+ifeq ($(WASI_SNAPSHOT), preview2)
+LIBC_OBJS += $(OBJDIR)/preview2_component_type.o
+endif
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -604,6 +607,10 @@ $(OBJDIR)/%.long-double.pic.o: %.c include_dirs
 	@mkdir -p "$(@D)"
 	$(CC) $(CFLAGS) -MD -MP -o $@ -c $<
 
+$(OBJDIR)/preview2_component_type.pic.o $(OBJDIR)/preview2_component_type.o: $(LIBC_BOTTOM_HALF_SOURCES)/preview2_component_type.o
+	@mkdir -p "$(@D)"
+	cp $< $@
```
sunfishcode, line 612: Is copying needed, or would it be enough to just add `$(LIBC_BOTTOM_HALF_SOURCES)/preview2_component_type.o` to `LIBC_OBJS`?

dicej: Sorry, I saw this and then forgot to follow up.  I'll see if I can eliminate the copy.  I think I was concerned adding a already-generated object file to `LIBC_OBJS` would break a pattern rule somewhere, but we'll see...

dicej: I took another look at this, and I think eliminating the copy would be more trouble than it's worth since the pattern rules for building the .o files for libc.a and libc.so each need their own version (i.e. the .o and the .pic.o, respectively).  The file is only 11KB so I don't think the copy adds noticeable bloat to the build.

sunfishcode:
Sounds good. Thanks!

PR #476, opened 2024-02-23
Document a plan for transitioning to preview 2 and beyond

abrown:
This document outlines how wasi-libc can manage the changes to WASI 0.2.0 as well as future 0.2.x releases. It's meant as a starting point for collaboration; I'm interested in your feedback.

[Rendered](https://github.com/abrown/wasi-libc/blob/preview2-plan/docs/preview2-transition.md)

### DIFF
--- a/docs/preview2-transition.md
+++ b/docs/preview2-transition.md
@@ -1,0 +1,55 @@
+# WASI Preview Transition
+
+How will wasi-libc transition to WASI preview 2? This document outlines a plan
+and describes how to maintain preview 1 support at the same time. This document
+will _not_ discuss how and if and when preview 1 support could be dropped from
+the repository; the WASI subgroup is a better forum for that.
+
+There seem to be several parts to this:
+
+1. __new targets__: to avoid confusion, let's adopt the target names agreed to
+   elsewhere: `wasm32-wasip1`, `wasm32-wasip2`. This means that new releases of
+   wasi-libc/wasi-sdk would contain new `wasm32-wasip*` subdirectories and users
+   could compile with `--target wasm32-wasip*`. The existing `wasm32-wasi` and
+   `wasm32-wasi-threads` targets can continue to coexist until some undefined
+   future date (note that the contents of `wasm32-wasip1` and `wasm32-wasi`
+   should be identical). Future releases of WASI, e.g., 0.2.1, 0.2.2, 0.2.x,
+   will result in new `wasm32-wasip2.*` subdirectories.
+
+2. __new internal macro definitions__: internally in wasi-libc, let's use
+   `#ifdef __wasilibc_wasip*` blocks to conditionally compile code that depends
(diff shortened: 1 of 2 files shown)


docs/preview2-transition.md
```diff
@@ -0,0 +1,55 @@
...
+There seem to be several parts to this:
+
+1. __new targets__: to avoid confusion, let's adopt the target names agreed to
+   elsewhere: `wasm32-wasip1`, `wasm32-wasip2`. This means that new releases of
+   wasi-libc/wasi-sdk would contain new `wasm32-wasip*` subdirectories and users
+   could compile with `--target wasm32-wasip*`. The existing `wasm32-wasi` and
+   `wasm32-wasi-threads` targets can continue to coexist until some undefined
+   future date (note that the contents of `wasm32-wasip1` and `wasm32-wasi`
+   should be identical). Future releases of WASI, e.g., 0.2.1, 0.2.2, 0.2.x,
+   will result in new `wasm32-wasip2.*` subdirectories.
```
alexcrichton, line 17: I might recommend avoiding having a directory-per-release and only having a `wasm32-wasip2` directory. Per https://github.com/bytecodealliance/wasmtime/issues/7860 and how other runtimes are expected to behave similarly it would be ok to take a wasi-libc compiled against WASI 0.2.2 and run it on a runtime that only supports WASI 0.2.0, so long as no new functionality from WASI 0.2.2 is actually used. C/Rust/etc are good at this since unused code can easily be statically removed, so I think that provides a smooth path to upgrade wasi-libc bindings here without disrupting the ecosystem.

docs/preview2-transition.md
```diff
@@ -0,0 +1,55 @@
...
+   versions. The `__wasilibc_wasip*` macro definitions exist to clearly identify
+   which parts belong to each preview (e.g., this should make maintenance easier
+   for those interested in supporting preview 1). Since we expect ABI
+   compatibility between previews, all prior versions should be defined; e.g.,
+   for target `wasip2.1`, both `__wasilibc_wasip2` and `__waslibc_wasip2.1`
+   would be defined but `__wasilibc_wasip1` would not.  PR [#449] proposed
+   `__wasilibc_use_preview2` but (a) `...wasip*` lines up the names with the
+   targets, making "what gets compiled where" more clear and (b) we will likely
+   need the ability to specify which precise version a block applies to in a
+   fine-grained way (e.g., `#ifdef __wasilibc_wasip0.2.2`).
```
alexcrichton, line 33: Sort of like above, I might recommend only `__wasilibc_wasip2`, but no further. (e.g. no per-point-release macros). Any one copy of the source of wasi-libc will have some version of WASI that it has bindings generated for, and the source would assume that (e.g. in theory there shouldn't need to be code for "if bigger than 0.2.1 add this code" since it would instead always be included once the WIT files are here)

PR #479, opened 2024-03-07
fix `#ifdef __cplusplus` guard in dirent.h

12101111:
fix: https://github.com/WebAssembly/wasi-sdk/issues/390

### DIFF
--- a/libc-bottom-half/headers/public/__header_dirent.h
+++ b/libc-bottom-half/headers/public/__header_dirent.h
@@ -21,15 +21,15 @@
 #define IFTODT(x) (__wasilibc_iftodt(x))
 #define DTTOIF(x) (__wasilibc_dttoif(x))
 
-int __wasilibc_iftodt(int x);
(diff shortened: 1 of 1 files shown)


libc-bottom-half/headers/public/__header_dirent.h
```diff
@@ -21,15 +21,15 @@
...
+#endif
 
 #include <__struct_dirent.h>
 #include <__typedef_DIR.h>
 
-#ifdef __cplusplus
-extern "C" {
-#endif
+int __wasilibc_iftodt(int x);
+int __wasilibc_dttoif(int x);
```
sunfishcode, line 32: It's a little untidy to include other headers inside the `extern "C"` block; could you move those `#include`s outside of the block?

sunfishcode (commented):
Thanks for fixing this! Just one small comment:

sunfishcode:
Thanks!

PR #482, opened 2024-03-13
add wasip2 implementations of more socket APIs

dicej:
This adds `wasm32-wasip2` implementations of `shutdown`, `getsockopt`, and `setsockopt`.  It also extends the existing `ioctl` implementation to handle both p1 and p2 file descriptors since we can't know until runtime which kind we have. Once we've moved `wasm32-wasip2` fully to WASI 0.2 and remove the need for the p1 adapter, we'll be able to switch to separate p1 and p2 `ioctl` implementations.


sunfishcode (approved):
Looks good! It's a little unfortunate to add new code to the cloudlibc directory, but I know why it's there, and we can factor that out later.

PR #483, opened 2024-03-15
Add libsetjmp.a/so

yamt:
Add setjmp/longjump support based on Wasm EH proposal.

It's provided as a separate library (libsetjmp) from libc so that runtimes w/o EH support can still load libc.so.

To use this setjmp/longjmp implementation, an application should be compiled with `-mllvm -wasm-enable-sjlj` and linked with `-lsetjmp`. (You need an LLVM with the change mentioned below.)

Also, you need a runtime with EH support to run such an application.

If you want to use the latest EH instructions, you can use `binaryen --translate-eh-old-to-new` on your application.

Note: You don't need to translate libsetjmp.a/so to the new EH. While LLVM currently produces bytecode for an old version of the EH proposal, luckily for us, the bytecode used in this library (ie. the tag definition and the "throw" instruction) is compatible with the latest version of the proposal.

The runtime logic is basically copy-and-paste from:
    https://github.com/yamt/garbage/tree/wasm-sjlj-alt2/wasm/longjmp

The corresponding LLVM change:
    https://github.com/llvm/llvm-project/pull/84137
    (Note: you need this change to build setjmp/longjmp using code.
    otoh, you don't need this to build libsetjmp.)

A similar change for emscripten:
    https://github.com/emscripten-core/emscripten/pull/21502

An older version of this PR, which doesn't require LLVM changes:
    https://github.com/WebAssembly/wasi-libc/pull/467

Discussion:
    https://docs.google.com/document/d/1ZvTPT36K5jjiedF8MCXbEmYjULJjI723aOAks1IdLLg/edit

An example to use the latest EH instructions:
```
clang -mllvm -wasm-enable-sjlj -o your_app.wasm your_app.c -lsetjmp
wasm-opt --translate-eh-old-to-new -o your_app.wasm your_app.wasm
toywasm --wasi your_app.wasm
```
Note: use toywasm built with `-DTOYWASM_ENABLE_WASM_EXCEPTION_HANDLING=ON`.

An example to use the older EH instructions, which LLVM currently produces:
```
clang -mllvm -wasm-enable-sjlj -o your_app.wasm your_app.c -lsetjmp
iwasm your_app.wasm
```
Note: use wasm-micro-runtime built with `-DWAMR_BUILD_EXCE_HANDLING=1`.
Note: as of writing this, only the classic interpreter supports EH.

### DIFF
--- a/libc-top-half/musl/src/setjmp/wasm32/rt.c
+++ b/libc-top-half/musl/src/setjmp/wasm32/rt.c
@@ -1,0 +1,83 @@
+/*
+ * a runtime implementation for
+ * https://github.com/llvm/llvm-project/pull/84137
+ * https://docs.google.com/document/d/1ZvTPT36K5jjiedF8MCXbEmYjULJjI723aOAks1IdLLg/edit
+ */
+
+#include <stddef.h>
+#include <stdint.h>
+
+/*
+ * function prototypes
+ */
+void __wasm_setjmp(void *env, uint32_t label, void *func_invocation_id);
+uint32_t __wasm_setjmp_test(void *env, void *func_invocation_id);
+void __wasm_longjmp(void *env, int val);
+
+/*
+ * jmp_buf should have large enough size and alignment to contain
+ * this structure.
+ */
+struct jmp_buf_impl {
+        void *func_invocation_id;
+        uint32_t label;
+
+        /*
+         * this is a temorary storage used by the communication between
+         * __wasm_sjlj_longjmp and WebAssemblyLowerEmscriptenEHSjL-generated
+         * logic.
+         * ideally, this can be replaced with multivalue.
+         */
+        struct arg {
+                void *env;
+                int val;
+        } arg;
+};
+
+void
+__wasm_setjmp(void *env, uint32_t label, void *func_invocation_id)
+{
+        struct jmp_buf_impl *jb = env;
+        if (label == 0) { /* ABI contract */
+                __builtin_trap();
+        }
+        if (func_invocation_id == NULL) { /* sanity check */
+                __builtin_trap();
+        }
+        jb->func_invocation_id = func_invocation_id;
+        jb->label = label;
+}
+
+uint32_t
+__wasm_setjmp_test(void *env, void *func_invocation_id)
+{
+        struct jmp_buf_impl *jb = env;
+        if (jb->label == 0) { /* ABI contract */
+                __builtin_trap();
+        }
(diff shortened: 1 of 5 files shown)


leddoo:
this looks really cool!
i was just trying to add sjlj support myself, but couldn't find any docs for `-wasm-enable-sjlj`.

i don't know how to propose changes to a PR, so i'll just write a comment.
the `setjmp.h` header is broken.
- the `#include <bits/setjmp.h>` include should not have been removed, `__jmp_buf` is undefined.
- i think you forgot to remove the `#else #warning setjmp is not yet implemented for WASI` bit at the bottom. that errors because the `#ifndef __wasilibc_unmodified_upstream` is already closed above.

yamt:
> this looks really cool! i was just trying to add sjlj support myself, but couldn't find any docs for `-wasm-enable-sjlj`.
> 
> i don't know how to propose changes to a PR, so i'll just write a comment. the `setjmp.h` header is broken.
> 
>     * the `#include <bits/setjmp.h>` include should not have been removed, `__jmp_buf` is undefined.
> 
>     * i think you forgot to remove the `#else #warning setjmp is not yet implemented for WASI` bit at the bottom. that errors because the `#ifndef __wasilibc_unmodified_upstream` is already closed above.

oops. fixed. thank you.

libc-top-half/musl/src/setjmp/wasm32/rt.c
```diff
@@ -0,0 +1,83 @@
+/*
+ * a runtime implementation for
+ * https://github.com/llvm/llvm-project/pull/84137
+ * https://docs.google.com/document/d/1ZvTPT36K5jjiedF8MCXbEmYjULJjI723aOAks1IdLLg/edit
```
sunfishcode, line 4: Ideally we should copy the content of this into the [tool-conventions](https://github.com/WebAssembly/tool-conventions/) repo. Would you be interested in doing this, or would you mind if I did it?

sunfishcode (approved):
This looks good to me; just one note below:

libc-top-half/musl/src/setjmp/wasm32/rt.c
yamt: i agree we should document in the tool-conventions repo.
i think i can write something next month.

yamt: i wrote something. https://github.com/WebAssembly/tool-conventions/pull/225

sunfishcode:
Thanks for filing that. This looks good!

PR #487, opened 2024-03-24
update old WASI and improving emmalloc.c

PiCoPress:
There are some unefficient code

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -82,6 +82,12 @@ jobs:
         echo "NM=$CLANG_DIR/llvm-nm" >> $GITHUB_ENV
       if: matrix.os == 'ubuntu-latest'
 
+    - name: Disable libsetjmp for old LLVM
+      shell: bash
+      run: |
+        echo "BUILD_LIBSETJMP=no" >> $GITHUB_ENV
+      if: matrix.clang_version == '10.0.0'
+
     - name: Build libc
       shell: bash
       run: |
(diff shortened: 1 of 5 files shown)


emmalloc/emmalloc.c
```diff
@@ -297,9 +297,12 @@ static int compute_free_list_bucket(size_t allocSize)
 {
   if (allocSize < 128) return (allocSize >> 3) - 1;
   int clz = __builtin_clz(allocSize);
-  int bucketIndex = (clz > 19) ? 110 - (clz<<2) + ((allocSize >> (29-clz)) ^ 4) : MIN(71 - (clz<<1) + ((allocSize >> (30-clz)) ^ 2), NUM_FREE_BUCKETS-1);
```
sbc100, line 300: These change to emmalloc.c seem unrelated to the WASI headers change.  Can you make a seperate PR.

Also, this change seems to include a bunch of formatting changes that make the real change hard to review.  Could you revert the formatting changes (Or make them separate PR if you like)?

Finally. emmalloc is copied from the emscripten project, so it probably better to make any fixes there:  (https://github.com/emscripten-core/emscripten/blob/main/system/lib/emmalloc.c) before copying them downstream here.   That actually good reason not to make any formatting changes downstream since it makes comparison with the upstream version harder.

sunfishcode:
I agree with @sbc100; could you submit the emmalloc.c changes to Emscripten first?

PR #488, opened 2024-03-28
implement `getsockname`, `getpeername`, and `getaddrinfo`

dicej:
This also includes stubs for `gethostbyname`, `gethostbyaddr`, etc. which were necessary to get CPython to build.  I believe it will be possible to implement them all properly at some point, but don't have the bandwidth at the moment.

Finally, this includes a few fixes for issues I missed in earlier PRs that surfaced when running the CPython `asyncio` test suite.


sunfishcode (approved):

abrown:
Looking at this again, did we mean to expose the `tcp|udp_get*` functions as a defined symbol (e.g., `tcp_getpeername`)? cc: @dicej

dicej:
@abrown It depends on what you mean by "expose".  Everything's exposed due to `-fvisibility=default` [here](https://github.com/dicej/wasi-libc/blob/b5095caa742e7168db766507d04277676dcb3d36/Makefile#L623C30-L623C42) and will presumably remain that way until we start annotating deliberately-exported functions with explicit visibility (at which point we can stop using `-fvisibility=default`).  Those functions aren't in any public header file, though, so they're not exposed in that sense AFAIK.

PR #489, opened 2024-04-03
add `__wasilibc_reset_preopens`

dicej:
This resets the preopens table to an uninitialized state, forcing it to be reinitialized next time it is needed.  This is useful when pre-initializing using e.g. `wizer` or `component-init`.  Such tools are capable of taking a snapshot of a running application which may later be resumed on an unrelated runtime (which may have its own, unrelated preopens).


sunfishcode:
Makes sense, thanks!

PR #494, opened 2024-05-04
Update `_POSIX_THREAD_XX` macro definitions

kateinoigakukun:
This is a follow-up to https://github.com/WebAssembly/wasi-libc/pull/356

* `_POSIX_THREAD_PROCESS_SHARED`: Functions part of Thread Process-Shared Synchronization option like `pthread_barrierattr_getpshared` are defined when threads is enabled, so define it.
*  `_POSIX_THREAD_SAFE_FUNCTIONS`: "thread-safe functions" are defined when threads is enabled.
* `_POSIX_THREAD_ATTR_STACKADDR`: wasi-libc doesn't provide `pthread_attr_{get,set}stackaddr` for now, so we should not define it.
* `_POSIX_THREAD_ATTR_STACKSIZE`: wasi-libc has `pthread_attr_getstack` definition
* `_POSIX_THREAD_PRIORITY_SCHEDULING`: WASI has no CPU scheduling support
* `_POSIX_THREAD_CPUTIME`: Although WASI Preview 1 has [`thread_cputime_id`](https://github.com/WebAssembly/WASI/blob/70214b878af4ce45889b4ad9d26a7ac98db8931b/legacy/preview1/witx/typenames.witx#L30), wasi-libc doesn't define `CLOCK_THREAD_CPUTIME_ID`.


sbc100 (approved):

abrown (approved):
Makes sense to me as well; thanks for the PR!

PR #495, opened 2024-05-07
Default wasm32-wasip2 objects to `-fPIC` by default

alexcrichton:
This commit updates all of the `*.a` archives built for the `wasm32-wasip2` target to use `-fPIC` instead of the default non-PIC flag. This makes them suitable for linking into either a dynamic library or a non-dynamic binary without recompiling libc. This is intended to help integrate shared libraries into other precompiled languages, such as Rust's standard library, as it makes its way upstream.

Lots of discussion about this also happened on #429 so I wanted to highlight that this is only happening for the wasm32-wasip2 target and no other targets. For example neither wasm32-wasi nor wasm32-wasip1 are affected. Only users of wasm32-wasip2, of which there aren't too too many, are possibly affected by the runtime overheads here. I have not measured the hypothetical runtime myself, however.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -601,6 +601,15 @@ $(OBJDIR)/libdl.so.a: $(LIBDL_SO_OBJS)
 
 $(OBJDIR)/libsetjmp.so.a: $(LIBSETJMP_SO_OBJS)
 
+# For the wasip2 target default to `-fPIC` for all archives to make them
+# suitable for linking later into a shared object if desired. This can come with
+# modest runtime overhead so this isn't the default for all targets, so other
+# targets below compile non-PIC versions of objects and use those for archives
+# instead.
+ifeq ($(WASI_SNAPSHOT),p2)
+$(SYSROOT_LIB)/%.a: $(OBJDIR)/%.so.a
+	cp $< $@
+else
 $(SYSROOT_LIB)/libc.a: $(LIBC_OBJS)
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
This is intended to eventually accompany https://github.com/rust-lang/rust/pull/124858 on the Rust side of things.

sbc100 (commented):
Happy to see this land.  

Hopefully there will not be any issues with linking PIC objects into static binaries, but I will be happy to look into them if they show up.

TerrorJack:
But for `wasm32-wasip2` target, the various `.so` files have already been built; I'm mildly curious why rust toolchain is unable to use those.

sbc100:
> But for `wasm32-wasip2` target, the various `.so` files have already been built; I'm mildly curious why rust toolchain is unable to use those.

The literal `.so` files themselves cannot be used to static linking, just like on native platforms.  They are only useful for the dynamic linking case.

alexcrichton:
Thinking more on this, that's a good point @TerrorJack. I blindly did this to align with what I was changing the Rust target to do but while it's appropriate in Rust there's no need here in C. The Rust target is changing the default for objects on wasm32-wasip2 to ensure that the precompiled standard library can be used with shared objects but it's not required to. If a shared object is being created though then libc.so will have to come from somewhere and it'll probably be an external `clang` which already has the bits and pieces and Rust's `libc.a` won't be used.

Given that I don't think that this is actually necessary, so I'll close this.

PR #496, opened 2024-05-10
dlmalloc: account the footprint of the initial heap

yamt:
While malloc_stats and friends are disabled and unused for wasi-libc, it's neater to be consistent.

Background: My colleagues for some reasons enabled malloc_stats and asked me why it reports negative values.

Note: Depending __heap_base, init_top() adjusts the address for alignment. I think the amount of this adjustment is reported as "used" by malloc_stats. I don't bother to "fix" it.

### DIFF
--- a/dlmalloc/src/malloc.c
+++ b/dlmalloc/src/malloc.c
@@ -5250,6 +5250,8 @@ static void try_init_allocator(void) {
   gm->least_addr = base;
   gm->seg.base = base;
   gm->seg.size = initial_heap_size;
+  gm->footprint = initial_heap_size;
+  gm->max_footprint = initial_heap_size;
   gm->magic = mparams.magic;
   gm->release_checks = MAX_RELEASE_CHECK_RATE;
   init_bins(gm);


sunfishcode:
This looks reasonable to me. It doesn't even look wrong that this doesn't account for `init_top` adjusting for alignment, because it's about the amount of memory malloc has "obtained" from the system, whether it's used for user data or not.

We may want to `#ifdef` out the `footprint` and `max_footprint` fields altogether by default, to save code size, because we don't expose that functionality in the default wasi-libc configuration. But we can do that separately.

sunfishcode (approved):

sbc100:
Do you know why upstream dlmalloc doesn't do this?  Is this a bug in dlmalloc?

Can you mark these lines with `__wasilibc_unmodified_upstream` like we do elsewhere?

sunfishcode:
This code is code that we added for WASI, so it's already under `__wasilibc_unmodified_upstream` guards. And we don't expose the footprint functionality publicly, which is why we didn't notice this before.

yamt:
> This looks reasonable to me. It doesn't even look wrong that this doesn't account for `init_top` adjusting for alignment, because it's about the amount of memory malloc has "obtained" from the system, whether it's used for user data or not.

i agree it isn't necessarily wrong. i mentioned it just because my colleagues will likely have a question.

> We may want to `#ifdef` out the `footprint` and `max_footprint` fields altogether by default, to save code size, because we don't expose that functionality in the default wasi-libc configuration. But we can do that separately.

maybe.

PR #500, opened 2024-05-21
Extend wasi-emulated-mman with `mprotect`

whitequark:
This implementation never changes actual memory protection, but it does verify that the address range and flags are valid. The direct motivation is fixing a linker error where LLVM links to `mprotect` in dead code.


libc-bottom-half/mman/mman.c
```diff
@@ -122,3 +123,35 @@ int munmap(void *addr, size_t length) {
...
+    // Range must be in bounds of linear memory.
+    size_t memory_size = __builtin_wasm_memory_size(0) * PAGESIZE;
+    if (end > memory_size) {
+        errno = ENOMEM;
+        return -1;
+    }
+
+    // Can only protect memory as read/write (which is a no-op).
+    if (prot != (PROT_READ | PROT_WRITE)) {
+        errno = EINVAL;
```
yamt, line 151: ENOTSUP?

whitequark: I don't see ENOTSUP [here](https://www.man7.org/linux/man-pages/man2/mprotect.2.html).

yamt: see https://pubs.opengroup.org/onlinepubs/9699919799/functions/mprotect.html#tag_16_334_05

whitequark: Ah, I shouldn't have been lazy. Fixed!

sbc100:
Would it make sense if patch LLVM to honor some kind of HAVE_MPROTECT instead/also?

whitequark:
> Would it make sense if patch LLVM to honor some kind of HAVE_MPROTECT instead/also?

That's what I've been doing so far, but:

a. There is pushback against adding more compiletests, as they are slow and some find them hard to maintain.
b. A sensible implementation for wasi-emulated-mman is possible, so why not add it here? It may benefit other applications.

sbc100:
> > Would it make sense if patch LLVM to honor some kind of HAVE_MPROTECT instead/also?
> 
> That's what I've been doing so far, but:
> 
> a) There is pushback against adding more compiletests, as they are slow and some find them hard to maintain. 

Do they need to be compiletests?  Can't it just be something that is hardcoded into the CMakeLists on the LLVM side?  something like:

```
if (IS_WASM) set(HAVE_MMAP, 0)
``

> b) A sensible implementation for wasi-emulated-mman is possible, so why not add it here? It may benefit other applications.

Fair enough, I agree it makes sense to emulate all these mman symbols.

However I do think it would be better for llvm to not include the dead code that calls mmap/mprotect at all.  It seems like a good in general.   Perhaps we can do both things?

sbc100 (approved):

whitequark:
> Can't it just be something that is hardcoded into the CMakeLists on the LLVM side? something like:

Maybe, but I still don't understand exactly what changes to `CMakeLists.txt` for WASI will be upstreamable, and it seems there are objections to having those `#define`s in `config.h.cmake` too. So for now I'm avoiding doing any speculative work. If the current set of compiletests is accepted then one more for `mprotect()` is probably fine too (but I must point out that it will require a different approach because wasi-libc advertises support for `mprotect()` in the headers; it just doesn't have the symbol).


> However I do think it would be better for llvm to not include the dead code that calls mmap/mprotect at all. It seems like a good in general. Perhaps we can do both things?

It looks like there's currently no way to disable the JIT (which is a little odd to me), which is what probably needs to be done for this. LTO builds don't suffer from this problem fwiw.

sunfishcode (approved):
Looks good to me.

whitequark:
Thanks!

PR #502, opened 2024-05-27
Makefile: separate the target to create empty dummy libraries

yamt:
During develompment, it's sometimes convenient to have a separate target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -785,7 +785,7 @@ endif
 
 libc: include_dirs $(STATIC_LIBS)
 
-finish: startup_files libc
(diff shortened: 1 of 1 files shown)


sunfishcode (approved):

sunfishcode:
The CI failure may be due to the directory not being created. The other rules start with a `mkdir -p` on their output directory to ensure that it exists.

yamt:
> The CI failure may be due to the directory not being created. The other rules start with a `mkdir -p` on their output directory to ensure that it exists.

fixed. thank you.

PR #505, opened 2024-06-04
Add LTO build option

yamt:
An old PR which I used as a base:
https://github.com/WebAssembly/wasi-libc/pull/150

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,16 @@ BUILD_LIBC_TOP_HALF ?= yes
 BUILD_LIBSETJMP ?= yes
 # The directory where we will store intermediate artifacts.
 OBJDIR ?= build/$(TARGET_TRIPLE)
+
+# LTO; no, full, or thin
+# Note: thin LTO here is just for experimentation. It has known issues:
+# - https://github.com/llvm/llvm-project/issues/91700
+# - https://github.com/llvm/llvm-project/issues/91711
+LTO ?= no
+ifneq ($(LTO),no)
+CLANG_VERSION ?= $(shell ${CC} -dumpversion)
+override OBJDIR := $(OBJDIR)/llvm-lto/$(CLANG_VERSION)
+endif
 # The directory where we store files and tools for generating WASIp2 bindings
 BINDING_WORK_DIR ?= build/bindings
 # URL from which to retrieve the WIT files used to generate the WASIp2 bindings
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -25,6 +25,16 @@ BUILD_LIBC_TOP_HALF ?= yes
 BUILD_LIBSETJMP ?= yes
 # The directory where we will store intermediate artifacts.
 OBJDIR ?= build/$(TARGET_TRIPLE)
+
+# LTO; no, full, or thin
+# Note: thin LTO here is just for experimentation. It has known issues:
+# - https://github.com/llvm/llvm-project/issues/91700
+# - https://github.com/llvm/llvm-project/issues/91711
```
sbc100, line 32: In addition to these issue I think you will find that any symbols that implement libcalls, or required by the libcall implemenation must currently be excluded from LTO.  See https://github.com/emscripten-core/emscripten/blob/3de1b9091adbadc46706959bd97efa033214bc2c/tools/system_libs.py#L1007-L1020

yamt: is it still necessary after https://reviews.llvm.org/D71738? do you have a test case?

yamt:
wasi-sdk counterpart https://github.com/WebAssembly/wasi-sdk/pull/424

Makefile
yamt: i have excluded __cxa_atexit since then.
other things listed in your emscripten link seem like compiler-rt, which i don't have a plan to build with LTO for now.t

sbc100: I think one issue is that compiler-rt can have transitive dependencies into libc.   I imagine you will run into similar issues here once you start testing more extensively with LTO, but perhaps I'm wrong, or perhaps things have changed in some way.  

Anyway, I guess its fine to land this and then fix issues as they come up in the field (which is basically what we did in emscripten).

yamt: i glanced at compiler-rt.
i couldn't find problematic transitive dependencies.
obvious dependencies like memset etc are covered by https://reviews.llvm.org/D71738.

yamt:
can this land?

yamt:
closed by mistake and then reopened.

Makefile
```diff
@@ -627,6 +658,8 @@ $(SYSROOT_LIB)/libsetjmp.a: $(LIBSETJMP_OBJS)
 
 $(PIC_OBJS): CFLAGS += -fPIC -fvisibility=default
 
+$(LIBC_NONLTO_OBJS): CFLAGS += -fno-lto
```
sbc100, line 661: nit: I wonder if we could instead filter out the `-flto` flag instead of passing both `-flto` and `-fno-lto`?

sbc100 (approved):
lgtm.   We can always add object to LIBC_NONLTO_OBJS later if we find any issues going forward.

Makefile
yamt: done

PR #508, opened 2024-06-21
0-initialize thread-specific data upon thread creation.

veluca93:
If thread-specific data is not set to 0 upon thread creation, `__pthread_tsd_run_dtors` will end up running destructors passing uninitialized memory as memory addresses, which can lead to memory corruption.

This issue can be triggered when malloc() returns a memory address that was freed before, as in that case memory is not zeroed out.


yamt (approved):
lgtm

abrown (approved):
Same here; thanks for the contribution!

PR #510, opened 2024-06-25
`times` should always return 0 for `tms_cutime`

kateinoigakukun:
`tms_cutime` is the sum of the user times of child processes *excluding the current process*. Since WASI doesn't provide a way to spawn a new process, this value should always be 0.

Interestingly PSPSDK had exactly the same issue and @fjtrujy found wasi-libc has the issue too 😅  https://github.com/pspdev/pspsdk/pull/193/files/eef342187319ead21cfe9a6598fd79a21188c1fa#diff-8318a7fe507925448c5a85d6bfdeb9f720a24b6d38d583eea8542758898cc4e7


yamt (approved):
lgtm

PR #517, opened 2024-07-18
Use the correct compiler-rt

yamt:
eg. when a user specifies non-default -resource-dir.

cf. https://github.com/WebAssembly/wasi-sdk/pull/446

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,7 +66,7 @@ ifeq ($(WASI_SNAPSHOT), p2)
 TARGET_TRIPLE = wasm32-wasip2
 endif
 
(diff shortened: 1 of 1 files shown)


yamt:
@alexcrichton

alexcrichton:
Why does the builtins lib need to be passed to the compiler specifically when creating `*.so`? Shouldn't the compiler automatically add it?

yamt:
> Why does the builtins lib need to be passed to the compiler specifically when creating `*.so`? Shouldn't the compiler automatically add it?

because it's suppressed by -nodefaultlibs.

yamt:
@alexcrichton do you have any other concerns on this PR?

alexcrichton (approved):

Issue #520, opened 2024-07-26
FTS Support


kateinoigakukun:
wasi-libc does not include fts(3) implementation now, and it prevents us from reusing other platforms' code while porting [swift-foundation](https://github.com/apple/swift-foundation) to WASI.

Can we ship https://github.com/void-linux/musl-fts licensed under BSD 3-Clause as a part of wasi-libc?

The implementation is well-matured and well-used in the Alpine ecosystem. The only change we need to apply to the implementation is not to depend on `fchdir` by making `FTS_NOCHDIR` default.

sbc100:
Since we already have some BSD-licensed parts of wasi-libc it doesn't seem unreasonable to me.  

Could we include it via a submodule and upstream the `FTS_NOCHDIR` mod you mention?

PR #522, opened 2024-08-01
Initial FTS support

kateinoigakukun:
Close https://github.com/WebAssembly/wasi-libc/issues/520

Add FTS implementation derived from musl-fts with a few modifications. The compiled fts.o is included in the libc.a archive, and the fts.h header is installed in the sysroot (`include/fts.h`).

* fts/musl-fts: Add a copy of the musl-fts sources with modifications.
* fts/patches: A set of patches to apply to the musl-fts sources.
  * Upstream pull request: https://github.com/void-linux/musl-fts/pull/14
* fts/update-musl-fts.sh: A script to update the musl-fts sources with the patches applied.
* fts/config.h: A configuration header included by the musl-fts sources.
* test/smoke: Add a test suite for wasi-libc specific features that libc-test does not cover.

### DIFF
--- a/fts/patches/0001-Add-conditional-handling-for-fchdir-and-chdir-functi.patch
+++ b/fts/patches/0001-Add-conditional-handling-for-fchdir-and-chdir-functi.patch
@@ -1,0 +1,102 @@
+From 878675b2299c23593a9adee25d5c828fb760819e Mon Sep 17 00:00:00 2001
+From: Yuta Saito <kateinoigakukun@gmail.com>
+Date: Thu, 1 Aug 2024 01:50:49 +0000
+Subject: [PATCH] Add conditional handling for fchdir and chdir functions
+
+Define preprocessor conditions for `HAVE_FCHDIR` to handle platforms without
+fchdir. On these platforms, pretend that `FTS_NOCHDIR` is always set.
+
+This change is necessary to build the `fts` module on the top of the WASI and
+wasi-libc, which doesn't provide the `fchdir`.
+
+Upstream pull request: https://github.com/void-linux/musl-fts/pull/14
+---
+ fts.c | 24 ++++++++++++++++++++++++
+ 1 file changed, 24 insertions(+)
+
+diff --git a/fts.c b/fts.c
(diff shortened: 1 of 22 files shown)


sbc100 (commented):
Thanks for working on this.  

I assume if/when the upstream patch lands we can replace most of this with a git submodule?  (i.e. remove the update script and the patch system?).

I wonder if its worth waiting a few days to see if upstream will simply accept the patch?

kateinoigakukun:
Given that the latest release was 8 years ago and the main maintainer has no public activity last 3 years, it will take a long time to land the patch in the upstream.

sbc100:
> Given that the latest release was 8 years ago and the main maintainer has no public activity last 3 years, it will take a long time to land the patch in the upstream.

Fair point :)

fts/patches/0001-Add-conditional-handling-for-fchdir-and-chdir-functi.patch
```diff
@@ -0,0 +1,102 @@
...
+ 
+ /* fts_build flags */
+ #define	BCHILD		1		/* fts_children */
+@@ -136,6 +143,11 @@ fts_open(char * const *argv, int options,
+ 
+ 	_DIAGASSERT(argv != NULL);
+ 
++#if !HAVE_FCHDIR
++        /* If we don't have fchdir, pretend that FTS_NOCHDIR is always set. */
++        options |= FTS_NOCHDIR;
```
yamt, line 43: doesn't this introduce the behavior changes visible to fts-using application?
i guess it's safer to make fts_open fail unless FTS_NOCHDIR is given.

sunfishcode (approved):
I haven't reviewed the FTS code itself, but overall this looks good to me. If anyone has concerns about the BSD-3-clause license, please let us know, here or in a new issue.

kateinoigakukun:
@sunfishcode @sbc100 Can we merge this into the main branch as we haven't had any license concerns for the past 3 weeks?

sunfishcode:
Merged!

PR #525, opened 2024-08-06
Improve some pthreads stub functions, batch 1

ArcaneNibble:
This is the next part of breaking up #518 into smaller PRs.

This is the rest of the commits which change the existing `THREAD_MODEL=posix` functionality. It:

* _removes_ some functions which are optional and which were already nonfunctional.
* (Continues to) establish a precedent of trying to remain as compatible with Open Group specifications as possible, even when there are major differences in WASI capabilities (i.e. thread cancellation)

Compared to the RFC PR, the `pthread_atfork` stub has been dropped as it is now officially obsolete as of the very recent Issue 8 of the specifications.

### DIFF
--- a/libc-top-half/musl/include/sched.h
+++ b/libc-top-half/musl/include/sched.h
@@ -16,7 +16,6 @@ extern "C" {
 
 #include <bits/alltypes.h>
 
-#ifdef __wasilibc_unmodified_upstream /* WASI has no CPU scheduling support. */
 struct sched_param {
 	int sched_priority;
 	int __reserved1;
@@ -31,6 +30,7 @@ struct sched_param {
 	int __reserved3;
 };
 
+#ifdef __wasilibc_unmodified_upstream /* WASI has no CPU scheduling support. */
 int    sched_get_priority_max(int);
 int    sched_get_priority_min(int);
 int    sched_getparam(pid_t, struct sched_param *);
--- a/libc-top-half/musl/src/thread/pthread_attr_setschedparam.c
+++ b/libc-top-half/musl/src/thread/pthread_attr_setschedparam.c
@@ -2,6 +2,10 @@
 
 int pthread_attr_setschedparam(pthread_attr_t *restrict a, const struct sched_param *restrict param)
 {
+#ifdef __wasilibc_unmodified_upstream
 	a->_a_prio = param->sched_priority;
+#else
+	if (param->sched_priority != 0) return ENOTSUP;
+#endif
 	return 0;
 }
--- a/Makefile
+++ b/Makefile
@@ -300,6 +300,7 @@ LIBC_TOP_HALF_MUSL_SOURCES += \
         thread/pthread_barrierattr_init.c \
         thread/pthread_barrierattr_setpshared.c \
         thread/pthread_cleanup_push.c \
+        thread/pthread_cancel.c \
         thread/pthread_cond_broadcast.c \
         thread/pthread_cond_destroy.c \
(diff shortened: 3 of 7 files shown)


libc-top-half/musl/include/sched.h
```diff
@@ -38,6 +38,10 @@ int    sched_getscheduler(pid_t);
 int    sched_rr_get_interval(pid_t, struct timespec *);
 int    sched_setparam(pid_t, const struct sched_param *);
 int    sched_setscheduler(pid_t, int, const struct sched_param *);
+#else /* ... but this struct isn't optional. */
+struct sched_param {
+	int sched_priority;
+};
```
sbc100, line 44: Should we just use the struct above (i.e. shrink the `ifdef` region above)?

libc-top-half/musl/src/thread/pthread_attr_setschedparam.c
```diff
@@ -2,6 +2,11 @@
 
 int pthread_attr_setschedparam(pthread_attr_t *restrict a, const struct sched_param *restrict param)
 {
+#ifdef __wasilibc_unmodified_upstream
 	a->_a_prio = param->sched_priority;
 	return 0;
+#else
+	if (param->sched_priority == 0) return 0;
+	return ENOTSUP;
```
sbc100, line 10: Can we flip this so the error case is first and `return 0` is the last line?  (then that last line can also go outside the if/else)

sbc100 (approved):

libc-top-half/musl/include/sched.h
ArcaneNibble: Done, except I decided to make two separate ifdef regions instead of putting a single close bracket inside a `#else` region (which looked ugly)

libc-top-half/musl/src/thread/pthread_attr_setschedparam.c
ArcaneNibble: Done

libc-top-half/musl/include/sched.h
ArcaneNibble: On second thought, decided to simply keep all of the reserved fields in the struct, `#ifdef`-ing out only the functions. This seems to be what you originally meant?

sbc100: Yup!

sbc100 (approved):
lgtm, but lets wait for @abrown to chip in too

abrown (approved):
@ArcaneNibble, thanks for working on this. One thing I didn't understand: you mention that

> the `pthread_atfork` stub has been dropped as it is now officially obsolete as of the very recent Issue 8 of the specifications.

Which specification? Which issue? Maybe a link would help me understand what you mean here.

Otherwise I think this makes sense. We've talked for some time about adding failing stubs to get more programs to compile and this goes down that path.

ArcaneNibble:
The specification I'm referring to is "The Open Group Base Specifications Issue 8, IEEE Std 1003.1-2024"

The relevant page is at https://pubs.opengroup.org/onlinepubs/9799919799/functions/pthread_atfork.html

As stated in the "Change History" section, `pthread_atfork` is officially marked as obsolete even though it hasn't been completely removed yet. This is because the function is basically impossible to use correctly (as explained in the "Application Usage" section). Because this is a "new" implementation of pthreads, I don't see a strong need for a failing stub for this function.

abrown:
Ok, thanks for that link! And I agree with the rationale for leaving it out.

Issue #485, opened 2024-03-16
main in a library regression


yamt:
recent wasi-libc is broken if you put main in a library.

test case: https://github.com/yamt/garbage/tree/d422b1073125b3fc8f843ce4b12280ab3982fb84/c/main_in_lib
```
spacetanuki% ./build-native.sh
spacetanuki% ./a.out
this is foo
spacetanuki% WASI_SDK=/opt/wasi-sdk-20.0 ./build-wasi.sh
spacetanuki% toywasm --wasi a.out
this is foo
spacetanuki% WASI_SDK=/opt/wasi-sdk-21.0 ./build-wasi.sh
spacetanuki% toywasm --wasi a.out                       
Error: [trap] unreachable executed (4): unreachable at 0001a5
frame[  2] funcpc 0001a5 (<unknown>:undefined_weak:main) callerpc 0022de
  param [0] = 00000001
  param [1] = 00010b30
frame[  1] funcpc 00226c (<unknown>:__main_void) callerpc 0001df
  local [0] = 00010b00
  local [1] = 00010b30
  local [2] = 00010b20
frame[  0] funcpc 0001ab (<unknown>:_start)
  local [0] = 00000000
2024-03-16 14:47:15 (1710568035.864123000): [117042600] instance_execute_func failed with -1
2024-03-16 14:47:15 (1710568035.864618000): [117042600] invoke failed with -1
spacetanuki% 
```

i guess this is a regression caused by https://github.com/WebAssembly/wasi-libc/pull/429

sbc100:
Yes, it looks like its specifically this line: https://github.com/WebAssembly/wasi-libc/pull/429/files#r1290434731.

yamt:
a simpler case (just an empty file w/o main function)

```shell
spacetanuki% cat a.c
spacetanuki% /opt/wasi-sdk-20.0/bin/clang a.c
wasm-ld: error: /Users/yamamoto/wasm/wasi-sdk-20.0/bin/../share/wasi-sysroot/lib/wasm32-wasi/libc.a(__main_void.o): undefined symbol: main
clang-16: error: linker command failed with exit code 1 (use -v to see invocation)
spacetanuki% /opt/wasi-sdk-21.0/bin/clang a.c
spacetanuki% toywasm --wasi a.out
Error: [trap] unreachable executed (4): unreachable at 0000e6
current pc 0000e6
[...]

PR #528, opened 2024-08-19
Add crt2-command.o to provide __main_void

yamt:
cf. https://github.com/WebAssembly/wasi-libc/issues/485


sunfishcode:
Could you describe somewhere how this change fixes the bug?

yamt:
> Could you describe somewhere how this change fixes the bug?

i added some explanation to the PR description.

PR #530, opened 2024-08-29
Preserve timestamps for installed headers to improve incremental builds

kateinoigakukun:
This change makes it so that the timestamps of installed header files are only updated when the contents of the header files change. This change is beneficial for incremental builds of projects depending on wasi-libc. For example, the Swift project builds wasi-libc as part of its build process, and this change reduces the number of build products in the Swift build that are unnecessarily rebuilt.

The following commands can be used to verify the changes:
```bash
make clean && make -j16 && (
  cd sysroot && find . -name "*.h" -type f -exec stat -c '%n %y' {} \;
) > sysroot.v1.timestamp && \
make -j16 && (
  cd sysroot && find . -name "*.h" -type f -exec stat -c '%n %y' {} \;
) > sysroot.v2.timestamp
diff sysroot.v1.timestamp sysroot.v2.timestamp
```
The diff should show no changes in timestamps.

Eventually, unnecessary copyings themselves should be skipped but it requires making the wasi-libc build system to support complete incremental build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -807,13 +807,7 @@ $(LIBWASI_EMULATED_PTHREAD_OBJS) $(LIBWASI_EMULATED_PTHREAD_SO_OBJS): CFLAGS += 
 $(EMMALLOC_OBJS): CFLAGS += \
     -fno-strict-aliasing
 
-include_dirs:
-	#
-	# Install the include files.
-	#
-	mkdir -p "$(SYSROOT_INC)"
-	cp -r "$(LIBC_BOTTOM_HALF_HEADERS_PUBLIC)"/* "$(SYSROOT_INC)"
-
+$(SYSROOT_INC)/bits/alltypes.h: $(LIBC_TOP_HALF_MUSL_DIR)/tools/mkalltypes.sed $(LIBC_TOP_HALF_MUSL_DIR)/arch/wasm32/bits/alltypes.h.in $(LIBC_TOP_HALF_MUSL_DIR)/include/alltypes.h.in
 	# Generate musl's bits/alltypes.h header.
 	mkdir -p "$(SYSROOT_INC)/bits"
 	sed -f $(LIBC_TOP_HALF_MUSL_DIR)/tools/mkalltypes.sed \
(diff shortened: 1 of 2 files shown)


libc-bottom-half/headers/public/__wasi_snapshot.h
kateinoigakukun, line 1: This file was conditionally overwritten by Makefile, but it makes it difficult to generate the file by a regular make rule, so I changed to generate it in Makefile consistently.

Makefile
```diff
@@ -981,7 +988,7 @@ check-symbols: startup_files libc
 
 install: finish
 	mkdir -p "$(INSTALL_DIR)"
-	cp -r "$(SYSROOT)/lib" "$(SYSROOT)/share" "$(SYSROOT)/include" "$(INSTALL_DIR)"
+	cp -p -r "$(SYSROOT)/lib" "$(SYSROOT)/share" "$(SYSROOT)/include" "$(INSTALL_DIR)"
```
sbc100, line 991: How about using `cp -ar` everywhere?

kateinoigakukun: `-a` has subtle behavior differences across implementations (e.g. darwin cp does not preserve hard link but GNU coreutils does.) and it's not a part of POSIX spec unlike `-p` https://pubs.opengroup.org/onlinepubs/9699919799.2018edition/utilities/cp.html
So I prefer `-p` unless we have clear needs for `-a`

Makefile
```diff
@@ -764,32 +764,39 @@ $(LIBWASI_EMULATED_PROCESS_CLOCKS_OBJS) $(LIBWASI_EMULATED_PROCESS_CLOCKS_SO_OBJ
...
 	mkdir -p "$(SYSROOT_INC)/bits"
 	sed -f $(LIBC_TOP_HALF_MUSL_DIR)/tools/mkalltypes.sed \
 	    $(LIBC_TOP_HALF_MUSL_DIR)/arch/wasm32/bits/alltypes.h.in \
 	    $(LIBC_TOP_HALF_MUSL_DIR)/include/alltypes.h.in \
 	    > "$(SYSROOT_INC)/bits/alltypes.h"
 
+$(SYSROOT_INC)/__wasi_snapshot.h:
+	mkdir -p "$(SYSROOT_INC)"
+ifeq ($(WASI_SNAPSHOT), p2)
+	printf '#ifndef __wasilibc_use_wasip2\n#define __wasilibc_use_wasip2\n#endif\n' \
```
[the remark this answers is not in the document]
kateinoigakukun: I just kept the existing way but I don't find any reason not to use echo. Let me do that.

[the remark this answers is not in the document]
kateinoigakukun: Ah, we need to use printf for Windows... https://github.com/WebAssembly/wasi-libc/actions/runs/10618495813/job/29433860327?pr=530#step:9:99

kateinoigakukun:
Closing in favor of https://github.com/WebAssembly/wasi-libc/pull/549

abrown:
@kateinoigakukun, sorry about duplicating this effort; I must have missed this was here! Hopefully #549 does essentially what you need and maybe #552 improves things further. Let me know if the timestamps are still getting modified too much!

kateinoigakukun:
@abrown No problem at all :) I've verified your patch satisfied my need well. Thank you for working on it!

PR #533, opened 2024-09-09
Add SO_BROADCAST and SO_LINGER to the list of defined constants

### LINKED ISSUE #531: Some `SO_*` constants like `SO_BROADCAST` and `SO_LINGER` aren't defined
While WASI does not (yet) support UDP broadcast or TCP linger, it would still be useful if these constants were defined, similar to how all the `IP_*`, `IPV6_*` and `TCP_*` constants are defined, even though WASI doesn't support all the related functionality.

Specifically, this came up while adding socket support to the Rust `wasm32-wasip2` target. Since the part of the implementation that is common to all targets makes use of these constants, adding them to wasi-libc would allow this to be done without introducing conditional compilation to the common implementation.

nickrum:
Instead of just defining the constants which are needed by Rust's standard library, would it make sense to simply expose all `SO_` constants? And if so, would it be a good idea to simply use the definitions from the [original header file](https://github.com/WebAssembly/wasi-libc/blob/1b19fc65ad84b223876c50dd4fcd7d5a08c311dc/libc-top-half/musl/include/sys/socket.h) instead of copying them over?

Fixes #531

### DIFF
--- a/expected/wasm32-wasip2/predefined-macros.txt
+++ b/expected/wasm32-wasip2/predefined-macros.txt
@@ -1776,9 +1776,11 @@
 #define SOL_UDP 17
 #define SOMAXCONN 128
 #define SO_ACCEPTCONN 30
+#define SO_BROADCAST 6
 #define SO_DOMAIN 39
 #define SO_ERROR 4
 #define SO_KEEPALIVE 9
+#define SO_LINGER 13
 #define SO_PROTOCOL 38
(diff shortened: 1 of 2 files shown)


alexcrichton:
Apologies for the last response on this, but I'm going to close this with some expanded rationale [here](https://github.com/WebAssembly/wasi-libc/issues/531#issuecomment-2932745549)

PR #549, opened 2024-11-19
Avoid re-copying include headers

abrown:
This adds a `.stamp` file to the build directory that tracks when all headers have been copied over to the sysroot `include` directory. Previously, the `include_dirs` target was phony, which `make` will run every time, regardless of whether it is needed. This was part of the problem with all of libc being rebuilt at each `make` invocation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,10 @@ ifeq ($(WASI_SNAPSHOT), p2)
 TARGET_TRIPLE = wasm32-wasip2
 endif
 
+# These artifacts are "stamps" that we use to mark that some task (e.g., copying
+# files) has been completed.
+INCLUDE_DIRS := $(OBJDIR)/copy-include-headers.stamp
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -836,7 +845,11 @@ ifeq ($(WASI_SNAPSHOT), p2)
 		> "$(SYSROOT_INC)/__wasi_snapshot.h"
 endif
```
abrown, line 846: @sunfishcode, I actually have a Bash script that we could also use here to move out all of these commands out of the Makefile along with the list of `MUSL_OMIT_HEADERS`. Should I bring that in to this PR?

sunfishcode: That sounds like a good change; would you mind submitting a separate PR for that?

sunfishcode (approved):
This looks good to me; it just needs a merge conflict resolution.

Makefile
abrown: #552

PR #550, opened 2024-11-19
Avoid rebuilding empty placeholder libraries

abrown:
Previously, these were rebuilt each time as a part of the phone `dummy_libs` target. This change only rebuilds them if they don't exist.


sunfishcode (approved):
Looks good!

PR #551, opened 2024-11-20
Avoid rebuilding `check-symbols` artifacts

abrown:
Like other recent PRs, this change uses true file dependencies to avoid rebuilding the files used to check which symbols are defined or not. The true `git diff` test is still run each time, however.


Makefile
```diff
@@ -906,58 +906,57 @@ finish: check-symbols
...
-	# Collect metadata on the sysroot and perform sanity checks.
-	#
+$(SYSROOT_SHARE)/defined-symbols.txt: startup_files libc
 	mkdir -p "$(SYSROOT_SHARE)"
-
-	#
-	# Collect symbol information.
-	#
+	"$(NM)" --defined-only \
+	    $(SYSROOT_LIB)/libc.a $(SYSROOT_LIB)/libwasi-emulated-*.a $(SYSROOT_LIB)/*.o \
```
abrown, line 922: @sunfishcode: one thing that doesn't make a ton of sense is why we collect the defined symbols for `libc.a` and `libwasi-emulated-*.a` but then when we collect the undefined symbols we switch up which files we're looking at: `libc.a` and `libc-*.a`. I would have expected these to be the same files in both cases but that results in slightly different `undefined-symbols.txt` (e.g., `__floatunsitf`).

abrown: We don't want to just check `*.a` and `*.o` in both places?

sunfishcode: That's a good point. I can't think of a reason offhand why they're different.

Makefile
```diff
@@ -921,58 +921,57 @@ finish: check-symbols
...
-	#
-	# Collect symbol information.
-	#
+	"$(NM)" --defined-only \
+	    $(SYSROOT_LIB)/*.a $(SYSROOT_LIB)/*.o \
+	    |grep ' [[:upper:]] ' \
+	    |sed 's/.* [[:upper:]] //' \
+	    |LC_ALL=C sort \
+	    |uniq \
+	    > "$@"
```
abrown, line 942: Also, I can move all these to separate scripts if we want (here or in a different PR).

PR #552, opened 2024-11-20
Move installation of include headers into a separate script (NFC)

abrown:
This change is not meant to change any functionality, only move some Bash-specific logic out of the `Makefile` into its own script: `install-include-headers.sh`. This reduces the perceived complexity of the `Makefile` but the complexity is still there, tucked away in this script. This script also has the advantage that it can be run separately if needed.

This commit comes after a few different attempts at building up `Makefile` lists of headers to copy over along with the various locations they must be copied from. It is certainly possible to do this, but due to how we need to remove some headers from the list, it ends up being easier to just `cp` and then `rm`, which this script retains.

### DIFF
--- a/scripts/install-include-headers.sh
+++ b/scripts/install-include-headers.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+#
+# Install all the necessary necessary include files from wasi-libc
+# (`$WASI_LIBC`) into the sysroot include directory (`$SYSROOT_INC`). It uses a
+# passed target triple (`$TARGET_TRIPLE`) to condition some of the
+# copied/generated files.
+#
+# Usage: SYSROOT_INC=... TARGET_TRIPLE=... ./install-include-headers.sh
+
+set -e
+
+PARENT_DIR=$(dirname $(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd))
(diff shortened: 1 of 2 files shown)


sbc100:
> This change is not meant to change any functionality

I kind of like the llvm (and now emscripten) convention of adding `. NFC` to the end of the PR title for such changes.

scripts/install-include-headers.sh
```diff
@@ -0,0 +1,94 @@
+#!/usr/bin/env bash
+#
+# Install all the necessary necessary include files from wasi-libc
+# (`$WASI_LIBC`) into the sysroot include directory (`$SYSROOT_INC`). It uses a
+# passed target triple (`$TARGET_TRIPLE`) to condition some of the
+# copied/generated files.
+#
+# Usage: SYSROOT_INC=... TARGET_TRIPLE=... ./install-include-headers.sh
+
+set -e
```
sunfishcode, line 10: I recommend adding `set -euo pipefail` to all bash scripts. In particular here, `-u` will catch the case where someone forgets to set `SYSROOT_INC` or `TARGET_TRIPLE`.

sunfishcode (commented):
Overall this looks good; just one minor comment:

sunfishcode (approved):

PR #553, opened 2024-11-20
Move all bindings-related code together (NFC)

abrown:
This change is an effort to move all dependencies of the phony `bindings` target into one place, for clarity.


sunfishcode:
I can dig it.

PR #554, opened 2024-11-21
test: improve test infrastructure

abrown:
This change represents a rather large re-design in how `wasi-libc` builds and runs its tests. Initially, #346 retrieved the `libc-test` repository and built a subset of those tests to give us some amount of test coverage. Later, because there was no way to add custom C tests, #522 added a `smoke` directory which allowed this. But (a) each of these test suites was built and run separately and (b) it was unclear how to add more tests flexibly--some tests should only run on `*p2` targets or `*-threads` targets, e.g.

This change reworks all of this so that all tests are built the same way, in the same place. For downloaded tests like those from `libc-test`, I chose to add "stub tests" that `#include` the original version. This not only keeps all enabled tests in one place, it also allows us to add "directives," C comments that the `Makefile` uses to filter out tests for certain targets or add special compile, link or run flags. These rudimentary scripts, along with other Bash logic I moved out of the Makefile now live in the `scripts` directory.

Finally, all of this is explained more clearly in an updated `README.md`. The hope with documenting this a bit better is that it would be easier for drive-by contributors to be able to either dump in new C tests for regressions they may find or enable more libc-tests. As of my current count, we only enable 40/75 of libc-test's functional tests, 0/228 math tests, 0/69 regression tests, and 0/79 API tests. Though many of these may not apply to WASI programs, it would be nice to explore how many more of these tests can be enabled to increase wasi-libc's test coverage. This change should explain how to do that and, with directives, make it possible to condition how the tests compile and run.


test/scripts/failed-tests.sh
```diff
@@ -0,0 +1,24 @@
...
+#
+# Usage: failed-tests.sh [TESTDIR]
+#
+# Optionally set the VERBOSE environment variable to see the list of failed
+# tests.
+
+set -e
+
+TESTDIR=${1:-.}
+FAILED=$(find $TESTDIR -type f -and -name *.log -and -not -size 0)
```
sbc100, line 14: The comment above says `.err` but here you are looking for `.log`?

test/scripts/run-test.sh
```diff
@@ -0,0 +1,22 @@
...
+ENGINE="${ENGINE:-wasmtime}"
+[ -n "$WASM" ] || (echo "missing WASM variable" && exit 1)
+[ -n "$DIR" ] || (echo "missing DIR variable" && exit 1)
+
+pushd $DIR > /dev/null
+    mkdir -p fs
+    echo "$ENGINE $WASM" > cmd.sh
+    chmod +x cmd.sh
+    ./cmd.sh &> output.log
+    [ $? -eq 0 ] || echo "Test failed" >> output.log
```
sbc100, line 20: It looks like the string "Test failed" at the end of `output.log` is the signal that the test failed?  Is that right?

test/scripts/run-test.sh
```diff
@@ -0,0 +1,22 @@
...
+popd > /dev/null
+
```
sbc100, line 22: Trailing newline here.

sbc100 (commented):
Wow!  Very nice setup.  Kind of reminds me of llvm's lit/filecheck system.

test/scripts/run-test.sh
abrown: It's actually if there's any output at all; appending this string just makes sure of that.

test/README.md
```diff
@@ -1,6 +1,85 @@
...
+$ make TARGET_TRIPLE=...
+Tests passed
+```
+
+Note that `wasm-ld` must be available, so an additional
+`CC=<wasi-sdk>/bin/clang` may be necessary. Each test runs in a directory that
+looks like (see [`run-test.sh`]):
+
+```sh
+$ ls run/$TARGET_TRIPLE/misc/some-test.c
```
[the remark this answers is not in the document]
abrown: Yeah, I guess that's true. I was trying to make it clear which test it was coming from... but you're right that it just looks confusing.

test/README.md
```diff
@@ -1,6 +1,85 @@
...
+
+### Adding tests
+
+To add a test, create a new C file in [`src/misc`]:
+
+```c
+// filter.py(TARGET_TRIPLE): !wasm32-wasip2
+// add-flags.py(CFLAGS): ...
+// add-flags.py(LDFLAGS): ...
+// add-flags.py(RUN): ...
```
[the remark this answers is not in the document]
abrown: Agreed on the `//!` bit. Still mulling over the `.py` suggestion: one thing that always bugs me with infrastructure stuff like this is that I don't know where to look if something goes wrong. I figured if I put the `.py` extension on there most of us would think, "oh, I see, this logic is from some file... let's bust out `find`..." But that may not be as clear as I think?

abrown:
> Wow! Very nice setup. Kind of reminds me of llvm's lit/filecheck system.

Heh, a poor man's version unfortunately, but it should work ok to enable more tests. I think one future thing we'll want to add is the ability to express "_if_ we're compiling target X, _then_ add flags Y and Z." But I didn't bother with that just yet.

test/scripts/run-test.sh
abrown: All of my files have that to avoid GitHub's red-line warnings.

test/Makefile
```diff
@@ -9,37 +9,43 @@
...
+
 ##### DOWNLOAD #################################################################
 
 LIBC_TEST_URL ?= https://github.com/bytecodealliance/libc-test
 LIBC_TEST = $(DOWNDIR)/libc-test
 LIBRT_URL ?= https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-24/libclang_rt.builtins-wasm32-wasi-24.0.tar.gz
 LIBRT = $(DOWNDIR)/libclang_rt.builtins-wasm32.a
 WASMTIME_URL ?= https://github.com/bytecodealliance/wasmtime/releases/download/v26.0.1/wasmtime-v26.0.1-x86_64-linux.tar.xz
-WASMTIME = $(DOWNDIR)/$(shell basename $(WASMTIME_URL) .tar.xz)/wasmtime
+WASMTIME = $(abspath $(DOWNDIR)/$(shell basename $(WASMTIME_URL) .tar.xz)/wasmtime)
```
sbc100, line 41: Huh, TIL `basename` took and optional suffix

test/Makefile
```diff
@@ -61,151 +67,149 @@ $(WASM_TOOLS): | $(DOWNDIR)
...
-	$(LIBC_TEST)/src/functional/wcstol.c
-
-# Part of the problem including more tests is that the `libc-test`
-# infrastructure code is not all Wasm-compilable. As we include more tests
-# above, this list will also likely need to grow.
-COMMON_TEST_INFRA = \
-	$(LIBC_TEST)/src/common/path.c \
+INFRA_OBJDIR := $(OBJDIR)/common
+$(INFRA_OBJDIR):
+	mkdir -p $@
```
sbc100, line 87: You use `@mkdir` above but just `mkdir` here.  Is that intended?

test/Makefile
```diff
@@ -61,151 +67,149 @@ $(WASM_TOOLS): | $(DOWNDIR)
...
+# Build the common test infrastructure. Part of the problem including more tests
+# is that the `libc-test` infrastructure code is not all Wasm-compilable. As we
+# include more tests above, this list will also likely need to grow.
+INFRA_FILES = $(LIBC_TEST)/src/common/path.c \
 	$(LIBC_TEST)/src/common/print.c \
 	$(LIBC_TEST)/src/common/rand.c \
 	$(LIBC_TEST)/src/common/utf8.c
+$(INFRA_FILES): $(DOWNLOADED)
+INFRA_WASM_OBJS := $(patsubst $(LIBC_TEST)/src/common/%.c,$(OBJDIR)/common/%.wasm.o,$(INFRA_FILES))
+$(OBJDIR)/common/%.wasm.o: $(LIBC_TEST)/src/common/%.c | $(INFRA_OBJDIR)
```
sbc100, line 98: What does the vertical bar here in the rule do?

test/src/libc-test/functional/argv.c
```diff
@@ -0,0 +1,2 @@
+//! add-flags.py(CFLAGS): -I.
+#include "build/download/libc-test/src/functional/argv.c"
```
sbc100, line 2: Should we add a comment to all these that says something like `Auto-generated by genrate-stubs.sh`?

test/scripts/run-test.sh
sbc100: This one seems to have an extra one though.

sbc100 (approved):

test/Makefile
abrown: 😁

test/src/libc-test/functional/argv.c
abrown: Good idea, added.

test/Makefile
abrown: It's an [order-only prerequisite](https://www.gnu.org/software/make/manual/make.html#Prerequisite-Types): `make` ensures the prerequisite is built but it doesn't worry about when it has been updated; for some reason I've found this helpful with directories when I don't want to trigger a rebuild of the actual objects.

PR #555, opened 2024-11-23
Avoid re-copying the startup object files

abrown:
This adds a `.stamp` file to the process of copying over the `crt` files to the sysroot directory to avoid re-copying on each `make` invocation.


sunfishcode:
Thanks!

PR #560, opened 2025-01-19
test: enable pthread-related libc-tests

kateinoigakukun:
This change enables the pthread-related tests in the libc-test suite. The tests are enabled only for the `wasm32-wasip1-threads` target, which is the only target that supports threads at the moment.

The following pthread tests are still disabled:
- pthread_cancel-points.c
- pthread_cancel.c
- pthread_robust.c

This is a preparative change for https://github.com/swiftwasm/swift/issues/5598

### DIFF
--- a/test/src/libc-test/functional/pthread_cond.c
+++ b/test/src/libc-test/functional/pthread_cond.c
@@ -1,0 +1,5 @@
+//! filter.py(TARGET_TRIPLE): wasm32-wasip1-threads
+//! add-flags.py(CFLAGS): -I.
+//! add-flags.py(LDFLAGS): -Wl,--import-memory,--export-memory,--shared-memory,--max-memory=1073741824
+//! add-flags.py(RUN): --wasi threads
+#include "build/download/libc-test/src/functional/pthread_cond.c"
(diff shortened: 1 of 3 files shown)


test/Makefile
```diff
@@ -129,6 +129,7 @@ CFLAGS ?= --target=$(TARGET_TRIPLE) --sysroot=$(SYSROOT_DIR)
 # Always include the `libc-test` infrastructure headers.
 CFLAGS += -I$(INFRA_HEADERS_DIR)
 ifneq ($(findstring -threads,$(TARGET_TRIPLE)),)
+LDFLAGS += -Wl,--import-memory,--export-memory,--shared-memory,--max-memory=1073741824
```
abrown, line 132: Did you find that doing this same thing via the `//! add-flags.py(LDFLAGS): ...` annotation in the test files didn't work?

test/Makefile
```diff
@@ -179,6 +180,7 @@ ifeq ($(TARGET_TRIPLE), wasm32-wasip2)
 ENGINE += --wasm component-model
 OBJPAT := $(OBJDIR)/%.component.wasm
 else
+ENGINE += --wasi threads
```
abrown, line 183: Similarly, does the `//! add-flags.py(RUN): ...` annotation not work here?

test/Makefile
kateinoigakukun: Yes, it should work but I did it in this way as I think those flags are not a property of test cases but a property of the compiler target. Same with the engine flags.

abrown: I see what you mean. Practically, though, it might be more convenient to move this to the test files since there may be discrepancies between targets; e.g., what if other threads tests don't require setting up 1G of max memory?

abrown: And, honestly, the simpler we can make this `Makefile`, the better.

kateinoigakukun: Ok, it makes sense to me 👍

abrown (approved):
Thank you for this!

PR #563, opened 2025-01-21
test: add browser test harness

kateinoigakukun:
This commit adds a browser test harness to run the tests in the browser.

## Motivation

We are heavily using wasi-libc on browsers but we don't have any test for the case in wasi-libc. In theory, there should be no behavioral difference between wasmtime and browsers  (as long as WASI implementations are correct) but browsers have their own limitations in their Wasm engines. For example, memory.atomic.wait32 is not permitted on the main thread.

We are working on adding some mitigation for such browser-specific issues (https://github.com/WebAssembly/wasi-libc/pull/562) and this test harness will help to validate the fix.

### DIFF
--- a/test/scripts/browser-test/run-test.mjs
+++ b/test/scripts/browser-test/run-test.mjs
@@ -1,0 +1,230 @@
+/**
+ * This script is served by `harness.mjs` and runs in the browser.
+ */
+import { WASI, File, OpenFile, ConsoleStdout, PreopenDirectory } from 'https://cdn.jsdelivr.net/npm/@bjorn3/browser_wasi_shim@0.3.0/+esm'
+import { polyfill } from 'https://cdn.jsdelivr.net/npm/wasm-imports-parser@1.0.4/polyfill.js/+esm';
+
+/**
+ * @param {{
+ *   module: WebAssembly.Module,
+ *   addToImports: (importObject: WebAssembly.Imports) => Promise<void>
+ * }}
+ */
+export async function instantiate({ module, addToImports }) {
+    const args = ["target.wasm"]
+    const env = []
+    const fds = [
+        new OpenFile(new File([])), // stdin
+        ConsoleStdout.lineBuffered((stdout) => {
+            console.log(stdout);
+        }),
+        ConsoleStdout.lineBuffered((stderr) => {
+            console.error(stderr);
+        }),
+    ];
+    const wasi = new WASI(args, env, fds);
+
+    const importObject = {
+        wasi_snapshot_preview1: wasi.wasiImport,
+    };
+    await addToImports(importObject);
+    const instance = await WebAssembly.instantiate(module, importObject);
+    return { wasi, instance };
+}
+
+class WorkerInfoView {
+    /**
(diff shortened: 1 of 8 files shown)


test/scripts/browser-test/run-test.mjs
```diff
@@ -0,0 +1,231 @@
...
+    const env = []
+    const fds = [
+        new OpenFile(new File([])), // stdin
+        ConsoleStdout.lineBuffered((stdout) => {
+            console.log(stdout);
+        }),
+        ConsoleStdout.lineBuffered((stderr) => {
+            console.error(stderr);
+        }),
+        new PreopenDirectory("/", new Map()),
```
abrown, line 24: Shouldn't there be some way to use the preopens we pass along in the `--dir` CLI args... or no?

How do the tests that need special access work?! Or maybe that's just the `fts.c` test` that we skip in browser?

abrown (approved):
I ran all of this locally to understand what is going on and I think this makes sense. It does increase the amount of code to maintain but it seems worth it in order to check this in the browser. Let's hope the CDN artifacts don't cause an issue during CI (they shouldn't, right?) but if they do we could figure out workarounds.

Thanks @kateinoigakukun for working on this!

test/scripts/browser-test/run-test.mjs
kateinoigakukun: Good catch. The preopen is unnecessary and I don't think it's even worth adding `--dir` emulation support to the harness. File system emulation is not part of wasi-libc's responsibility and it's not the main focus here, so I just skipped tests using `--dir` for now.

test/scripts/browser-test/run-test.mjs
```diff
@@ -0,0 +1,230 @@
+/**
+ * This script is served by `harness.mjs` and runs in the browser.
+ */
+import { WASI, File, OpenFile, ConsoleStdout, PreopenDirectory } from 'https://cdn.jsdelivr.net/npm/@bjorn3/browser_wasi_shim@0.3.0/+esm'
+import { polyfill } from 'https://cdn.jsdelivr.net/npm/wasm-imports-parser@1.0.4/polyfill.js/+esm';
```
sbc100, line 5: It seems like we are taking a dependency here this WASI implementation.    Maybe this should be documented somewhere?   What happens if this breaks?   Who maintains this WASI implementation?

kateinoigakukun: Added a sentence about the dependency to test/README.md: https://github.com/WebAssembly/wasi-libc/pull/576
If the dependency breaks, wasi-libc's CI can fail, but nothing affects the users of wasi-libc. Although it's maintained under [a personal namespace](https://github.com/bjorn3/browser_wasi_shim/), the implementation is the most reliable one on browser today as far as I know.

abrown: FWIW, I see this browser testing infrastructure as useful but experimental. If we were to run into many problems consistently, I think it would make sense to rip it out.  (and I reviewed it with that in mind). But I'm hopeful that things are stable enough at this point that that won't be a problem.

PR #565, opened 2025-02-02
Pin MacOS Github runner version in ci to MacOS 15

mcbarton:
Rather than using Macos-latest this makes the ci used a pinned version of Macos (set to the latest version available). This is so if you happen to have an issue with the MacOS build you know exactly which version of MacOS it is running, rather than having to dig into the ci, or Github runner documentation.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


sbc100 (approved):
lgtm, but can you add the motivation for the change to the PR description?

mcbarton:
> lgtm, but can you add the motivation for the change to the PR description?

Added to the PR description.

PR #567, opened 2025-02-02
Update Ubuntu version in ci to Ubuntu 24.04

mcbarton:
This PR updates the version of Ubuntu used in the ci to the latest version available for a Github runner Ubuntu 24.04.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,12 +7,12 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-22.04, macos-latest, windows-latest]
+        os: [ubuntu-24.04, macos-latest, windows-latest]
         clang_version: [10.0.0]
         # use different LLVM versions among oses because of the lack of
         # official assets on github.
         include:
-          - os: ubuntu-22.04
+          - os: ubuntu-24.04
             clang_version: 10.0.0
             llvm_asset_suffix: x86_64-linux-gnu-ubuntu-18.04
           - os: macos-latest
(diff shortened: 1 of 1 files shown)


mcbarton:
@alexcrichton Does wasm-libc need libtinfo5 explicity, or can I update this PR to use libtinfo6 rather than the workaround currently in the PR?

sbc100:
> @alexcrichton @sbc100 Does wasm-libc need libtinfo5 explicity, or can I update this PR to use libtinfo6 rather than the workaround currently in the PR?

I believe the dependency on libinfo comes from clang/llvm not from wasi libc specifically.    What happens if you simply remove the installation of libtinfo5?  I assume some pre-built version of llvm will fail to execute?

sbc100:
Yes, it looks like the step below downloaded https://github.com/llvm/llvm-project/releases/download/llvmorg-${{ matrix.clang_version }}/clang+llvm-${{ matrix.clang_version }}-${{ matrix.llvm_asset_suffix }}.tar.xz.   Can you confirm that the version of llvm we are using still depends on libtinfo5?

sbc100:
It looks like if we drop support for llvm versions old than 17 we could also drop support for tinfo5.    As as long we want to continue to support llvm versions 16.0.0 and 10.0.0 then we should continue to test on the older version of ubuntu that has the right shared libraries.

sbc100 (changes requested):
I think we should close this PR and consider this again if/when we decide to drop support for older llvm versions.

mcbarton:
Closing PR as requested, and will reopen if support for older versions of llvm are dropped.

PR #568, opened 2025-02-02
[ci] Add Ubuntu arm Github runner

mcbarton:
This PR adds the recently generally available Ubuntu arm Github runners to the ci.


.github/workflows/main.yml
```diff
@@ -30,6 +30,14 @@ jobs:
           - os: windows-latest
             clang_version: 16.0.0
             enable_pic: true
+          - os: ubuntu-24.04-arm
+            clang_version: 10.0.0
+            llvm_asset_suffix: x86_64-linux-gnu-ubuntu-18.04
+          - os: ubuntu-24.04-arm
+            clang_version: 16.0.0
+            llvm_asset_suffix: x86_64-linux-gnu-ubuntu-18.04
+            enable_pic: true
```
sbc100, line 39: I think maybe just one arm config might be enough.  Maybe drop the first one?

mcbarton: Done.

mcbarton:
Set to draft until I can work out how to get libtinfo5 on Ubuntu 24.04 arm. Will debug in a vm locally.

mcbarton:
@sbc100 @alexcrichton Before I look into the failing tests, can you both take a look, and see if there is an easy fix? I've changed the PR so that it is editable by maintainers.

.github/workflows/main.yml
```diff
@@ -64,7 +75,11 @@ jobs:
     - name: Install LLVM tools (MacOS)
       shell: bash
       run: |
-        curl -sSfL https://github.com/llvm/llvm-project/releases/download/llvmorg-${{ matrix.clang_version }}/clang+llvm-${{ matrix.clang_version }}-${{ matrix.llvm_asset_suffix }}.tar.xz | tar xJf -
+        if [[ "${{ matrix.os }}" == *"arm"* ]]; then
+          curl -sSfL https://github.com/llvm/llvm-project/releases/download/llvmorg-${{ matrix.clang_version }}/clang+llvm-${{ matrix.clang_version }}-${{ matrix.llvm_asset_suffix }}.tar.xz | tar xJf -
+        else
+          curl -sSfL https://github.com/llvm/llvm-project/releases/download/llvmorg-${{ matrix.clang_version }}/clang+llvm-${{ matrix.clang_version }}-${{ matrix.llvm_asset_suffix }}.tar.xz | tar xJf -
```
sbc100, line 81: I don't see any differences between the if and else branch here..  am I missing something?

sbc100:
Looks like the problem is likely wrong version of wasmtime being downloaded: `Saving to: ‘build/download/wasmtime-v26.0.1-x86_64-linux.tar.xz’.

.github/workflows/main.yml
mcbarton: It was the same I just hadn't noticed to after I changed llvm_asset_suffix. Reverted this change.

test/Makefile
```diff
@@ -37,9 +37,18 @@ LIBC_TEST_URL ?= https://github.com/bytecodealliance/libc-test
...
+	WASMTIME_URL ?= https://github.com/bytecodealliance/wasmtime/releases/download/v26.0.1/wasmtime-v26.0.1-aarch64-linux.tar.xz 
+else
+	WASMTIME_URL ?= https://github.com/bytecodealliance/wasmtime/releases/download/v26.0.1/wasmtime-v26.0.1-x86_64-linux.tar.xz
+endif
 WASMTIME = $(abspath $(DOWNDIR)/$(shell basename $(WASMTIME_URL) .tar.xz)/wasmtime)
-WASM_TOOLS_URL ?= https://github.com/bytecodealliance/wasm-tools/releases/download/v1.220.0/wasm-tools-1.220.0-x86_64-linux.tar.gz
+ifeq ($(UNAME_D),aarch64)
+	WASM_TOOLS_URL ?= https://github.com/bytecodealliance/wasm-tools/releases/download/v1.220.0/wasm-tools-1.220.0-aarch64-linux.tar.gz 
+else
+	WASM_TOOLS_URL ?= https://github.com/bytecodealliance/wasm-tools/releases/download/v1.220.0/wasm-tools-1.220.0-x86_64-linux.tar.gz
```
sbc100, line 50: Can you instead define an `ARCH` variable and use that in the definition of the URLs, avoiding the need to duplicate the URLs themselves?

mcbarton: Done.

test/Makefile
```diff
@@ -37,9 +37,10 @@ LIBC_TEST_URL ?= https://github.com/bytecodealliance/libc-test
 LIBC_TEST = $(DOWNDIR)/libc-test
 LIBRT_URL ?= https://github.com/WebAssembly/wasi-sdk/releases/download/wasi-sdk-24/libclang_rt.builtins-wasm32-wasi-24.0.tar.gz
 LIBRT = $(DOWNDIR)/libclang_rt.builtins-wasm32.a
-WASMTIME_URL ?= https://github.com/bytecodealliance/wasmtime/releases/download/v26.0.1/wasmtime-v26.0.1-x86_64-linux.tar.xz
+ARCH := $(shell uname -m)
+WASMTIME_URL ?= https://github.com/bytecodealliance/wasmtime/releases/download/v26.0.1/wasmtime-v26.0.1-$(ARCH)-linux.tar.xz 
 WASMTIME = $(abspath $(DOWNDIR)/$(shell basename $(WASMTIME_URL) .tar.xz)/wasmtime)
-WASM_TOOLS_URL ?= https://github.com/bytecodealliance/wasm-tools/releases/download/v1.220.0/wasm-tools-1.220.0-x86_64-linux.tar.gz
+WASM_TOOLS_URL ?= https://github.com/bytecodealliance/wasm-tools/releases/download/v1.220.0/wasm-tools-1.220.0-$(ARCH)-linux.tar.gz
```
sbc100, line 43: Looks like maybe you added a trailing space here?

.github/workflows/main.yml
```diff
@@ -30,16 +30,27 @@ jobs:
...
         submodules: true
 
-    - name: Install libtinfo5
+    - name: Install libtinfo
       run: |
         set -ex
         sudo apt-get update
-        sudo apt-get install -y libtinfo5
+        # Workaround to get libtinfo5 installed on Ubuntu 24.04 taken from 
+        # https://askubuntu.com/questions/1531760/how-to-install-libtinfo5-on-ubuntu24-04
```
sbc100: Remove this comment, it seem inaccurate now?

sbc100 (approved):

.github/workflows/main.yml
mcbarton: Done as you was putting in the comment.

test/Makefile
mcbarton: Done.

.github/workflows/main.yml
```diff
@@ -30,16 +30,25 @@ jobs:
...
         submodules: true
 
-    - name: Install libtinfo5
+    - name: Install libtinfo
       run: |
         set -ex
         sudo apt-get update
-        sudo apt-get install -y libtinfo5
+        if [[ "${{ matrix.os }}" == *"24"* ]]; then
+          sudo apt-get install -y libtinfo6
```
sbc100, line 48: It looks like this is not actually necessary since its already installed (`libtinfo6 is already the newest version (6.4+20240113-1ubuntu2).`)

Maybe instead you could update the  `if:` line below to say `if: startsWith(matrix.os, 'ubuntu-22')`?

sbc100: Then you could revert both `name` and `run` keys here

mcbarton: Done.

sbc100 (approved):

PR #569, opened 2025-02-02
Update wasi-headers/WASI submodule to latest tagged version (v0.2.3)

mcbarton:
The WASI submodule references a commit which is 3 years old. This PR updates it to use the latest tagged version (v0.2.3) which was tagged 2 months ago.

### DIFF
--- a/tools/wasi-headers/Cargo.toml
+++ b/tools/wasi-headers/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "wasi-headers"
-version = "0.0.1"
+version = "0.2.3"
 authors = ["Dan Gohman <sunfish@mozilla.com>", "Pat Hickey <phickey@fastly.com>"]
 license = "Apache-2.0"
 edition = "2018"


mcbarton:
@alexcrichton @sbc100 pinging for review, and hopefully approval of the workflow run.

mcbarton:
@alexcrichton pinging for review

alexcrichton:
@mcbarton you are generating quite a lot of PRs on wasi-sdk and wasi-libc. I'm not paid to work on this repository full-time, nor are others I believe. Your PRs are in my inbox and I'm going to get to them when I get to them, there's no need to actively ping each PR within hours of being opened or sooner. It takes time to review PRs and in OSS review can't always happen according to your schedule. For example I'm currently travelling out-of-country and am very busy with other activities and am not available to review.

I would ask that you be a bit more patient with your review requests and not quickly and actively ping for review (unless it's 2+ weeks after the initial PR). Folks will try to get to your PR when reasonable, but frequently pinging on PRs is only going to frustrate both maintainers and yourself. The reality is that maintenance is provided to you for free just as you're providing this PR to us for free. I want to review your work, but please be understanding that there's not something like an SLA which states that reviews are guaranteed to happen within a certain time period.

alexcrichton:
Could you elaborate more on the rationale for doing this? It's true the submodule is old, but it's also true that the `*.witx` files for WASIp1 are also just as old and haven't changed in the interim. Additionally they're unlikely to ever change again at this point so there's no really much need to update things. If anything the submodule should be dropped in favor of just checking in the generated header files at this point if a change must be made.

That's why I'm curious if you have further motivations beyond "update all the things" because otherwise this change, while it's not bad by any means, doesn't really accomplish much. Additionally the version of the bindings generator doesn't really need to be updated since that's just a version marker required in Rust/Cargo and doesn't actually correspond to anything.

mcbarton:
> Could you elaborate more on the rationale for doing this? It's true the submodule is old, but it's also true that the `*.witx` files for WASIp1 are also just as old and haven't changed in the interim. Additionally they're unlikely to ever change again at this point so there's no really much need to update things. If anything the submodule should be dropped in favor of just checking in the generated header files at this point if a change must be made.
> 
> That's why I'm curious if you have further motivations beyond "update all the things" because otherwise this change, while it's not bad by any means, doesn't really accomplish much. Additionally the version of the bindings generator doesn't really need to be updated since that's just a version marker required in Rust/Cargo and doesn't actually correspond to anything.

There was no real motivation  for this PR and the other update things PRs beyond I'm just trying to help contribute to the web assembly tools I've been try to make use of these last few months. I just saw it as something that may have been overlooked while maintainers were busy making bigger changes. I'm not familiar enough with the ins and outs of how the web assembly repos are related at the moment, but I'm trying to work it out so I can put more substantial contributions to the organisation in future. If the update is not needed, then I am happy to close the PR and look for other improvements.

alexcrichton:
Ok sounds reasonable, and thanks! In that case for this submodule specifically I think it's reasonable to leave it as-is. In the future I think this'll naturally get deleted as it's replaced with a vendored copy of the header so there's no need to generate it from source any more.

PR #570, opened 2025-02-03
Update MacOS-15 ci job building against llvm 15.0.7 to llvm 16.0.0

mcbarton:
The ci jobs are either building against 10.0.0 or 16.0.0 except the MacOS job which builds against 15.0.7 . The PR updates the MacOS so it is consistent with the llvms used on the other operating systems.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ jobs:
             llvm_asset_suffix: x86_64-linux-gnu-ubuntu-18.04
             enable_pic: true
           - os: macos-15
-            clang_version: 15.0.7
+            clang_version: 16.0.0
             llvm_asset_suffix: x86_64-apple-darwin21.0
           - os: windows-2025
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -25,7 +25,7 @@ jobs:
             llvm_asset_suffix: x86_64-linux-gnu-ubuntu-18.04
             enable_pic: true
           - os: macos-15
-            clang_version: 15.0.7
+            clang_version: 16.0.0
```
sbc100, line 28: I think the reason for the different version here is a deliberate choice to make sure wasi-libc is compatible with different versions of llvm.   If you want to propose dropping support for a certain older versions then please open a bug and we can discuss it there.

mcbarton: It wasn't so much about dropping support. I just didn't understand why it doesn't match the llvm version targeted for other operating systems. If its deliberate that it is different then I will close the PR.

PR #575, opened 2025-02-04
Update ADAPTER_URL to reference v29.0.1

mcbarton:
This PR updates ADAPTER_URL to reference the latest release version (versions 29.0.1)


mcbarton:
@alexcrichton @sbc100 pinging for review

abrown (approved):

abrown (commented):
Actually, on second thought, I wonder what you're trying to accomplish here? Were you seeing some kind of issue or just updating as a best practice? There's a Wasmtime version a few lines up that you might want to change as well. (And we could create an extra variable to control both or something like that, but I'm not too worried about that).

[edit: never mind I see the other PR doing that now!]

PR #581, opened 2025-05-08
Preserve file timestamps when copying files in the install target

kateinoigakukun:
This change makes it so that the timestamps of installed header files are only updated when the contents of the header files change. This change is beneficial for incremental builds of projects depending on wasi-libc. For example, the Swift project builds wasi-libc as part of its build process, and this change reduces the number of build products in the Swift build that are unnecessarily rebuilt.

This is a 2nd attempt of https://github.com/WebAssembly/wasi-libc/pull/530 but much simpler thanks to @abrown 's build work

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -816,7 +816,7 @@ endif
 
 install: finish
 	mkdir -p "$(INSTALL_DIR)"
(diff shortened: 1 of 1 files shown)


abrown (approved):
Makes sense to me!

kateinoigakukun:
Thanks! Could you merge this branch to main?

kateinoigakukun:
Thanks so much!

PR #584, opened 2025-05-25
clang > 20.1.4 compat: unterminated strings

Lukasdoe:
clang introduced a new warning (following gcc) for string initialization to char arrays that do not fit the null terminal. 

Introduced here: https://github.com/llvm/llvm-project/pull/137829
Included in LLVM version [LLVM 20.1.4](https://github.com/llvm/llvm-project/releases/tag/llvmorg-20.1.4)+

More detail:
-Wunterminated-string-initialization warns for c string literals that are written to a buffer too small to include a null terminator. This warning is also included in -Wextra which is enabled for wasi-libc and fails compilation due to -Werror.

Possible fixes:
 1. deactivate the warning
 2. add `__attribute__ ((nonstring))` to the three occurrences where this is a problem.

Chosen solution: 1.
Since the attribute is not available on older clang versions, it would trigger another warning there.

(included a small formatting change in the commit - pls ignore)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -422,7 +422,8 @@ CFLAGS += -Wall -Wextra -Werror \
   -Wno-missing-braces \
   -Wno-ignored-pragmas \
   -Wno-unused-but-set-variable \
-  -Wno-unknown-warning-option
+  -Wno-unknown-warning-option \
+  -Wno-unterminated-string-initialization
 
 # Configure support for threads.
 ifeq ($(THREAD_MODEL), single)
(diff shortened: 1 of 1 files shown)


yamt:
>  Since the attribute is not available on older clang versions, it would trigger another warning there.

have you considered __has_attribute?

yamt:
> the three occurrences where this is a problem.

for those who wonder where they are:
```
libc-top-half/musl/src/stdio/vfprintf.c:170:2: warning: initializer-string for character array is too long, array size is 16 but initializer has size 17 (including the null terminating character); did you mean to use the 'nonstring' attribute? [-Wunterminated-string-initialization]
  170 |         "0123456789ABCDEF"
      |         ^~~~~~~~~~~~~~~~~~
libc-top-half/musl/src/locale/pleval.c:90:31: warning: initializer-string for character array is too long, array size is 11 but initializer has size 12 (including the null terminating character); did you mean to use the 'nonstring' attribute? [-Wunterminated-string-initialization]
   90 |         static const char opch[11] = "|&=!><+-*%/";
      |                                      ^~~~~~~~~~~~~
libc-top-half/musl/src/locale/pleval.c:91:31: warning: initializer-string for character array is too long, array size is 6 but initializer has size 7 (including the null terminating character); did you mean to use the 'nonstring' attribute? [-Wunterminated-string-initialization]
   91 |         static const char opch2[6] = "|&====";
      |                                      ^~~~~~~~
```

abrown (approved):

PR #588, opened 2025-06-18
Refactor how CI works in this repository

### LINKED ISSUE #587: ci: improve the testing of multiple configurations
In #586, @ncruces is running into limitations with how CI is set up for this repository: it is difficult (impossible?) to test wasi-libc when compiled with different configurations than the default. To me, the issue boils down to this kind of CI job:

https://github.com/WebAssembly/wasi-libc/blob/4720b34e91ccca1b3d4d1df3fd20e6b578898d6a/.github/workflows/main.yml#L133-L144

We set up the compilers, build all of the libc targets, and then test them all sequentially. I believe this is happening because:
1. setting up Clang is a bit tricky for each runner so we just want to do it once; we could fix this by creating a custom action encapsulating this CI logic
2. we need to package up the built artifacts for all of the targets into one `sysroot-*.tgz` file for wasi-sdk to consume; I'm not sure what to do about this but possibly wasi-sdk could retrieve more artifacts (?)

Ideally, the choice of target to build and other compiler flags should be part of the build matrix. This would (a) make it easier for us to add a new matrix line to build and test the SIMD functionality, e.g., but also (b) would parallelize the build, speeding up CI. It might also have the benefit of (c) clearly identifying which artifacts wasi-sdk pulls in, because looking at `main.yml` I worry that we may not always be picking up the right artifacts (e.g., `enable_pic: true`?). Am I missing any other rationale to why things are the way they are?

### Solutions

Here are some things I've thought through, but I'm looking for others as well:

- __merge repositories__: we could merge wasi-sdk and wasi-libc as suggested in #427. This could significantly reduce the Clang set up since we're building it in wasi-sdk regardless. I'm not sure it totally solves the problem because we would still want to be able to test multiple configurations of wasi-libc.
- __create a custom setup action__: this would make it easier to add new, separate CI jobs but we likely don't want separate CI jobs
- __package artifacts per target__: if we create separate `sysroot-*.tgz` files per WASI target and per build configuration, we can both parallelize the build and easily add new build steps. I'm not sure how we would uniquely identify the build configuration, though; perhaps we hash something, or force ourselves to add an ID in `main.yml`. If we had that, though, would something like the following work: `sysroot-{os}-clang-{clang-version}-{wasi target}-{build config id}.tgz`? @alexcrichton, would we then need to understand the difference in wasi-sdk's CMake files? Or no, since it is built again there?

Any feedback is appreciated here as we think through how to enable better testing of wasi-libc.

alexcrichton:
There are a number of changes in this commit aimed at addressing #587 and making it easier to test #586 in CI. Notable changes here are:

* The matrix of what to test is much different from before. One matrix entry now builds just one target and optionally tests that target.

* The CI matrix ensures that wasi-libc builds on a variety of platforms, e.g. Windows/macOS/Linux as well as Linux aarch64.

* The CI matrix has one entry for building with an older version of LLVM. This version was bumped from LLVM to LLVM 11 since LLVM is installed through `apt-get`, not through downloads any more.

* On Linux LLVM/Clang are downloaded through `apt-get` instead of from llvm-project binaries to avoid dealing with `libtinfo5` and dependencies.

* The CI matrix has a test job per-target. This can be expanded/shrunk as necessary but currently everything is tested with LLVM 16 (as before) and only on Linux (also as before). The test run is seqeunced to happen after the build of libc itself.

* The CI matrix has split out V8 headless tests into their own job to avoid running multiple suites of tests in a single job.

* Installation of LLVM is refactored to a separate action to reduce the noise in `main.yml`.

* Setting `TARGET_TRIPLE` can now be done through environment variables as opposed to only through arguments to `make`.

* Handling of `BULITINS_LIB` has improved. Previously the build target for `libc_so` would modify the compiler's resource directory and this is updated to use a custom directory in `OBJDIR`.

* Arranging compiler-rt for tests is now done with `-resource-dir` instead of copying the directory into the system compiler's location.

Overall it's the intention that no amount of testing is lost in this PR. The goal is to expand things out in such a way that it's much easier to add one-off tests of wasi-libc in various build configurations and such. The theory is that this is as "simple" as adding a new matrix entry, copied from previous ones, customized with various variables and environment variables to affect the build (e.g. `CFLAGS`).

Closes #587

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,21 +23,26 @@ jobs:
           - name: Build on Linux x86_64
             os: ubuntu-24.04
             clang_version: 16
+            upload: linux-x86_64-clang-16
           - name: Build on Linux aarch64
             os: ubuntu-24.04-arm
             clang_version: 16
+            upload: linux-aarch64-clang-16
           - name: Build on macOS aarch64
             os: macos-15
             clang_version: 15.0.7
             llvm_asset_suffix: arm64-apple-darwin22.0
+            upload: macos-clang-15
           - name: Build on Windows x86_64
             os: windows-2025
             clang_version: 16.0.0
+            upload: windows-clang-16
 
           # Historical versions of LLVM (11.0.0 currently)
           - name: Build with LLVM 11
             os: ubuntu-22.04
             clang_version: 11
+            upload: linux-x86_64-clang-11
             env:
               BUILD_LIBSETJMP: no
 
@@ -52,6 +57,7 @@ jobs:
             os: ubuntu-24.04
             clang_version: 16
             test: true
+            upload: wasm32-wasi
             env:
               TARGET_TRIPLE: wasm32-wasi
               MAKE_TARGETS: "default libc_so"
@@ -59,6 +65,7 @@ jobs:
             os: ubuntu-24.04
             clang_version: 16
             test: true
+            upload: wasm32-wasip1
             env:
               TARGET_TRIPLE: wasm32-wasip1
               MAKE_TARGETS: "default libc_so"
@@ -66,6 +73,7 @@ jobs:
             os: ubuntu-24.04
             clang_version: 16
             test: true
+            upload: wasm32-wasip2
             env:
               TARGET_TRIPLE: wasm32-wasip2
               WASI_SNAPSHOT: p2
@@ -74,6 +82,7 @@ jobs:
             os: ubuntu-24.04
             clang_version: 16
             test: true
+            upload: wasm32-wasi-threads
             env:
               TARGET_TRIPLE: wasm32-wasi-threads
               THREAD_MODEL: posix
@@ -81,6 +90,7 @@ jobs:
             os: ubuntu-24.04
             clang_version: 16
             test: true
+            upload: wasm32-wasip1-threads
             env:
               TARGET_TRIPLE: wasm32-wasip1-threads
               THREAD_MODEL: posix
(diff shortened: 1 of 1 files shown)


alexcrichton:
Oh, I forgot to also mention, but this deletes the `actions/upload-artifact` of the sysroot at the end of the build. That doesn't actually go anywhere else at this time and AFAIK it's not really possible to depend on an artifact of a build as GitHub cleans them out relatively aggressively. That being said I'm just guessing this won't break anyone. If it does it shouldn't be hard to add back in.

alexcrichton:
Timing wise [this is before this PR](https://github.com/WebAssembly/wasi-libc/actions/runs/15407431129) and [this is this PR](https://github.com/WebAssembly/wasi-libc/actions/runs/15744558521/job/44377897852?pr=588), summarized as:

* 10.5 minutes total wall clock before, 2m after
* Most jobs before were 3-4 minutes with the LLVM 10 build being an outlier at 1m15s. Most jobs after are less than 1 minute with Windows being the outlier at 1.5m

sbc100:
> Oh, I forgot to also mention, but this deletes the `actions/upload-artifact` of the sysroot at the end of the build

Why do this is github is going to do it anyway?

Isn't it maybe nice to be able to download the artifacts or a PR for local testing sometimes?

alexcrichton:
Sure seems reasonable. I prefer to keep things out of CI where possible as it's generally not fun to read yml, but it's pretty small in this case and if it benefits someone it's not too much harm. The origins of this upload date back to 9580a259270d0987905d1322b00bd397cf2ce1bc and the original migration from Azure Pipelines to GitHub Actions FWIW.

alexcrichton:
No that's not intentional.  If you can figure out why that works on CI but doesn't work for you that'd be appreciated!

yamt:
> No that's not intentional. If you can figure out why that works on CI but doesn't work for you that'd be appreciated!

because the ci supplies compiler-rt in a bit unusual way?
https://github.com/WebAssembly/wasi-libc/blob/553305fcc23b879ecd736e5deec874d3f108214c/.github/workflows/main.yml#L158-L160

PR #589, opened 2025-06-19
Skip `pthread_cond_busywait` test in browser tests

kateinoigakukun:
The `pthread_cond` test uses `nanosleep` which eventually calls `poll_oneoff` but `@bjorn3/browser_wasi_shim` does not implement it. So we skip the test for now, but we will revisit it later once https://github.com/bjorn3/browser_wasi_shim/pull/88 is merged.


bjorn3:
https://github.com/bjorn3/browser_wasi_shim/pull/88 has been merged.

Issue #580, opened 2025-04-18
Using SIMD for `string.h`


ncruces:
I make a no-CGO Go SQLite driver, by compiling the amalgamation to Wasm, then loading the result with [wazero](https://github.com/tetratelabs/wazero).

To compile SQLite, I use [wasi-sdk](https://github.com/WebAssembly/wasi-sdk) which uses wasi-libc, based on musl. I'd heard that musl is slow(er than glibc), which is true, to a point.

musl uses SWAR on a `size_t` to implement various functions in `string.h`. This is fine, except `size_t` is just 32-bit on Wasm, whereas most CPUs these days are 64-bit.

I found that implementing a few of those functions with Wasm SIMD128 can make them go around 4x faster.

Other functions don't even use SWAR; redoing those can make them 16x faster.

I found that using SIMD intrinsics (rather than SWAR) seemingly makes it _easier_ to avoid UB, but the code would definitely benefit from more eyeballs. The code (MIT licensed) is at: https://github.com/ncruces/go-sqlite3/blob/main/sqlite3/libc/string.h

I also reimplemented `qsort` but that's another issue.

See [this](https://github.com/ncruces/go-sqlite3/actions/runs/14516931864) for some benchmarks on both x86-64 and Aarch64 (wazero only).

Reviewing this, testing it across other runtimes and platforms, figuring out where it fits and your build infra are all a significant time investment, so opening an issue just to see if there's even interest in something like this (and hopefully find if someone would be willing to pick up some of the work I'm less familiar with).

sunfishcode:
This is cool, and I expect it's of interest to many wasi-libc users.

At a quick glance at the code, I have a few initial notes:

Since these routines do increase code size, would it make sense to enclose the SIMD code in `#ifndef __OPTIMIZE_SIZE__`?

For `memcpy`, `memset``, and `memmove`, does implementing them using the `__builtin_*` form help? I expect clang should do that optimization automatically.

This approach of defining these functions in a header file has the convenient property that they won't require us to publish separate libc.a builds for simd128 and no-simd128, which seems like a good way to get started here. That said, it also does mean that when libc itself calls `memcpy`, it won't get the optimized form, because we don't currently build it with simd128 enabled. So we may eventually need to look at distributing separate libc.a builds for simd128 and no-simd128, as there are many engines out there that don't support simd128.

ncruces:
[...]

PR #592, opened 2025-07-01
Optional SIMD memchr

ncruces:
Continuing #580, followup to #586.

Chose `memchr` because it's somewhat similar to `strlen`, but also because it is the basis for `strnlen` (and in that capacity, for `strndup` and `strlcat`) and is also used by `strstr`, `fnmatch`.

### DIFF
--- a/libc-top-half/musl/src/string/memchr.c
+++ b/libc-top-half/musl/src/string/memchr.c
@@ -2,6 +2,10 @@
 #include <stdint.h>
 #include <limits.h>
 
+#ifdef __wasm_simd128__
+#include <wasm_simd128.h>
+#endif
+
 #define SS (sizeof(size_t))
 #define ALIGN (sizeof(size_t)-1)
 #define ONES ((size_t)-1/UCHAR_MAX)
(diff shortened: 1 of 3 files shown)


libc-top-half/musl/src/string/memchr.c
```diff
@@ -10,6 +14,52 @@
...
+		return NULL;
+	}
+
+	// memchr must behave as if it reads characters sequentially
+	// and stops as soon as a match is found.
+	// Aligning ensures loads beyond the first match are safe.
+	// Casting through uintptr_t makes this implementation-defined,
+	// rather than undefined behavior.
+	uintptr_t align = (uintptr_t)src % sizeof(v128_t);
+	const v128_t *v = (v128_t *)((uintptr_t)src - align);
```
abrown, line 31: @alexcrichton, this would be another spot to use inline assembly, if possible.

ncruces:
Updated with the inline assembly, and suggestions in https://github.com/WebAssembly/wasi-libc/pull/593#issuecomment-3029975380, @alexcrichton please validate.

Also tweaked naming and formatting in `strlen`, for consistency.

abrown (approved):

PR #597, opened 2025-07-17
Optional SIMD str(c)spn

ncruces:
Continuing #580, implements `strspn` and `strcspn`.

This one follows the same general structure as #586, #592 and #594, but uses a somewhat more complicated algorithm, described [here](http://0x80.pl/notesen/2018-10-18-simd-byte-lookup.html).

I used the Geoff Langdale alternative implementation (the tweet as since disappeared) which is correctly described there but has a subtle bug in the implementation: https://github.com/WojciechMula/simd-byte-lookup/issues/2

Since the complexity needed for `__wasm_v128_bitmap256_t` is shared for both `strspn` and `strcspn`, I moved the implementation to a common file, when SIMD is used.

The tests follow a similar structure as the previous ones, and cover the bug, which I was found through fuzzing.

### DIFF
--- a/libc-top-half/musl/src/string/strspn_simd.c
+++ b/libc-top-half/musl/src/string/strspn_simd.c
@@ -14,34 +14,37 @@
 // SIMDized check which bytes are in a set (Geoff Langdale)
 // http://0x80.pl/notesen/2018-10-18-simd-byte-lookup.html
 
+// This is the same algorithm as truffle from Hyperscan:
+// https://github.com/intel/hyperscan/blob/v5.4.2/src/nfa/truffle.c#L64-L81
+// https://github.com/intel/hyperscan/blob/v5.4.2/src/nfa/trufflecompile.cpp
+
 typedef struct {
   __u8x16 lo;
   __u8x16 hi;
 } __wasm_v128_bitmap256_t;
 
 __attribute__((always_inline))
-static void __wasm_v128_setbit(__wasm_v128_bitmap256_t *bitmap, int i) {
-  uint8_t hi_nibble = (uint8_t)i >> 4;
-  uint8_t lo_nibble = (uint8_t)i & 0xf;
-  bitmap->lo[lo_nibble] |= (uint8_t)((uint32_t)1 << (hi_nibble - 0));
-  bitmap->hi[lo_nibble] |= (uint8_t)((uint32_t)1 << (hi_nibble - 8));
+static void __wasm_v128_setbit(__wasm_v128_bitmap256_t *bitmap, uint8_t i) {
+  uint8_t hi_nibble = i >> 4;
+  uint8_t lo_nibble = i & 0xf;
+  bitmap->lo[lo_nibble] |= (uint8_t)(1u << (hi_nibble - 0));
+  bitmap->hi[lo_nibble] |= (uint8_t)(1u << (hi_nibble - 8));
 }
 
 __attribute__((always_inline))
 static v128_t __wasm_v128_chkbits(__wasm_v128_bitmap256_t bitmap, v128_t v) {
   v128_t hi_nibbles = wasm_u8x16_shr(v, 4);
-  v128_t bitmask_lookup = wasm_u8x16_const(1, 2, 4, 8, 16, 32, 64, 128,  //
-                                           1, 2, 4, 8, 16, 32, 64, 128);
+  v128_t bitmask_lookup = wasm_u64x2_const_splat(0x8040201008040201);
   v128_t bitmask = wasm_i8x16_relaxed_swizzle(bitmask_lookup, hi_nibbles);
 
   v128_t indices_0_7 = v & wasm_u8x16_const_splat(0x8f);
   v128_t indices_8_15 = indices_0_7 ^ wasm_u8x16_const_splat(0x80);
 
-  v128_t row_0_7 = wasm_i8x16_swizzle(bitmap.lo, indices_0_7);
-  v128_t row_8_15 = wasm_i8x16_swizzle(bitmap.hi, indices_8_15);
+  v128_t row_0_7 = wasm_i8x16_swizzle((v128_t)bitmap.lo, indices_0_7);
+  v128_t row_8_15 = wasm_i8x16_swizzle((v128_t)bitmap.hi, indices_8_15);
 
   v128_t bitsets = row_0_7 | row_8_15;
-  return wasm_i8x16_eq(bitsets & bitmask, bitmask);
+  return bitsets & bitmask;
 }
 
 size_t strspn(const char *s, const char *c)
@@ -90,7 +93,7 @@ size_t strspn(const char *s, const char *c)
 
   for (; *c; c++) {
     // Terminator IS NOT on the bitmap.
-    __wasm_v128_setbit(&bitmap, *c);
+    __wasm_v128_setbit(&bitmap, (uint8_t)*c);
   }
 
   for (;;) {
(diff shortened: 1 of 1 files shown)


libc-top-half/musl/src/string/strspn_simd.c
```diff
@@ -0,0 +1,175 @@
...
+  uint8_t lo_nibble = (uint8_t)i & 0xf;
+  bitmap->lo[lo_nibble] |= (uint8_t)((uint32_t)1 << (hi_nibble - 0));
+  bitmap->hi[lo_nibble] |= (uint8_t)((uint32_t)1 << (hi_nibble - 8));
+}
+
+__attribute__((always_inline))
+static v128_t __wasm_v128_chkbits(__wasm_v128_bitmap256_t bitmap, v128_t v) {
+  v128_t hi_nibbles = wasm_u8x16_shr(v, 4);
+  v128_t bitmask_lookup = wasm_u8x16_const(1, 2, 4, 8, 16, 32, 64, 128,  //
+                                           1, 2, 4, 8, 16, 32, 64, 128);
```
abrown, line 34: These are both `-128` in the [original algorithm], no?

[original algorithm]: https://web.archive.org/web/20250409223546/http://0x80.pl/notesen/2018-10-18-simd-byte-lookup.html

libc-top-half/musl/src/string/strspn_simd.c
```diff
@@ -0,0 +1,175 @@
...
+      align = 0;
+      addr += sizeof(v128_t);
+    }
+  }
+
+  __wasm_v128_bitmap256_t bitmap = {};
+
+  for (; *c; c++) {
+    // Terminator IS NOT on the bitmap.
+    __wasm_v128_setbit(&bitmap, *c);
```
abrown, line 93: Just a note for future reference: I was initially a bit concerned here that we will incur startup costs too heavy for the "check a small string" use case (?). But of course it's better to loop over `c` once up front rather than at each character in `s` like the scalar version does.

libc-top-half/musl/src/string/strspn_simd.c
```diff
@@ -0,0 +1,175 @@
...
+      // At least one bit will be set, unless align cleared them.
+      // Knowing this helps the compiler if it unrolls the loop.
+      __builtin_assume(mask || align);
+      // If the mask became zero because of align,
+      // it's as if we didn't find anything.
+      if (mask) {
+        // Find the offset of the first one bit (little-endian).
+        return addr - (uintptr_t)s + __builtin_ctz(mask);
+      }
+    }
```
abrown, line 120: We keep on using this pattern; is the logic repetitive enough that we should define a helper function somewhere?

libc-top-half/musl/src/string/strspn_simd.c
```diff
@@ -0,0 +1,175 @@
...
+  __u8x16 lo;
+  __u8x16 hi;
+} __wasm_v128_bitmap256_t;
+
+__attribute__((always_inline))
+static void __wasm_v128_setbit(__wasm_v128_bitmap256_t *bitmap, int i) {
+  uint8_t hi_nibble = (uint8_t)i >> 4;
+  uint8_t lo_nibble = (uint8_t)i & 0xf;
+  bitmap->lo[lo_nibble] |= (uint8_t)((uint32_t)1 << (hi_nibble - 0));
+  bitmap->hi[lo_nibble] |= (uint8_t)((uint32_t)1 << (hi_nibble - 8));
```
abrown, line 27: I'm interested in understanding the codegen of this: so the `...[lo_nibble] |=` is generating some `i8x16.replace_lane` but somehow also `OR`-ing the high nibble bits? What is emitted by LLVM here?

libc-top-half/musl/src/string/strspn_simd.c
```diff
@@ -0,0 +1,175 @@
...
+// SIMDized check which bytes are in a set (Geoff Langdale)
+// http://0x80.pl/notesen/2018-10-18-simd-byte-lookup.html
+
+typedef struct {
+  __u8x16 lo;
+  __u8x16 hi;
+} __wasm_v128_bitmap256_t;
+
+__attribute__((always_inline))
+static void __wasm_v128_setbit(__wasm_v128_bitmap256_t *bitmap, int i) {
```
abrown, line 23: Why not the following?

```suggestion
static void __wasm_v128_setbit(__wasm_v128_bitmap256_t *bitmap, uint8_t i) {
```

libc-top-half/musl/src/string/strspn_simd.c
ncruces: Yes. In two's complement, `int8(-128)` is equal to `uint8(128)`. In interpreting the algorithm, I found unsigned clearer, so used that.

I guess Intel lacks the unsigned version of `_mm_setr_epi8` intrinsic (which would make to the same instruction anyways…)?

What do you think is best, a comment explaining this, or `wasm_i8x16_const`?

ncruces:
Just to document the decision of using Geoff Langdale's variant of the algorithm:

It trades (in the original by Wojciech Muła):
- 1 `wasm_v128_bitselect` _or_ `wasm_i8x16_relaxed_laneselect`
- 2 `wasm_i8x16_swizzle` _or_ `wasm_i8x16_relaxed_swizzle`
- 1 `wasm_i8x16_lt`
- 1 `wasm_v128_and`

By (in the variant with Geoff Langdale input):
- 2 `wasm_i8x16_swizzle`
- 1 `wasm_v128_xor`
- 1 `wasm_v128_or`

libc-top-half/musl/src/string/strspn_simd.c
ncruces: The scalar version does the same: it iterates over `c` once, building a bitmap:
https://github.com/WebAssembly/wasi-libc/blob/main/libc-top-half/musl/src/string/strspn.c

They use some "inscrutable" (but well known) macros to build a more straightforward bitmap in stack memory.
I used this function to build our weird 256-bit bitmap "directly" into a pair of `v128` vectors.

libc-top-half/musl/src/string/strspn_simd.c
ncruces: LLVM cheats and uses the stack, `wasm-opt` (and `wasm-ctor-eval`) then removes any traces of `$__stack_pointer` in this build:
https://github.com/ncruces/go-sqlite3/blob/b72fd5db/sqlite3/libc/libc.wat#L1646-L1726

libc-top-half/musl/src/string/strspn_simd.c
ncruces: Sure. It moves the cast out to the loop, best not rely on implicit `char` to `uint8_t`.

Arguably, this bit would be more correct with `unsigned char` and `unsigned`.

libc-top-half/musl/src/string/strspn_simd.c
ncruces: Yeah, but it's always subtly different.

Some times it's (1) `cmp`, `any_true` and `bitmask`; others (2) `cmp`, `all_true` and `(uint16_t)~bitmask`.

Then, `strlen` gets away with doing `!all_true` before the `cmp` and `bitmask`; `memrchr` does `clz` instead of `ctz`.

I also don't see how to get those two main patterns to inline without going into macros, and inlining is important. That `__builtin_assume` makes the compiler figure out it's good to unroll the first iteration to make the subsequent ones significantly simpler.

ncruces:
Just for documentation purposes (I can add this to the code too), this algorithm is basically the same as `truffle` in Hyperscan:
https://github.com/intel/hyperscan/blob/v5.4.2/src/nfa/truffle.c#L65
https://github.com/intel/hyperscan/blob/v5.4.2/src/nfa/trufflecompile.cpp

The bitset is represented in the exact same way, `truffle` trades an `and` for an `andnot`.

ncruces:
Having tested, it seems the using the opposite predicate (complement) is best for `strspn`, but not `strcspn`, at least on one Intel chip. Not sure if it's worth the additional complexity/duplication.

libc-top-half/musl/src/string/strspn_simd.c
ncruces: Done.

abrown:
Ok, I took another look at this today and ran some fuzzing on it. I found an issue with `strspn`&mdash;not OOB-related this time!&mdash;where the scalar and SIMD versions both successfully return, but return different results:

```
args: [Text(Buffer { data: [97], offset: 130851 }), Text(Buffer { data: [97], offset: 67071 })]
program 0: /.../tests/strspn-scalar.wasm
result 0: Ok(0)
program 1: /.../tests/strspn-simd.wasm
result 1: Ok(1)
Error: executions produced different results
```

Hopefully the above output is somewhat decipherable, but to explain: for `size_t strspn(const char *s, const char *accept)` this fuzz case creates `s` at offset 130851 containing the character `a` and does the same for `accept` at offset 67071. The rest of memory is cleared with 0s. When we run each version (`strspn-scalar.wasm` and `strspn-simd.wasm`) a different `size_t` is returned. Any thoughts?

ncruces:
I… again, don't know what to say, because 1 is the correct answer. A string with "a" has a span of 1 "a" characters. Since I'm sure musl can't be wrong about this… I don't know.

PR #598, opened 2025-07-18
Optional SIMD memrchr

ncruces:
Continuing #580. `memrchr` is mostly used to implement `strrchr`, so that's what I test.

It has the advantage that we know the length and can access the entire buffer, so no undefined behavior for this one. I does do unaligned reads, however, so `wasm_v128_load` is used to dereference the pointer. It uses SIMD while there are 16 or more bytes to read, then fallback to scalar.

The only other notable feature is using `clz` rather than `ctz`.

### DIFF
--- a/libc-top-half/musl/src/string/memrchr.c
+++ b/libc-top-half/musl/src/string/memrchr.c
@@ -1,7 +1,29 @@
 #include <string.h>
 
+#ifdef __wasm_simd128__
+#include <wasm_simd128.h>
+#endif
+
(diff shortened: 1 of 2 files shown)


abrown (approved):
Thanks! I've been fuzzing this, comparing output of the scalar and SIMD versions, and things seem to check out. I'll also note that just enabling `__wasm_simd128__` (without this logic) seemed to do some amount of vectorization &mdash; something to compare against later.

PR #603, opened 2025-07-28
Optional SIMD memcmp

ncruces:
This the last PR I'm putting forward for #580.

`memcmp` has the advantage that we know the lengths and can access the entire buffers (no undefined behavior). It has the difficulty that the buffers may not share an alingment, so it uses `wasm_v128_load`.

It uses SIMD if there are 16 or more bytes to read, otherwise it fallbacks to scalar. If the number of bytes is larger than 16, but not a multiple of 16, the second iteration retests some already tested bytes, to "align" the remaining length to a multiple of 16. Making the first (rather than the last) iteration special unnecessarily "wastes" these comparisons, but helps the compiler partially unroll the loop.

### DIFF
--- a/libc-top-half/musl/src/string/memcmp.c
+++ b/libc-top-half/musl/src/string/memcmp.c
@@ -1,7 +1,42 @@
 #include <string.h>
 
+#ifdef __wasm_simd128__
+#include <wasm_simd128.h>
+#endif
+
 int memcmp(const void *vl, const void *vr, size_t n)
 {
+#if defined(__wasm_simd128__) && defined(__wasilibc_simd_string)
+	if (n >= sizeof(v128_t)) {
+		// memcmp is allowed to read up to n bytes from each object.
+		// Find the first different character in the objects.
+		// Unaligned loads handle the case where the objects
+		// have mismatching alignments.
+		const v128_t *v1 = (v128_t *)vl;
+		const v128_t *v2 = (v128_t *)vr;
+		while (n) {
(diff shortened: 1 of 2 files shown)


abrown:
Ok, I'm puzzling over a fuzzing issue with this one. Here is the scenario:
- the memory length of the Wasm program is 131072 bytes
- I write a 16-byte `s1` string (e.g., `aaa...`) any normal place in the module
- I write another 16-byte `s2` string two bytes from the end, at offset 131072
- I run `memcmp(s1, s2, 16)` to trigger the behavior in this PR

If `s2` is `aa`, both the scalar and the SIMD versions of this function fail due to an OOB access. If `s2` is `ab`, then the scalar version does not fail, but returns as expected (`-1`); but the SIMD version fails.

Now, this is a quite fundamental question for this kind of PR: should we match the scalar behavior (i.e., what users might expect)? Or should we punt because the inputs were malformed (i.e., indexed past the end of memory)? @sunfishcode, @alexcrichton, @sbc100: any thoughts on this?

abrown:
> If so, I would consider this UB, so anything goes, meaning that either faulting or returning error or return success are all valid.

I guess I was getting to this, cautiously. There are several functions in this SIMD vein that will have a similar caveat: for any future readers, note that using the SIMD versions of these functions in wasi-libc could have different behavior than their scalar version when using data close to the end of memory. But this is not new to WebAssembly, since SIMD versions compiled to native machine code would also deal with protection boundaries that causing similar differences in behavior.

Other than that, my fuzzing shows no other differences and I propose we merge this.

abrown (approved):

PR #611, opened 2025-08-11
Port _Exit() to use wasip2 method

catamorphism:
WARNING: This commit edits the auto-generated files wasip2.h and wasip2.c to add the _Noreturn attribute to `exit_exit`.

### DIFF
--- a/libc-bottom-half/headers/public/wasi/wasip2.h
+++ b/libc-bottom-half/headers/public/wasi/wasip2.h
@@ -1084,7 +1084,7 @@ extern bool environment_initial_cwd(wasip2_string_t *ret);
 
 // Imported Functions from `wasi:cli/exit@0.2.0`
 // Exit the current instance and any linked instances.
-extern void exit_exit(exit_result_void_void_t *status);
+_Noreturn extern void exit_exit(exit_result_void_void_t *status);
 
 // Imported Functions from `wasi:io/error@0.2.0`
 // Returns a string that is suitable to assist humans in debugging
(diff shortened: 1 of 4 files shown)


catamorphism:
I'm not sure if wit-bindgen has a way to mark a function as "not returning" so it would generate the `_Noreturn` attribute automatically? I couldn't find one.

libc-bottom-half/headers/public/wasi/wasip2.h
```diff
@@ -1084,7 +1084,7 @@ extern bool environment_initial_cwd(wasip2_string_t *ret);
 
 // Imported Functions from `wasi:cli/exit@0.2.0`
 // Exit the current instance and any linked instances.
-extern void exit_exit(exit_result_void_void_t *status);
+_Noreturn extern void exit_exit(exit_result_void_void_t *status);
```
pchickey, line 1087: Since editing these files is unexpected, but I agree necessary in this case, please add to the comment here noting this signature has been edited from the generated code. I'm not sure if anyone will ever read the comment, but it might make it stick out more in a git diff if the files get regenerated in the future.

pchickey (approved):
wit-bindgen doesn't have the ability to annotate with _Noreturn because exit is the one and only special case which requires it. We had put it in witx and witx-bindgen to handle the special case, but decided not to carry on with that, since there's no way to nontrivially virtualize a call to exit. I'm resigned to exit always being a strange, special case.

In general the wasm vm still doesn't provide a reasonable way to implement exit besides host magic - I had thought thered be a way to implement it with the exceptions proposal, but since exit forbids running destructors, we'd have to come up with some way to avoid those if we were to implement exit by unwinding to the return site of main.

PR #612, opened 2025-08-11
Port random numbers to use wasip2 methods

catamorphism:


catamorphism:
I'm debugging an issue where the new test I added fails with a release build (`-O2 -NDEBUG`) but passes with a debug build (`-O0 -g`).

catamorphism:
Bug should be fixed now.

PR #613, opened 2025-08-11
Remove uses of preview1 component adapter

catamorphism:
Includes removing remaining uses of wasip1 functions with wasip2 defined


catamorphism:
> It looks like there's still a CI failure due to a problem with random

Fixed (this was because of https://github.com/WebAssembly/wasi-libc/pull/620 -- I guess I hadn't rebased locally.)

pchickey (approved):

PR #620, opened 2025-08-27
Replace wasi-libc's guest PRNG with a call to `__wasi_random_get`.

sunfishcode:
Always call `__wasi_random_get` rather than doing a guest PRNG, because Wasm engines may snapshot or even clone Wasm state, which we don't have any visibility into.

We therefore effectively expect that `__wasi_random_get` is "fast", presumably using a PRNG-style implementation rather than a slower raw-entropy-style implementation.

### DIFF
--- a/libc-top-half/sources/arc4random.c
+++ b/libc-top-half/sources/arc4random.c
@@ -4,130 +4,20 @@
 #include <unistd.h>
 #include <stdlib.h>
 
-#define RNG_RESERVE_LEN 512
-
-#define CHACHA20_KEYBYTES 32
-#define CHACHA20_BLOCKBYTES 64
-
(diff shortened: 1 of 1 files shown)


pchickey (approved):

yamt:
> We therefore effectively expect that `__wasi_random_get` is "fast", presumably using a PRNG-style implementation rather than a slower raw-entropy-style implementation.

have you made a research if existing implementations are "fast" as expected here?

sunfishcode:
I have not made a comprehensive survey, but it's common for implementations to use sources such as /dev/urandom or `getrandom` which use PRNGs rather than raw entropy.

yamt:
this change has been made with an assumption that `__wasi_random_get` is fast enough.
now we received a feedback saying it was not really true.
i guess we should consider a revert.

PR #630, opened 2025-10-01
poll_wasip1: ignore entries with no events requested

vapier:
OpenSSH allocates a pollfd structure for all of its channels, but doesn't request events for all of them all the time.  If it wants to check a subset, it clears events, but keeps nfds at the max. This causes the wasip1 implementation to incorrectly abort as it assumes that if POLLRDNORM & POLLWRNORM didn't match, it must be due to requesting unsupported events, and then it errors out with ENOSYS.  Simply ignore the case where no events are requested and it all works fine.


vapier:
thx bud

PR #637, opened 2025-10-22
Add tests for non-blocking sockets

catamorphism:
Also fix bug that was causing stdout/stderr/stdin to be initialized repeatedly

Also fixed getsockopt() to set the error to EINPROGRESS if called on a socket in the Connecting state

### DIFF
--- a/test/src/misc/poll-nonblocking-socket.c
+++ b/test/src/misc/poll-nonblocking-socket.c
@@ -1,0 +1,60 @@
+//! filter.py(TARGET_TRIPLE): wasm32-wasip2
+//! add-flags.py(RUN): --wasi=inherit-network=y
+#include <errno.h>
+#include <fcntl.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <netdb.h>
+#include <sys/socket.h>
+#include <netinet/in.h>
+#include <arpa/inet.h>
+#include <poll.h>
+#include "test.h"
+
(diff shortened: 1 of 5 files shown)


test/src/misc/poll-nonblocking-socket.c
```diff
@@ -0,0 +1,77 @@
...
+    server_address.sin_addr.s_addr = htonl(INADDR_ANY);
+    server_address.sin_port = htons(server_port);
+    TEST(bind(server_socket_fd, (struct sockaddr*)&server_address, sizeof(server_address)) != -1);
+    // Listen on socket
+    char buffer[BUFSIZE];
+    socklen_t client_len = sizeof(struct sockaddr_in);
+    int client_socket_fd = -1;
+    struct sockaddr_in client_address;
+    int32_t bytes_read = 0, total_bytes_read = 0;
+//    TEST(listen(server_socket_fd, 1) != -1);
```
pchickey: Commented-out `listen` here - should be uncommented
```suggestion
    TEST(listen(server_socket_fd, 1) != -1);
```

pchickey (approved):

test/src/misc/poll-nonblocking-socket.c
catamorphism: I left this test in a confusing state -- it was supposed to test something else originally, but now it just tests what happens when a connect request fails. So the code for the server doesn't need to be there at all. I fixed that and force-pushed.

PR #641, opened 2025-11-07
Generate bindings for `wasi:cli/imports`, not `wasi:cli/command`

### LINKED ISSUE #640: `libc.so` exports `wasi:cli/run@0.2.0#run`, but it shouldn't
Starting with WASI-SDK 28, the `libc.so` build for `wasm32-wasip2` exports `wasi:cli/run@0.2.0#run`, which means an application component that includes `libc.so` cannot also export it.  In addition, it means every component that includes `libc.so` now exports that function whether it intended to or not.

I don't believe `libc.so` should export any component-level functions whatsoever -- only import them.  It's up to the application to export any component-level functions.

alexcrichton:
Closes #640

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1091,7 +1091,7 @@ bindings: $(BINDING_WORK_DIR)/wasi-cli $(BINDING_WORK_DIR)/wit-bindgen
 			< wasip2.c \
 			> ../../libc-bottom-half/sources/wasip2.c && \
 		rm wasip2.c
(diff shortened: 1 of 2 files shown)


pchickey:
Just to be clear: this change works because, by eliminating `LDFLAGS += -Wl,--skip-wit-component` in the build of components against this libc, wasi-sdk's linker will insert a `wasi:cli/run@0.2.0#run` export pointing to the libc `_start` symbol.

PR #645, opened 2025-11-08
wasip2: Don't store pollables in descriptor table

alexcrichton:
This reduces the amount of state that needs to be managed in the table and also mirrors the adapter where it creates pollables on-the-fly. In general pollables are intended to be cheap to create so it should be ok to avoid caching them and carefully managing where they're located.

This commit additionally refactors the implementation of `poll` to be split into separate functions rather than all one large function.


alexcrichton:
That's reasonable yeah, I'll split out the refactorings from this and land those separately.

PR #646, opened 2025-11-10
Rewrite testing with CMake's CTest

### LINKED ISSUE #604: build: Refactor how components are created/tested
On #599, @alexcrichton [suggested](https://github.com/WebAssembly/wasi-libc/pull/599#issuecomment-3120168350):  

*    Refactor how components are created/tested
        Previously wasm-tools component new was executed, but that predates a lot of bits and pieces today so I'd drop that entirely as well as the whole *.core.wasm and *.component.wasm distinction, nowadays we should only emit *.wasm.
        For wasip2 testing clang will complain that it can't execute wasm-component-ld (which internally does basically what wasm-tools component new does). For that you can arrange for wasm-component-ld to be installed on the builder (e.g. cargo install wasm-component-ld@VERSION, or something similar)

For the component-ld bits, if you're willing, my goal is to get the testing pretty close to what the toolchain does "for real" in the wasi-sdk artifacts which is to use wasm-component-ld to produce a component natively. If you'd prefer though I can also work on said refactoring if it's not clear how to apply it.

---

We agreed to land the PR without this change; filing an issue so that I or someone else can do it in the future.

### LINKED ISSUE #627: Tests can't write to standard output streams
#625 shows that it would be nice to have tests that write to stdout / stderr. Currently any test that writes to stdout / stderr is considered a failure. It would be good to extend the test harness to allow for such tests. The same goes for reading from stdin, although that would be a little more complicated.

alexcrichton:
This commit rewrites the tests in this repository to using CMake's testing harness CTest. This is borne out of current and historical frustrations I've had with testing in this repository with the goal of laying a solid foundation for future growth with testing. Some specific rationale for this change is:

* Current tests cannot be run in parallel. This means that tests also cannot be built in parallel since it's bundled together right now. The reason for this is due to the introduction of sockets-related tests that all use the same port. CTest has the ability to specify that a set of tests all lock an abstract "resource" which means that they run serially, solving this issue.

* Debugging test failures is difficult right now where a test can be reported as failed but a file must be consulted to see the output of the test. CTest has robust facilities and flags for capturing test output and reporting it, solving this issue.

* Running one test right now is not easy to do. A test can be run but whether it passed/failed needs to be consulted via an output file. The `ctest` command line has baked-in abilities to run one or a subset of tests.

* Adding a timeout to tests (often helpful for networking-related tests) is not currently possible. While enough shell-script-magic might make this work it's much easier to use CTest's built-in timeout mechanism.

* Current shell-script-based-testing has no means of testing what the output of a test is, but CTest has this built in. This means that there's now a test that prints hello world and will fail if nothing is printed.

* The makefile-based tests currently print no indication of what test is running. With sockets-tests often taking 1s+ each it means that a test run can seemingly hang while it's actually running tests. CTest has lots of built in options and a reasonable default for displaying test progress.

* Per-test configuration is done currently in the test file itself which is parsed out via Python scripts. While this works well enough it's project-specific and can be difficult to remember syntax. CMake and CTest combined have many options of how to configure tests in the same manner. A downside of CMake/CTest, however, is that test configuration is separate from the test itself (e.g. in `CMakeLists.txt` as opposed to the test).

* In broad strokes I expect CMake/CTest to be more robust about avoiding using stale artifacts or reporting on stale results. I got myself into weird situations with the makefile-based tests where the only way I could trust the output was to blow away the entire test build directory between each run. CTest/CMake are more often than not "right by default" in terms of build dependencies unlike makefiles where everything needs to be built from scratch.

* Personally I find shell scripts very hard to parse and read. The current `run-test.sh` is a large wad of shell and is difficult to debug and/or run tests in isolation. With CTest all handling of processes and such is handled by CTest itself and a verbose command line option (`-V`) can print the exact command per-test which means that Wasmtime, for example, can be directly executed for tests.

Overall I personally feel that CTest is a much more solid foundation on which to build tests in wasi-libc. It provides many more features by-default than building a custom test harness. While it's possible to build makefiles that handle all of the above cases better I don't think we have the development resources in wasi-libc to maintain such a test harness and suite. Instead I feel it's better to hitch the wagon to an externally-maintained solution with CTest and CMake.

This commit does not transition wasi-libc's main build system to CMake and instead leaves the top-level `Makefile` untouched. That's left for another day. Some other minor changes in this are:

* The `libc-test` dependency is now pinned to a commit
* Component tests use `wasm-component-ld` instead of `wasm-tools` to build components.
* Tests from `libc-test` no longer have a file in this repo with a `#include` of the source but are instead directly configured in CMake configuration.
* Two sockets-related tests are disabled due to spurious failures found while testing locally.
* The `test/Makefile`, `test/scripts/*.py`, and `run-test.sh` files are all deleted.
* With `test/src/libc-test` now being gone all tests are moved from `test/src/misc` to `test/src`.
* Some minor test adjustments were made which were found to be spurious failures while developing this.
* Tests for `wasm32-wasi-threads` were deleted since that was actually testing `wasm32-wasip1-threads` and didn't provide any benefit over that test builder.

cc #322
Closes #604
Closes #627

### DIFF
--- a/test/src/clock_nanosleep.c
+++ b/test/src/clock_nanosleep.c
@@ -19,11 +19,10 @@ int main(void)
     // Not much point running the test if the resolution is >= 1 sec
     TEST(clock_resolution.tv_sec==0);
 
-    // Adjust this number if necessary
-    // In wasip2 we don't get accurate results for sleeping < 1 ms
 
-    struct timespec time_to_sleep = { .tv_sec = 0,
-                                      .tv_nsec = 1E6 * clock_resolution.tv_nsec };
+    // Sleep for a total of 5ms
+    long ns_to_sleep = 5E6;
+    struct timespec time_to_sleep = { .tv_sec = 0, .tv_nsec = ns_to_sleep };
 
     struct timespec start_time, end_time;
     clock_gettime(CLOCK_MONOTONIC, &start_time);
@@ -32,14 +31,14 @@ int main(void)
     clock_gettime(CLOCK_MONOTONIC, &end_time);
     TEST(end_time.tv_sec - start_time.tv_sec <= 1);
 
-    long nanoseconds_elapsed = (end_time.tv_sec - start_time.tv_sec)
+    long nanoseconds_elapsed = (end_time.tv_sec - start_time.tv_sec) * 1E9
         - start_time.tv_nsec
         + end_time.tv_nsec;
 
     // Test that the difference between the requested amount of sleep
     // and the actual elapsed time is within an acceptable margin
-    double difference = abs(nanoseconds_elapsed - time_to_sleep.tv_nsec)
-                        / time_to_sleep.tv_nsec;
+    double difference = abs(nanoseconds_elapsed - ns_to_sleep)
+                        / ns_to_sleep;
 
     // Allow the actual sleep time to be twice as much as the requested time
     TEST(difference <= 1);


pchickey (approved):
I'm not a CTest expert by any means but I trust that its a better mechanism than the Makefile and shell that was here previously, which I also found pretty difficult to understand.

PR #655, opened 2025-11-10
wasip2: Don't reimplement `preopens.c`

alexcrichton:
This is really tricky path-related code and it's currently buggy on wasip2. Instead of duplicating it like before instead have only a single definition with some small handling of wasip1-vs-wasip2. This fixes `chdir`, for example, and handling of preopens. The fundamental bug from before is that a `wasip2_string_t` was handed out as a `const char*`, but it wasn't null-terminated. The bug is more apparent with all the code paths interleaved here.

This additionally fixes an issue where the previous implementation would continuously open new file descriptors for preopens. Instead now a single file descriptor is registered at preopens-registration-time which is then used throughout the lifetime of the program.

### DIFF
--- a/expected/wasm32-wasip2/defined-symbols.txt
+++ b/expected/wasm32-wasip2/defined-symbols.txt
@@ -339,7 +339,6 @@ __wasilibc_open_nomode
 __wasilibc_populate_preopens
 __wasilibc_pthread_self
 __wasilibc_random
-__wasilibc_register_preopened_fd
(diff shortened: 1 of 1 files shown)


alexcrichton:
Ah I can also note that this deletes `__wasilibc_register_preopened_fd` on WASIp2 targets as it should be less necessary and I think we'll want different APIs for interoperating between WASIp2 constructs and libc constructs.

PR #690, opened 2025-12-10
Fix a bug in `linkat`

alexcrichton:
This commit fixes a longstanding issue in wasi-libc where the `linkat` function, when both parameters are `AT_FDCWD` or absolute paths, would relativize the two input paths in such a way that the previous stomped over the first, meaning that the wrong actual syscall was issued. This fixes the problem in the same manner as other two-path-taking functions which is to use `find_relpath` for one path and `find_relpath_alt` for the other path.


yamt (approved):
lgtm

PR #697, opened 2025-12-22
Format wasip2/wasip1-threads files as well

alexcrichton:
Update CI to format a few more targets, then apply all the formatting.


pchickey (approved):
Formatting itself is only spot-checked

Issue #688, opened 2025-12-03
Provide macros to get the WASI and WASI SDK versions


brettcannon:
[Brought up on Zulip](https://bytecodealliance.zulipchat.com/#narrow/channel/219900-wasi/topic/How.20to.20get.20the.20Wasm.2C.20WASI.2C.20.26.20SDK.20version.20via.20C.20macros/with/561728461)[#wasi > How to get the Wasm, WASI, &amp; SDK version via C macros](https://bytecodealliance.zulipchat.com/#narrow/channel/219900-wasi/topic/How.20to.20get.20the.20Wasm.2C.20WASI.2C.20.26.20SDK.20version.20via.20C.20macros/with/561728461), having a way to get the compile-time values of the WASI version (e.g. `0.2`) as well as the WASI SDK version (e.g. `29`) would be great. I bring this up in the scope of CPython as there currently isn't a way to know these details from within CPython itself; see https://github.com/python/cpython/blob/main/Tools/wasm/README.md#detecting-webassembly-builds and how there is very little details about a WASI build compared to Emscripten.

This will also come into play when I attempt to get a Python Enhancement Proposal for wheels for WASI. There needs to be a way to get the [relevant information that would be embedded in the file name](https://bytecodealliance.zulipchat.com/#narrow/channel/219900-wasi/topic/Platform.20tags.20for.20packages.20targeting.20WASI/with/517642678)[#wasi > Platform tags for packages targeting WASI](https://bytecodealliance.zulipchat.com/#narrow/channel/219900-wasi/topic/Platform.20tags.20for.20packages.20targeting.20WASI/with/517642678), and this is what's missing (`__wasm32__` and `__wasm64__` should be enough for CPU architecture).

brettcannon:
#709 looks great as I can use `__wasi_sdk_version__` and set the string in Python's `sys` module!

But with #703 I'm not seeing an equivalent way of getting the WASI version w/o hard-coding checks for each version via `__wasipN__`. That has a drawback that future versions won't be supported w/o a code change. As well, what will be used when WASI reaches 1.0? 2.0?

brettcannon:
> code changes are warranted.

My only concern is when a Python version reaches EOL and so might not work with newer WASI versions. Then again, that would probably require other code updates, so updating one spot isn't a big thing if it's a compiler error.

I think I just talked myself out of caring and being fine with what we got. 😁

PR #703, opened 2026-01-05
Add `wasi/version.h`

alexcrichton:
As a step towards #688 this adds a new header which ensures that the `__wasipN__` variables are defined. These will eventually be provided via llvm/llvm-project#165345 but in the meantime this uses a header to define them.

This then additionally refactors the internals to have `wasi/wasip{1,2}.h` as well as `wasi/api.h` which includes the correct header. The previous `__wasilibc_use_wasip2` macro is removed entirely in favor of `__wasip2__` defined in this header.

### DIFF
--- a/libc-top-half/musl/src/thread/pthread_create.c
+++ b/libc-top-half/musl/src/thread/pthread_create.c
@@ -14,6 +14,7 @@
 
 #include <stdalign.h>
 #include <assert.h>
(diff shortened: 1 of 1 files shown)


alexcrichton:
Note that this is split out of #700 to see more about the end goal

pchickey (approved):

PR #710, opened 2026-01-07
wasip2: Improve "command" startup flow

alexcrichton:
This commit is an improvement to the WASIp2 target which changes how the interface `wasi:cli/run` is defined and exported. Previously the `wasm32-wasip2` target would define a `_start` wasm export which required the use of the WASIp1-to-WASIp2 adapter to hook it up to `wasi:cli/run`. This meant that despite libc not actually using any WASIp2 APIs the adapter was still used. The purpose of this commit is to change this.

The changes made here are:

* The linker-level `_start` symbol now has a wasm-level export name of `wasi:cli/run@0.2.0#run`. This means that `_start` isn't a wasm-level export any more, meaning there's nothing for the adapter to import.
* The signature of the C-level `_start` function is now different based on WASI version (returns an `int` in WASIp2, nothing on WASIp1)
* The `crt1-command.o` object file now contains type information for `wasi:cli/run` (which previously wasn't present anywhere in wasi-libc). This means that when this object is linked in `wit-component` will know what to do after linking.

The end result is that binaries are now produced entirely without the WASIp1-to-WASIp2 adapter and linking with `-Wl,--wasi-adapter=none`, for example, produces a working binary.

Another minor change made in this commit is that the `crt1-*.c` files are now included in `clang-format` like all other source files.

Build-wise this requires the usage of `wasm-tools` at build-time to embed type information in the clang-produced object file for `crt1-command.o`.

### DIFF
--- a/libc-bottom-half/CMakeLists.txt
+++ b/libc-bottom-half/CMakeLists.txt
@@ -216,6 +216,21 @@ elseif (WASI STREQUAL "p2")
   # wasip2: crt1-command.o is modified from what CMake produces to
   # additionally have a custom section representing the type information needed
   # for its contained export. This is the `wasi:cli/run` interface, for example.
+  #
+  # TODO: ideally this wouldn't need `wasm-tools` as that's an extra build tool
+  # needed here and it's best to slim down dependencies as much as possible.
+  # This additionally requires downloading/vendoring WITs which isn't great.
+  # Ideally the type information embedded here would be referenced dierctly in
+  # `crt1-command.c` itself in the source. That would likely require some
+  # combination of `__asm__` and `#embed` but I at least couldn't figure out how
+  # to get that working. I'm also not aware of a way to, in Clang, define a
+  # global in C that goes into a custom section in the output object. Another
+  # possible route would be to use the `--relocatable` option of `wasm-ld,` but
+  # it looks like that loses `__attribute__((export_name("...")))` information
+  # which this object file relies on. As a workaround `wasm-tools` is used for
+  # now. If you, dear reader, know of how to do this in the C source itself
+  # please feel free to open an issue or a PR and maintainers can work with you
+  # on getting that integrated.
   add_custom_command(
     OUTPUT ${SYSROOT_LIB}/crt1-command.o
     COMMAND


libc-bottom-half/crt/crt1-command.c
```diff
@@ -9,54 +9,61 @@ extern void __wasm_call_ctors(void);
...
-    }
+  if (r != 0) {
+    __wasi_proc_exit(r);
+  }
 #elif defined(__wasip2__)
-    if (r != 0) {
-        exit_result_void_void_t status = { .is_err = true };
-        exit_exit(&status);
-    }
+  return r != 0;
```
TartanLlama: Why is the `exit` call no longer needed?

libc-bottom-half/CMakeLists.txt
```diff
@@ -161,31 +161,66 @@ foreach(obj bottom-half-shared bottom-half-static)
...
+    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_OBJECTS:crt1-command> ${SYSROOT_LIB}/crt1-command.o
+    DEPENDS crt1-command $<TARGET_OBJECTS:crt1-command>
   )
-  add_custom_target(sysroot-startup-${filename}.o DEPENDS ${dst})
-  add_dependencies(sysroot-startup-objects sysroot-startup-${filename}.o)
-endforeach()
+else()
+  # wasip2+: crt1-command.o is modified from what CMake produces to
+  # additionally have a custom section representing the type information needed
+  # for its contained export. This is the `wasi:cli/run` interface, for example.
```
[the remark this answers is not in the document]
sbc100: That looks like something that it should be possible to support with the asm syntax yes.    Assuming that `#embed` plays nice with `__asm__` in general (i.e. on other platforms) i don't see why it shouldn't for us.

BTW, In this case are you actually embedding a wasm file as a custom section in another wasm file (i.e. the object file)?

libc-bottom-half/crt/crt1-command.c
```diff
@@ -9,54 +9,64 @@ extern void __wasm_call_ctors(void);
...
+#if defined(__wasip1__)
+__attribute__((export_name("_start"))) void _start(void)
+#elif defined(__wasip2__)
+// Note that this is manually doing what `wit-bindgen` might otherwise be
+// doing. Given the special nature of this symbol this skip the typical
+// `wit-bindgen` rigamarole and the signature of this function is simple enough
+// that this shouldn't be too problematic (in theory).
+__attribute__((export_name("wasi:cli/run@0.2.0#run"))) int _start(void)
+#elif defined(__wasip3__)
+__attribute__((export_name("wasi:cli/run@0.3.0-rc-2025-09-16#run"))) int
```
cpetig: Isn't `wasi:cli/run.run` an async function in wit, and thus would need a different annotation and a callback companion entry point? I guess this can work because the runtime adapts this blocking implementation?

alexcrichton: Correct yeah, this is intentionally a sync lift into an async function type which the runtime will handle with stack switching and such. While reactor-style programs will be able to work with callbacks I don't think there's any route for us to transform `int main() { ... }` in C/C++ into callback-y style.

libc-bottom-half/CMakeLists.txt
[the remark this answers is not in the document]
sbc100: Perhaps we could leave a comment there suggesting that a simpler solution might be possible with inline asm.

In the long run I do like the idea striving for "its just clang" as much as possible in wasi-sdk, but for the users and for the developers.   I know that was one of the original goals when we first started wasi-sdk.

pchickey (approved):

PR #711, opened 2026-01-11
Wasip3 file, stdout, tcp, poll and pipe implementations

cpetig:
This is a rebase of https://github.com/alexcrichton/wasi-libc/pull/1 after all dependencies landed.

This doesn't contain the run rework, as https://github.com/WebAssembly/wasi-libc/pull/710 will take care of that.

~~Also the expected symbols will be wrong in this initial dump as I separated out several parts.~~

PS: For now I prefer abort() over returning an error on unimplemented functionality as it is much more easy to debug.


libc-bottom-half/headers/private/wasi/descriptor_table.h
```diff
@@ -37,6 +41,11 @@ typedef struct descriptor_vtable_t {
   /// Same as `get_read_stream`, but for output streams.
   int (*get_write_stream)(void*, streams_borrow_output_stream_t*, off_t**, poll_own_pollable_t**);
 #endif
+#ifdef __wasip3__
+  // async interfaces
+  int (*read3)(void*, void *buf, size_t nbyte, waitable_t *waitable, wasip3_waitable_status_t *out, off_t**);
```
cpetig: Previously this just returned the stream_u8_t, but since each subsystem has its own (although compatible (?)) stream_read/write function I fused get_write_stream and write. Works for poll, so :shrug:

libc-bottom-half/sources/wasip3_stdio.c
```diff
@@ -2,10 +2,142 @@
...
+} stdio3_t;
+
+static void stdio3_free(void *data) {
+  stdio3_t *stdio = (stdio3_t *)data;
+  stdin_stream_u8_drop_readable(stdio->input);
+  stdin_future_result_void_error_code_drop_readable(stdio->future);
+  free(stdio);
+}
+
+static int stdio3_get_write_stream(void *data, void const *buf, size_t nbyte,
```
cpetig: still the old name (get_stream), rename to write

cpetig (commented):
There is plenty of code duplication between the wasip2 and wasip3 files, but the functions are different in central regards. We should keep de-duplication in mind after the functionality is more complete.

Ideally progress should be tracked by a test suite, I guess the old p2 test cases should give a good start.

cpetig:
The parts of this PR have now become separate PRs (linked above).

PR #713, opened 2026-01-13
Set CMAKE_C_COMPILER_WORKS globally

yamt:


alexcrichton:
That makes sense to me yeah where the wasm32-wasip3 target doesn't actually exist in any compiler/sysroot/etc. I understand the need for one-off applications of this flag, but I don't feel that it is necessary to unconditionally specify this with no possible override in the in-repo build configuration

PR #715, opened 2026-01-13
WASIP3 threading support

TartanLlama:
Adds `pthread` support for wasip3 based on the component model's cooperative threading builtins.

Currently a draft for progress tracking.


cpetig:
I think your assembly code to set the initial context from the globals got lost when the new run implementation was integrated. At least for me the initial context 0 (stack pointer) is zero and results in traps.

PR #716, opened 2026-01-13
Change `pthread_create` stub's error code

alexcrichton:
This changes from `EAGAIN` to `ENOTSUP` since this operation can't ever succeed, it's entirely unsupported. This is in relation to rust-lang/rust#151016 where in the Rust ecosystem crates were using the "Unsupported" error kind to gracefully handle errors from spawning a thread, such a falling back to single-threaded code. By switching to `EAGAIN` it caused these graceful fallbacks to not get triggered.

### DIFF
--- a/thread-stub/pthread_create.c
+++ b/thread-stub/pthread_create.c
@@ -8,12 +8,7 @@ weak_alias(dummy_0, __release_ptc);
 
 int __pthread_create(pthread_t *restrict res, const pthread_attr_t *restrict attrp, void *(*entry)(void *), void *restrict arg)
 {
-	/*
-		"The system lacked the necessary resources to create another thread,
-		or the system-imposed limit on the total number of threads in a process
-		{PTHREAD_THREADS_MAX} would be exceeded."
-	*/
-	return EAGAIN;
+	return ENOTSUP;
 }
 
 weak_alias(__pthread_create, pthread_create);


alexcrichton:
In my experience man pages list a subset of possible errors from functions, so I wouldn't personally consider man pages as an exhaustive source of truth for all possible errors. In the abstract "not supported" here feels more appropriate to me than "try again later" since it's impossible for this operation to succeed on `wasm32-wasip{1,2}` targets

alexcrichton:
I really don't think we should be using man pages/opengroup/etc as an ironclad set of guidelines for what is possible to return as an error code. For example their [documentation of `sigaction`](https://pubs.opengroup.org/onlinepubs/7908799/xsh/sigaction.html) says that it can only return `EINVAL` but in Linux it [might return `EFAULT`](https://github.com/torvalds/linux/blob/b54345928fa1dbde534e32ecaa138678fd5d2135/kernel/signal.c#L4640). My impression has always been that documented errors are examples, but not exhaustive. Especially once you factor in things like seccomp filters and such where arbitrary syscalls can be denied the documented list of return values is almost always never exhaustive.

sbc100:
If I ask AI "can pthead_create set errno to ENOTSUP and be spec compliant" is says no, but I guess its just using the same docs I'm referring to.

Note that emscripten's stub returns `EAGAIN` and we've not had any complaints or requests to change it so far: https://github.com/emscripten-core/emscripten/blob/d3d9e0a1ba2d57b812e63e0b96b24a850744ddc5/system/lib/pthread/library_pthread_stub.c#L94-L96.

I don't think its a huge deal either way, but couldn't the rust code instead be tought to fall back in the case of EAGAIN for pthread_create?   I guess we would need to change the emscripten version too if rust doesn't want to change?

sbc100 (commented):
OK, LGTM.

Equivalent emscripten change is https://github.com/emscripten-core/emscripten/pull/26105

PR #717, opened 2026-01-14
implement clock_sleep for wasip3

cpetig:


cpetig:
This directly poses some questions:
- should block_on subtask_state instead?
- the test fails with <br> `wasm trap: cannot block a synchronous task before returning`  <br>  Did I make a conceptual mistake? Is this a limitation in wasmtime?

alexcrichton:
I think for that you'll want to rebase on `main` to pick up https://github.com/WebAssembly/wasi-libc/pull/710

libc-bottom-half/cloudlibc/src/libc/time/clock_nanosleep.c
```diff
@@ -57,9 +60,29 @@ int clock_nanosleep(clockid_t clock_id, int flags, const struct timespec *rqtp,
...
-  errno = ENOTSUP;
-  return -1;
+  // Note: rmtp is ignored
+
+  if (clock_id != CLOCK_MONOTONIC) {
+    // wasip3 only provides a pollable for monotonic clocks
+    return ENOTSUP;
+  }
+
+  // Prepare pollable
```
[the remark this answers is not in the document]
alexcrichton: (similar to the pollable comment about monotonic clocks above)

libc-bottom-half/cloudlibc/src/libc/time/clock_nanosleep.c
```diff
@@ -57,9 +60,29 @@ int clock_nanosleep(clockid_t clock_id, int flags, const struct timespec *rqtp,
...
+  int64_t duration = (rqtp->tv_sec * NSEC_PER_SEC) + rqtp->tv_nsec;
+  wasip3_subtask_status_t status;
+  for (;;) {
+    if (flags & TIMER_ABSTIME) {
+      status = monotonic_clock_wait_until(duration);
+    } else {
+      status = monotonic_clock_wait_for(duration);
+    }
+
+    switch (wasip3_subtask_block_on(status)) {
```
alexcrichton: Thoughts on this:

* One possible change to make here is to generate bindings for sync lowerings of these functions rather than async lowerings. That would mean that the C bindings here would be a simple "call the function and return".
* I'd bikeshed the interface of `wasip3_subtask_block_on` a bit, specifically I don't think it should return anything. Blocking on completion of a subtask is, in theory, an infallible operation. I'd imagine that by calling it you're contractually saying "you now own the subtask, go deal with it". In that sense there's no way the subtask can be cancelled and it's required to finish in the component model.

That would then simplify this where the `switch` would't be necessary at all, and I believe the loop could also be removed.

As to whether it's reasonable to generate synchronous bindings I'm not entirely sure. For example I don't know if wasi-libc is going to want both synchronous and asynchronous bindings for some functions. If it only ever wants one we can just configure the bindings generator, but if it wants both then we'd either need to hand-write bindings for one or update wit-bindgen to be able to generate bindings for both. I'd personally tend towards though trying to use synchronous bindings over call-then-block-on unless needed otherwise since that'll reduce code size and give more information to the runtime to optimize and such.

libc-bottom-half/sources/wasip3_subtask.c
alexcrichton: For some other WASI-version-specific files I've been adding `#ifdef` guards which means the file is empty on some versions but populated on others. I then, however, also only include the file in CMake on some versions vs others. I'm fine either way here personally, but if there's no #ifdef guard, mind adding a `#error` if the wasi version is incorrect? (basically try to catch mistakes where this is compiled for the wrong target)

libc-bottom-half/cloudlibc/src/libc/time/clock_nanosleep.c
alexcrichton: Oh, also:

>  A state of starting will require waiting and re-issuing the command, so the block_on function would need to return to reissue the clock_wait_for, right? And cancellation would likely require to bubble up in the caller, wouldn't it?

For "STARTING" there's no need to re-invoke the caller since the task is queued up and waiting, and it's not possible for the callee to cancel itself, just the caller can do that. Given taht there's no need to re-issue anything and since this isn't itself cancelling there's no need to handle cancellation

libc-bottom-half/cloudlibc/src/libc/time/clock_getres.c
```diff
@@ -32,9 +32,19 @@ int clock_getres(clockid_t clock_id, struct timespec *res) {
     }
   }
 #elif defined(__wasip3__)
-  // TODO(wasip3)
-  errno = ENOTSUP;
-  return -1;
+  if (res != NULL) {
```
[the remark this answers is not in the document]
cpetig: I tried, whether the name difference for resolution renders it less readable is a matter of taste.

cpetig:
Could it be reasonable to implement p3 in this less efficient way (call async+waitable-set.wait) and later migrate towards calling sync imports when the path became more clear?

alexcrichton (approved):
Sounds reasonable to me to handle the sync lowering later yeah, thanks!

PR #718, opened 2026-01-14
P3 stdout implementation

cpetig:
This needed a bit more of file infrastructure than I anticipated. The code in wasip3_file_utils.c needs more work to set the proper errno when writes/reads fail, but for stdout/in it should be good for a first step.

### DIFF
--- a/libc-bottom-half/cloudlibc/src/libc/unistd/read.c
+++ b/libc-bottom-half/cloudlibc/src/libc/unistd/read.c
@@ -64,7 +64,7 @@ ssize_t read(int fildes, void *buf, size_t nbyte) {
   wasip3_waitable_status_t status =
       filesystem_stream_u8_read(stream->f0, buf, nbyte);
   size_t amount = wasip3_waitable_block_on(status, stream->f0);
-  if (amount > 0) {
+  if (amount > 0 || nbyte == 0) {
     if (off)
       *off += amount;
     return amount;
--- a/libc-bottom-half/cloudlibc/src/libc/unistd/write.c
+++ b/libc-bottom-half/cloudlibc/src/libc/unistd/write.c
@@ -86,7 +86,7 @@ ssize_t write(int fildes, const void *buf, size_t nbyte) {
     wasip3_waitable_status_t status =
         filesystem_stream_u8_write(write_end->output, buf, nbyte);
     size_t amount = wasip3_waitable_block_on(status, write_end->output);
-    if (amount > 0) {
+    if (amount > 0 || nbyte == 0) {
       if (off)
         *off += amount;
       return amount;


libc-bottom-half/cloudlibc/src/libc/unistd/read.c
```diff
@@ -53,9 +53,7 @@ ssize_t read(int fildes, void *buf, size_t nbyte) {
     *off += contents.len;
   return contents.len;
 #elif defined(__wasip3__)
-  // TODO(wasip3)
-  errno = ENOTSUP;
-  return -1;
+  return __wasilibc_read3(fildes, buf, nbyte);
```
alexcrichton: Currently the general style in wasi-libc is to have the implementation of this function inlined here (since I think this will forever be the only caller, right?)

Some helpers I could see being reasonable to put elsewhere, but for the main body of the implementation that seems like it'll be `read`-specific so I think it's reasonable to inline here (and in `write` below)

Although I'll be honest in terms of style I'm just winging it I don't have a strong rhyme or reason one way or the other. Keeping things as "one symbol per file and as few deps on other files as possible" seems to be the general ethos.

libc-bottom-half/headers/private/wasi/descriptor_table.h
```diff
@@ -37,6 +42,12 @@ typedef struct descriptor_vtable_t {
   /// Same as `get_read_stream`, but for output streams.
   int (*get_write_stream)(void*, streams_borrow_output_stream_t*, off_t**, poll_own_pollable_t**);
 #endif
+#ifdef __wasip3__
+  /// Start an asynchronous read or write, returns zero on success.
+  /// Stores the waitable, status and offset location.
+  int (*read3)(void*, void *buf, size_t nbyte, waitable_t *waitable, wasip3_waitable_status_t *out, off_t**);
+  int (*write3)(void*, void const *buf, size_t nbyte, waitable_t *waitable, wasip3_waitable_status_t *out, off_t**);
```
alexcrichton: Would it be possible to make this look more like `get_{read,write}_stream` above? Basically where these are in theory simple accessors for the underlying stream and that's it. That way the implementation of `read` isn't split up across this callback and the `wasip3_file_utils.c` file (and in the future I think would help deduplicate for, e.g., pipes and tcp streams

libc-bottom-half/sources/wasip3_stdio.c
```diff
@@ -2,10 +2,155 @@
...
 #ifdef __wasip3__
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+#include <wasi/descriptor_table.h>
+#include <wasi/wasip3.h>
+
+typedef struct {
+  stdin_stream_u8_t input;
+  stdin_future_result_void_error_code_t future;
```
alexcrichton: Perhaps `input_future` as a name to clearly connect it to the `input` variable? Also mind adding some `//`-style comments explaining what these fields are and how the futures/streams/tasks all relate?

libc-bottom-half/cloudlibc/src/libc/unistd/read.c
cpetig: I was hoping to factorize out common parts between read and write and having both in a single file (file_util) helps with that, previously I had considerable code inlined here and in write.c.

libc-bottom-half/sources/wasip3_stdio.c
```diff
@@ -2,10 +2,155 @@
...
+  *offs = NULL;
+  return 0;
+}
+
+static int stdio3_read(void *data, void *buf, size_t nbyte,
+                       waitable_t *waitable, wasip3_waitable_status_t *out,
+                       off_t **offs) {
+  stdio3_t *stdio = (stdio3_t *)data;
+  if (!stdio->input) {
+    errno = EBADF;
```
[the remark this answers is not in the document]
cpetig: Oh, I just checked with the Linux manual page for write. EOPNOTSUPP makes sense to me as well.

libc-bottom-half/sources/wasip3_stdio.c
```diff
@@ -2,10 +2,155 @@
...
+                                    stdout_result_void_error_code_t *result),
+    bool (*terminal)(terminal_stdout_own_terminal_output_t *ret)) {
+  stdio3_t *stdio = calloc(1, sizeof(stdio3_t));
+  if (!stdio) {
+    errno = ENOMEM;
+    return -1;
+  }
+  stdin_stream_u8_t read_side = stdin_stream_u8_new(&stdio->stdout);
+  wasip3_subtask_status_t res = (*func)(read_side, &stdio->stdout_result);
+  stdio->stdout_task = WASIP3_SUBTASK_HANDLE(res);
```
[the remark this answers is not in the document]
cpetig: The next write will pick up the error anyway, do you think the open should fail immediately? Wouldn't init returning -1 render the file descriptor infrastructure broken (at least the first open will fail in `descriptor_table_insert`, IIRC).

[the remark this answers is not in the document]
alexcrichton: I agree that returning -1 shouldn't happen, but handling it in this case (IMO) should effectively move stdout into a state of "it's closed, don't try again". I'd imagine that this would trigger writes to start returning EPIPE (or maybe 0) or something like that, but basically the case that the function returns immediately should be used to update the internal structure vs returning an error

libc-bottom-half/cloudlibc/src/libc/unistd/read.c
alexcrichton: Yeah if there's a large shared body of code between read/write definitely worth splitting out. I'd imagine though that ABI-wise there's details of read/write that aren't shared which would make sense to go here (vs having this function effectively just tail-call another function), but I'm also ok waiting to see how the dust settles in wasip3 to see what the best organization is

libc-bottom-half/sources/wasip3_stdio.c
cpetig: As I returned both the future and the stream back from stdio3_get_read, I combined these fields. Still there could be a comment, but with only two elements the purpose now seems obvious to me.

libc-bottom-half/headers/private/wasi/wasip3_block.h
```diff
@@ -4,8 +4,13 @@
 #include <wasi/api.h>
 
 #ifdef __wasip3__
+#include <__typedef_ssize_t.h>
```
alexcrichton: Would it be possible to use `stdint.h` to get `ssize_t`? (that feels more conventional than `__*`-prefixed headers)

libc-bottom-half/headers/private/wasi/wasip3_block.h
```diff
@@ -4,8 +4,13 @@
...
+#include <wasi/descriptor_table.h> // for waitable_t
+
 // Waits for a subtask to return
 void wasip3_subtask_block_on(wasip3_subtask_status_t status);
+// Waits for a (stream) transfer to complete
+ssize_t wasip3_waitable_block_on(wasip3_waitable_status_t status, waitable_t stream);
```
alexcrichton: Given that this can't fail, I think `size_t` might be a more appropriate typing for the return value?

libc-bottom-half/sources/wasip3_stdio.c
```diff
@@ -2,10 +2,173 @@
...
+};
+
+static int stdio_add_input() {
+  stdin3_t *stdio = calloc(1, sizeof(stdin3_t));
+  if (!stdio) {
+    errno = ENOMEM;
+    return -1;
+  }
+  stdin_tuple2_stream_u8_future_result_void_error_code_t stdin;
+  stdin_read_via_stream(&stdio->input);
```
alexcrichton: What would you think of deferring this call, specificallly `stdin_read_via_stream` (and `stdin_stream_u8_new` + invocation of `func` below) to when the read/write stream is requested? That'd help lazy-initialize things on-demand as necessary and in theory could help resource consumption by entirely avoiding creation of these resources in some situations.

cpetig:
(by the way: I found that zero size reads/writes block, because they confuse it with eof/error, already fixed on the file PR)

PR #719, opened 2026-01-15
Update bindings generation and bump WASIP3 to newest RC

TartanLlama:
The bindings generator currently uses the now-archived [WebAssembly/wasi-cli](https://github.com/WebAssembly/wasi-cli/) repo to download the wit files for wasip2 and wasip3. This repo is no longer being updated, and there is now a wasip3 release candidate that does not have wit files released in that repo.

This PR:
- updates the bindings generation to pull down wit files with [`wkg`](https://github.com/bytecodealliance/wasm-pkg-tools).
- bumps the targeted wasip3 version from rc-2025-09-16 to rc-2026-01-06
- updates the wasip3 clocks implementation to use the names that were changed in the latest RC release


libc-bottom-half/cloudlibc/src/common/time.h
```diff
@@ -18,8 +18,11 @@
...
 typedef __wasi_timestamp_t wasilibc_timestamp_t;
-#elif defined(__wasip2__) || defined(__wasip3__)
+#elif defined(__wasip2__)
 typedef wall_clock_datetime_t wasilibc_timestamp_t;
+#elif defined(__wasip3__)
+typedef filesystem_instant_t wasilibc_timestamp_t;
+typedef monotonic_clock_mark_t monotonic_clock_instant_t;
 #else
 # error "Unknown WASI version"
 #endif
```
ricochet: The wasi_clocks_system_clock_instant_t.seconds is now int64_t (signed) rather than uint64_t

We'll need to treat p2 and p3 differently L44-47 right? timespec->tv_sec is time_t which is typically long.

.github/workflows/main.yml
```diff
@@ -284,7 +284,7 @@ jobs:
     - name: Setup `wasmtime`
       uses: bytecodealliance/actions/wasmtime/setup@v1
       with:
-        version: "38.0.3"
+        version: "40.0.2"
```
alexcrichton: For this using `dev` I think could get things working today

PR #727, opened 2026-02-01
Wasip3: poll implementation

cpetig:
Without the separate file nor stream support this doesn't change any test to work, yet. (The code worked for me with tcp in the past, though)


alexcrichton:
That's a good point yeah, and I agree it'd be reasonable to use that to report back more events

alexcrichton:
With https://github.com/WebAssembly/wasi-libc/pull/782 I believe `poll` is now complete, so I'm going to close this. Thanks again for getting this started though!

PR #729, opened 2026-02-03
Refactor the `sockets_utils.h` headers

alexcrichton:
This commit refactors the helpers for TCP/UDP which predated much of the work around WASIp2-natively-in-libc. Changes here are:

* Helpers to use `__wasilibc_*` prefix that many other internal functions in wasi-libc use.
* Helpers directly set `errno` instead of returning a `bool` and requiring the caller to set `errno`.
* Some helpers were renamed/tweaked to be a bit more idiomatic with C.


rvolosatovs (approved):
I'm still new to the codebase, but the changes look good to me!

PR #731, opened 2026-02-11
Restructure thread implementations

TartanLlama:
This PR restructures the current two `pthreads` implementations into parallel directories to prepare for a third implementation for `wasip3`.

- Move the single-threaded stubbed implementation from `wasi-libc/thread-stub` to `wasi-libc/libc-top-half/musl/src/thread/single-threaded`
- Move the `wasi-threads` implementation from `wasi-libc/libc-top-half/musl/src/thread` to `wasi-libc/libc-top-half/musl/src/thread/wasi-threads`
- Move common files into `wasi-libc/libc-top-half/musl/src/thread/common`
- Delete the unused architecture subfolders from `wasi-libc/libc-top-half/musl/src/thread` 

The implementation for wasip3 will live in `wasi-libc/libc-top-half/musl/src/thread/coop-threads`, which will be created and populated in future PRs.


TartanLlama:
Yeah, the pthreads impl relies a lot on the musl internals for the opaque pthreads struct. @sunfishcode and I chatted about this briefly last week and our conclusion was that, given the low volume of upstream musl updates, it seems more valuable to focus on the maintainability of `wasi-libc` as its own implementation than to optimize for ease of merging upstream patches.

For the coop threading implementation, I started off just `ifdef`ing everywhere I needed changes, but it quickly became a bit of a mess for some of the more complex functions, and many of them just have entirely different implementations that would be living in the same files.

PR #732, opened 2026-02-11
don't mark just-connected socket ready-to-read in `poll(2)`

dicej:
This fixes a bug in the `poll(2)` implementation for `wasm32-wasip2` such that it spuriously set the `POLLRDNORM` bit on the `revents` field for a socket which just finished connecting even though there's nothing immediately available to read.  The fix is to set at most the `POLLWRNORM` bit in that situation.

### DIFF
--- a/test/src/poll-connect.c
+++ b/test/src/poll-connect.c
@@ -47,8 +47,8 @@ int main() {
   int client = socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK, 0);
   TEST2(client != -1);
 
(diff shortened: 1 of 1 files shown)


alexcrichton:
CI looks mostly realted to failing to build the test on WASIp1 in which case it's fine to `#ifdef` it all away, but there's also a [spurious failure](https://github.com/WebAssembly/wasi-libc/actions/runs/21913983606/job/63275867344?pr=732) for this test specifically

PR #746, opened 2026-02-20
WASIP3 Locking/Waiting Primitives

TartanLlama:
Includes https://github.com/WebAssembly/wasi-libc/pull/744

The first batch of changes for cooperative threading. This PR introduces the core locking and waiting primitives that will be used to implement many of the pthreads features, and reimplements the existing locks used throughout the codebase in terms of them.

- Implements `sched_yield` in terms of `wasip3_thread_yield`
- Adds a guest-space waitlist implementation in `__wait.c` consisting of:
    - `struct __waitlist_node`: a singly-linked list of TIDs
    - `__waitlist_wait_on`: adds the current thread to the waitlist, allocating the node on the stack 
    - `__waitlist_wake_(one,all)`: yields directly to one thread in the waitlist and unsuspends the rest
- Adds a cooperative lock implementation implemented in terms of the waitlist to `__lock.c`
    - Possible re-entrance of libc function is somewhat complicated by cooperative threading, because there are three relevant classes of sections of code: 
        - Sections that may be executed concurrently
        - Sections that may not be executed concurrently, but that contain no context-switch points (yields, import calls, indirect calls)
        - Sections that may not be executed concurrently, and that *do* contain context-switch points
    - To reflect this and avoid unnecessary locking/unlocking for sections in the second class, this PR adds a `WEAK_(UN)LOCK` macro in addition to the existing `(UN)LOCK`, where the former is a no-op in co-operative threading builds
    -  I'm not married to these names, and we may decide that the added complexity is not worth it, but I've left this in for now
- Existing uses of futexes are rewritten to use the cooperative lock when building for WASIP3

Because these patches do not yet include the ability to spawn new threads, which will require https://github.com/llvm/llvm-project/pull/175800, there is no additional testing added yet. Testing will be added as part of integration of the Open POSIX Test Suite, which will be coming in future PRs.


libc-top-half/musl/src/stdio/__lockfile.c
```diff
@@ -1,6 +1,24 @@
...
+#include <wasi/version.h>
 
+#ifdef __wasi_cooperative_threads__
+#include "lock.h"
+int __lockfile(FILE *f)
+{
+	// Allow recursive locking
+	if (f->lock.owner == __pthread_self()->tid)
+		return 0;
+
```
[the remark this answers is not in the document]
alexcrichton: (or does musl already handle that case elsewhere via the return value of this function?)

If musl already handles this, a comment similar to below where I'd request that the testing of `.owner` be moved to some sort of function in the lock header file

libc-top-half/musl/src/malloc/oldmalloc/malloc.c
```diff
@@ -27,6 +27,16 @@ static struct {
...
 
+#ifdef __wasi_cooperative_threads__
+/* The implementation has no yield points, so locks can be no-ops. */
+static inline void lock(volatile int *lk)
+{
+}
+
+static inline void unlock(volatile int *lk)
+{
+}
```
alexcrichton: Could this compile down to some small debug-mode-only instrumentation perhaps?

libc-top-half/musl/src/internal/stdio_impl.h
```diff
@@ -8,10 +8,21 @@
...
 
-#if defined(__wasilibc_unmodified_upstream) || defined(_REENTRANT)
+#include "lock.h"
+#include <wasi/version.h>
+
+#ifdef __wasi_cooperative_threads__
+#define FFINALLOCK(f) __lockfile((f))
+#define FLOCK(f) int __need_unlock = __lockfile((f))
+#define FUNLOCK(f) do { if (__need_unlock) __unlockfile((f)); } while (0)
+#define __STDIO_LOCK_INIT {0, 0}
```
alexcrichton: Aha looks like musl handles my concern below (above in review? who knows...). Not exactly how I'd imagine it being done but hey here it is

libc-top-half/musl/src/internal/stdio_impl.h
```diff
@@ -8,10 +8,21 @@
...
+#define __STDIO_LOCK_RESET(lock) do { (lock)->owner = 0; (lock)->waiters = NULL; } while (0)
```
alexcrichton: This might be replacable with `*lock = __STDIO_LOCK_INIT` given what it's doing, but this is all buried in musl code so seems reasonable either way, happy to defer to you

alexcrichton (approved):
I was waiting for CI to be green before digging in much further here, but in retrospect there's no need as I don't think the code will change that much. I've left some comments below but overall this looks all good to me, thanks! 

If you need help with CI just lemme know and I can try to help frob things.

PR #748, opened 2026-02-24
merge TCP implementations in `tcp.c`

rvolosatovs:
Merge TCP implementations in `tcp.c` as suggested in https://github.com/WebAssembly/wasi-libc/pull/745#issuecomment-3937107018

### DIFF
--- a/expected/wasm32-wasip3/defined-symbols.txt
+++ b/expected/wasm32-wasip3/defined-symbols.txt
@@ -339,8 +339,6 @@ __wasilibc_rmdirat
 __wasilibc_sockaddr_to_wasi
 __wasilibc_sockaddr_validate
 __wasilibc_stat
-__wasilibc_tcp_getsockopt
-__wasilibc_tcp_setsockopt
 __wasilibc_tell
 __wasilibc_unlinkat
 __wasilibc_unspecified_addr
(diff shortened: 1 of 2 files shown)


libc-bottom-half/sources/tcp.c
```diff
@@ -764,31 +853,408 @@ static int tcp_fcntl_setfl(void *data, int flags) {
...
+    return -1;
+  }
+
+  // Copy out integer value.
+  memcpy(optval, &value, *optlen < sizeof(int) ? *optlen : sizeof(int));
+  *optlen = sizeof(int);
+  return 0;
+}
+
+int __wasilibc_tcp_setsockopt(void *data, int level, int optname,
```
alexcrichton: Mind switching this back to `static int tcp_setsockopt`? (no need for `__wasilibc` and `static` helps keep the symbol private)

expected/wasm32-wasip3/defined-symbols.txt
```diff
@@ -1394,6 +1394,7 @@ tanh
 tanhf
 tanhl
 tanl
+tcp_getpeername
```
alexcrichton: This function is actually mistakenly missing `static`, mind adding it there to avoid this addition (and also remove it from the p2 target)

Issue #474, opened 2024-02-14
locking issues in chdir.c


yamt:
in make_absolute:
- __wasilibc_cwd_unlock is called twice
- static vars (make_absolute_buf/make_absolute_len) are used w/o lock

in chdir:
- static vars (relative_buf/relative_buf_len/__wasilibc_cwd_mallocd) are used w/o lock

PR #781, opened 2026-04-07
Make `make_absolute` thread-local in `chdir.c`

CPunisher:
`make_absolute` previously used a process-wide static buffer to assemble absolute paths for relative path resolution. In `_REENTRANT` builds, concurrent calls could overwrite or reallocate that shared buffer while another thread was still using it, leading to races and potentially invalid pointers.

Related https://github.com/WebAssembly/wasi-libc/issues/474


CPunisher:
Yeah thanks for your trust. I'm willing to continue, and I'll also consider the best approach.

PR #782, opened 2026-04-09
wasip3: Finish nonblocking I/O implementation for TCP

alexcrichton:
This commit fills out the final bits of nonblocking I/O for TCP sockets. This necessitates an implementation of nonblocking I/O for `stream<u8>` on both the reading and the writing side of things. Additionally things like TCP accepts now work, too. All tests are now passing on the WASIp3 target and the `FAILP3` directive is now removed since it's no longer necessary.

The internals of the implementation here are pretty gnarly. This is due to the need to bridge the readiness-based-model of `poll` with the completion-based model of the component model. This implementation contains optimizations for 0-length reads/writes to use those to signal readiness if the host supports that, but this falls back to internally-buffered reads/writes if that is not supported.

This should resolve all remaining TODOs for WASIp3 that I know of at least in wasi-libc, so after this it'd be switching to bug-finding mode and handling bug reports as projects are compiled with wasip3.

### DIFF
--- a/libc-bottom-half/sources/file_utils.c
+++ b/libc-bottom-half/sources/file_utils.c
@@ -392,7 +392,7 @@ ssize_t __wasilibc_write(wasi_write_t *write, const void *buffer,
     // If this stream isn't forced to buffer then do an opportunistic write
     // with the user-provided buffer. If this doesn't work (it was blocked
     // instead of completing), but it should work (because a previous
-    // zero-length operation succeeded which can sometimes mean it's erady),
+    // zero-length operation succeeded which can sometimes mean it's ready),
     // then it means we must buffer from now on. No matter what though if the
     // write is in-progress then it's cancelled here as the user's input buffer
     // won't persist beyond this function call. Throughout writing/cancellation
--- a/libc-bottom-half/sources/file_utils.c
+++ b/libc-bottom-half/sources/file_utils.c
@@ -82,6 +82,196 @@ int wasi_string_from_c(const char *s, wasi_string_t *out) {
   return 0;
 }
 
+#ifndef __wasip2__
+/// Update `state` with the result of `code` that happened.
+static size_t wasip3_io_update_code(wasip3_io_state_t *state,
+                                    wasip3_waitable_status_t code) {
+  assert(code != WASIP3_WAITABLE_STATUS_BLOCKED);
+  switch (WASIP3_WAITABLE_STATE(code)) {
+  case WASIP3_WAITABLE_COMPLETED:
+  case WASIP3_WAITABLE_CANCELLED:
+    break;
+  case WASIP3_WAITABLE_DROPPED:
+    state->flags |= WASIP3_IO_DONE;
+    break;
+  default:
+    assert(0 && "unexpected waitable state");
+    abort();
+  }
+
+  assert(state->flags & WASIP3_IO_INPROGRESS);
+  state->flags &= ~WASIP3_IO_INPROGRESS;
+
+  if (state->flags & WASIP3_IO_ZERO_INPROGRESS) {
+    state->flags &= ~WASIP3_IO_ZERO_INPROGRESS;
+
+    assert(!(state->flags & WASIP3_IO_SHOULD_BE_READY));
+    state->flags |= WASIP3_IO_SHOULD_BE_READY;
+  }
+
+  return WASIP3_WAITABLE_COUNT(code);
+}
+
+/// Update `write` with the result of `event` that happened.
+static size_t wasip3_io_update_event(wasip3_io_state_t *state,
+                                     wasip3_event_t *event) {
+  assert(event->event == WASIP3_EVENT_STREAM_WRITE ||
+         event->event == WASIP3_EVENT_STREAM_READ);
+  assert(event->waitable == state->stream);
+  return wasip3_io_update_code(state, event->code);
+}
+
+/// Attempts to resolve any pending write that may be in-progress on `write`.
+///
+/// This may notably end up issuing more writes to finish a buffered write that
+/// was previously flagged as completed.
+static int wasip3_write_resolve_pending(wasi_write_t *write) {
+  wasip3_event_t event;
+  wasip3_io_state_t *state = write->state;
+
+  // If there's nothing in-progress, then there's nothing to do, so bail out.
+  if (!(state->flags & WASIP3_IO_INPROGRESS))
+    return 0;
+  assert(state->buf);
+
+  // This loop ensures that the entirety of `write->buf` is written out. Libc
+  // already reported that the write succeeded so if a short write happens then
+  // it's got to get started back up again.
+  while (1) {
+    // For blocking I/O this awaits the previous result. For non-blocking I/O a
+    // `poll` operation is done.
+    if (write->blocking) {
+      if (!__wasilibc_waitable_block_on(state->stream, &event,
+                                        write->timeout)) {
+        errno = ETIMEDOUT;
+        return -1;
+      }
+    } else {
+      __wasilibc_poll_waitable(state->stream, &event);
+      if (event.event == WASIP3_EVENT_NONE) {
+        errno = EWOULDBLOCK;
+        return -1;
+      }
+    }
+
+    // Update the I/O internal state given the result of the write.
+    state->buf_start += wasip3_io_update_event(state, &event);
+
+    // While there's remaining writes to perform, kick those off here. If a
+    // write is blocked then nonblocking mode returns as such and blocking
+    // mode breaks out to turn this outer `while (1)` loop again to block
+    // on the result. If the write finishes immediately then state is updated
+    // again and then further continues.
+    while (!(state->flags & WASIP3_IO_DONE) &&
+           state->buf_start != state->buf_end) {
+      wasip3_waitable_status_t status = filesystem_stream_u8_write(
+          state->stream, state->buf + state->buf_start,
+          state->buf_end - state->buf_start);
+      state->flags |= WASIP3_IO_INPROGRESS;
+      if (status == WASIP3_WAITABLE_STATUS_BLOCKED) {
+        if (write->blocking) {
+          break;
+        } else {
+          errno = EWOULDBLOCK;
+          return -1;
+        }
+      }
+      state->buf_start += wasip3_io_update_code(state, status);
+    }
+
+    // We either broke out of the `while` loop normally, or we hit the `break`
+    // in the loop which wants to continue to the top of this outer loop. Test
+    // which it is here and act accordingly.
+    if ((state->flags & WASIP3_IO_DONE) || state->buf_start == state->buf_end) {
+      free(state->buf);
+      state->buf = NULL;
+      state->buf_start = 0;
+      state->buf_end = 0;
+      return 0;
+    }
+  }
+
+  abort();
+}
+
+/// Helper of `read`/`write` starting below.
+static bool wasip3_start_zero_length(wasip3_io_state_t *state,
+                                     wasip3_waitable_status_t status) {
+  assert(!(state->flags & WASIP3_IO_INPROGRESS));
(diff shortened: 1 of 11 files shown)


dicej (approved):
🎉

alexcrichton:
I'm going to go ahead and merge this @badeend but I'd still value your input if you get a chance in the future to take a look, happy to have follow-ups!

libc-bottom-half/sources/file_utils.c
```diff
@@ -82,6 +82,196 @@ int wasi_string_from_c(const char *s, wasi_string_t *out) {
...
+  state->buf = malloc(length);
+  if (!state->buf) {
+    errno = ENOMEM;
+    return -1;
+  }
+  state->buf_start = 0;
+  state->buf_end = 0;
+  wasip3_waitable_status_t status =
+      filesystem_stream_u8_read(state->stream, state->buf, length);
+  state->flags |= WASIP3_IO_INPROGRESS;
```
badeend, line 250: If `filesystem_stream_u8_read` completes immediately, then `state->flags` is left behind with the `WASIP3_IO_INPROGRESS` flag set, despite there not being anything in progress. Should this be handled here or is this already handled by the callers of this function somewhere?

`buf_end` likely needs to be incremented in that case also

libc-bottom-half/sources/tcp.c
```diff
@@ -290,6 +300,58 @@ static void tcp_setup_connected_state_wasip3(tcp_socket_t *socket) {
...
+  while (!(state->flags & TCP_LISTENING_DONE) &&
+         !(state->flags & TCP_LISTENING_ACCEPTING) &&
+         !(state->flags & TCP_LISTENING_ACCEPT_READY)) {
+    wasip3_waitable_status_t status = sockets_stream_own_tcp_socket_read(
+        state->stream, &state->accept_result, 1);
+    if (status == WASIP3_WAITABLE_STATUS_BLOCKED) {
+      state->flags |= TCP_LISTENING_ACCEPTING;
+      return true;
+    }
+    wasip3_tcp_accept_finish(state, status);
```
badeend, line 349: The first line of the `wasip3_tcp_accept_finish`'s implementation asserts that `TCP_LISTENING_ACCEPTING` should be set. Yet, the invocation here can only be reached while `TCP_LISTENING_ACCEPTING` is _not_ set.

libc-bottom-half/sources/file_utils.c
```diff
@@ -82,6 +82,196 @@ int wasi_string_from_c(const char *s, wasi_string_t *out) {
...
+/// Starts a zero-length read on the stream pointed to by `read`.
+///
+/// This requires that there's no active I/O on the stream at this time and
+/// that it's not a closed stream.
+static bool wasip3_read_start_zero_length(wasip3_io_state_t *state) {
+  return wasip3_start_zero_length(
+      state, filesystem_stream_u8_read(state->stream, NULL, 0));
+}
+
+/// Starts a zero-length read on the stream pointed to by `read`.
```
badeend, line 230:
```suggestion
/// Starts a nonzero-length read on the stream pointed to by `state`.
```

libc-bottom-half/sources/file_utils.c
```diff
@@ -82,6 +82,196 @@ int wasi_string_from_c(const char *s, wasi_string_t *out) {
...
+    state->flags &= ~WASIP3_IO_ZERO_INPROGRESS;
+
+    assert(!(state->flags & WASIP3_IO_SHOULD_BE_READY));
+    state->flags |= WASIP3_IO_SHOULD_BE_READY;
+  }
+
+  return WASIP3_WAITABLE_COUNT(code);
+}
+
+/// Update `write` with the result of `event` that happened.
```
badeend, line 115:
```suggestion
/// Update `state` with the result of `event` that happened.
```

libc-bottom-half/sources/file_utils.c
```diff
@@ -82,6 +82,196 @@ int wasi_string_from_c(const char *s, wasi_string_t *out) {
...
+  assert(!(state->flags & WASIP3_IO_DONE));
+  state->flags |= WASIP3_IO_INPROGRESS;
+  state->flags |= WASIP3_IO_ZERO_INPROGRESS;
+  if (status == WASIP3_WAITABLE_STATUS_BLOCKED)
+    return true;
+  wasip3_io_update_code(state, status);
+  return false;
+}
+
+/// Starts a zero-length write on the stream pointed to by `write`.
```
badeend, line 212:
```suggestion
/// Starts a zero-length write on the stream pointed to by `state`.
```

libc-bottom-half/sources/file_utils.c
```diff
@@ -82,6 +82,196 @@ int wasi_string_from_c(const char *s, wasi_string_t *out) {
...
+///
+/// This requires that there's no active I/O on the stream at this time and
+/// that it's not a closed stream.
+static bool wasip3_write_start_zero_length(wasip3_io_state_t *state) {
+  return wasip3_start_zero_length(
+      state, filesystem_stream_u8_write(state->stream, NULL, 0));
+}
+
+/// Starts a zero-length read on the stream pointed to by `read`.
```
badeend, line 221:
```suggestion
/// Starts a zero-length read on the stream pointed to by `state`.
```

libc-bottom-half/sources/file_utils.c
```diff
@@ -159,22 +349,134 @@ ssize_t __wasilibc_write(wasi_write_t *write, const void *buffer,
...
+      // again if zero-length reads/writes don't actually work.
+      assert(state->flags & WASIP3_IO_SHOULD_BE_READY);
+      continue;
+    }
+
+    break;
+  }
+
+  // At this point we're in nonblocking mode, we're required to buffer data,
+  // and most other flags should all be turned off. Here the input `bufffer` is
```
badeend, line 443:
```suggestion
  // and most other flags should all be turned off. Here the input `buffer` is
```

libc-bottom-half/sources/file_utils.c
```diff
@@ -82,6 +82,196 @@ int wasi_string_from_c(const char *s, wasi_string_t *out) {
...
+static bool wasip3_read_start_zero_length(wasip3_io_state_t *state) {
+  return wasip3_start_zero_length(
+      state, filesystem_stream_u8_read(state->stream, NULL, 0));
+}
+
+/// Starts a zero-length read on the stream pointed to by `read`.
+///
+/// This requires that there's no active I/O on the stream at this time and
+/// that it's not a closed stream.
+static int wasip3_read_start_nonzero(wasip3_io_state_t *state, size_t length) {
```
badeend, line 234: Could you throw an `assert(length > 0);` in here too?

IIUC, control flow may actually end up here with a `length` of 0 when `WASIP3_IO_MUST_BUFFER` is set. In that case the `malloc(length)` call below may legitimately return NULL. I feel something higher up the stack (`__wasilibc_read`?) should handle the zero-length read case.

libc-bottom-half/sources/file_utils.c
```diff
@@ -159,22 +349,134 @@ ssize_t __wasilibc_write(wasi_write_t *write, const void *buffer,
...
+  // At this point we're in nonblocking mode, we're required to buffer data,
+  // and most other flags should all be turned off. Here the input `bufffer` is
+  // copied into the internal `state` and the I/O operation is issued.
+  assert(state->flags & WASIP3_IO_MUST_BUFFER);
+  assert(!(state->flags & WASIP3_IO_SHOULD_BE_READY));
+  assert(!(state->flags & WASIP3_IO_INPROGRESS));
+  assert(!(state->flags & WASIP3_IO_ZERO_INPROGRESS));
+  assert(!(state->flags & WASIP3_IO_DONE));
+  assert(state->buf == NULL);
+  state->buf = malloc(length);
```
badeend, line 451: The `length` passed to `__wasilibc_write` may be `0`.

libc-bottom-half/sources/tcp.c
badeend, line 197: This is handing out a reference to a potentially zeroed out `state->receive` struct after `tcp_shutdown` has been called. (Same for `&state->send` in `tcp_get_write_stream`)

From the user's POV, reading from a shut down socket should return gracefully with 0 bytes being read. And writing to it should return EPIPE.
I suspect that the current code traps on the zeroed out stream handle.

libc-bottom-half/sources/file_utils.c
```diff
@@ -245,23 +547,195 @@ ssize_t __wasilibc_read(wasi_read_t *read, void *buffer, size_t length) {
...
+    // reading into.
+    state->buf_end = wasip3_io_update_event(state, &event);
+  }
+
+  assert(!(state->flags & WASIP3_IO_INPROGRESS));
+
+  // If this stream has internally buffered data that's ready to read then copy
+  // that out here.
+  if (state->buf)
+    return wasip3_read_complete_internally(state, buffer, length);
```
badeend, line 578: Not all exit paths update the `read->offset`. Sockets don't have an offset, but maybe regular file reads could be affected by this? (same for `__wasilibc_write`)

libc-bottom-half/sources/tcp.c
```diff
@@ -807,36 +908,95 @@ static int tcp_shutdown(void *data, int posix_how) {
...
     if ((events & POLLRDNORM) != 0) {
+#ifdef __wasip2__
       if (__wasilibc_poll_add_input_stream(
-              state, streams_borrow_input_stream(socket->state.connected.input),
-              &socket->state.connected.input_pollable) < 0)
+              state, streams_borrow_input_stream(conn->input),
+              &conn->input_pollable) < 0)
         return -1;
+#else
+      if (__wasilibc_read_poll(&conn->receive, state) < 0)
```
badeend, line 986: `conn->receive` may be zeroed out after this direction has been shut down.

libc-bottom-half/sources/file_utils.c
```diff
@@ -159,22 +349,134 @@ ssize_t __wasilibc_write(wasi_write_t *write, const void *buffer,
...
-  return write->eof(write->eof_data);
+
+  // If that write actually succeeded, however, then deal with it here. That
+  // means we have to deallocate the internal buffer and then handle the result
+  // of the write.
+  free(state->buf);
+  state->buf = NULL;
+  state->buf_start = 0;
+  state->buf_end = 0;
+  return wasip3_io_update_code(state, status);
```
badeend, line 479: Should this exit point include a check+call to `write->eof` ? Same question for `__wasilibc_read` where there are a couple of exits that don't explicitly check for end-of-straem.

libc-bottom-half/sources/tcp.c
badeend, line 177: Does the C/M automatically drop `future` handles after awaiting them? If not, then this code should explicitly drop the future before setting the field to 0. (same for tcp_write_eof)

badeend, line 172: this assert doesn't seem right

badeend (commented):
@alexcrichton I reviewed this change and found a few potential issues. Note that I'v only read the code and haven’t confirmed every suspicion. It could very well be that things are already handled properly, but just some place else that Ive missed.

alexcrichton:
Thanks @badeend! I believe I've handled all your comments now through these PRs:

https://github.com/WebAssembly/wasi-libc/pull/793
https://github.com/WebAssembly/wasi-libc/pull/792
https://github.com/WebAssembly/wasi-libc/pull/791
https://github.com/WebAssembly/wasi-libc/pull/790
https://github.com/WebAssembly/wasi-libc/pull/789
https://github.com/bytecodealliance/wasmtime/pull/13163
https://github.com/WebAssembly/wasi-libc/pull/788
https://github.com/WebAssembly/wasi-libc/pull/787
https://github.com/bytecodealliance/wasmtime/pull/13158
https://github.com/WebAssembly/wasi-libc/pull/786
https://github.com/WebAssembly/wasi-libc/pull/785

PR #783, opened 2026-04-13
wasip3: Perform a thread yield in `poll` with no timeout

alexcrichton:
This commit updates the `poll` implementation on wasip3 to insert a call to the canonical `thread.yield` function in the case that there's no timeout and nothing is ready yet. This explicitly provides the host an opportunity to deliver events and resolve any potential otherwise-deadlock if the guest is spinning waiting for events to happen with a timeout of 0.

This is related to bytecodealliance/wasmtime#13040 and is a partial guest-side solution for what's outlined there.


dicej (approved):
Thanks!

PR #786, opened 2026-04-21
Add a missing `drop-readable` call

alexcrichton:
I'm not sure how to add a test for this so the fix here is just in isolation.

### DIFF
--- a/libc-bottom-half/sources/tcp.c
+++ b/libc-bottom-half/sources/tcp.c
@@ -174,6 +174,7 @@ static int tcp_read_eof(void *data) {
     __wasilibc_future_block_on(sockets_future_result_void_error_code_read(
                                    state->receive_result, &result),
                                state->receive_result);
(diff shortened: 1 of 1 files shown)


libc-bottom-half/sources/tcp.c
```diff
@@ -174,6 +174,7 @@ static int tcp_read_eof(void *data) {
     __wasilibc_future_block_on(sockets_future_result_void_error_code_read(
                                    state->receive_result, &result),
                                state->receive_result);
+    sockets_future_result_void_error_code_drop_readable(state->receive_result);
```
badeend, line 177: Maybe also take a look at tcp_write_eof

badeend (approved):

dicej (approved):

alexcrichton:
If this becomes too much of a problem in the future, too, I might take a look at enabling a mode of the CLI to assert that all handles are closed by the time `main` exits and add support to wasi-libc for that.

PR #788, opened 2026-04-21
wasip3: Fix fallback code when buffering writes

alexcrichton:
This commit fixes mistakes from #782 where the fallback code for buffering writes wasn't handled properly. Specifically when a completion event came in through `poll` it didn't fully update the internal state of the stream, such as kicking off further writes or cleaning up the internal buffer. This is fixed by refactoring the logic already present during a normal `write` to get executed in this event.

Testing this is unfortunately not trivial. There aren't really all that many streams we have access to natively in Wasmtime, and TCP is "well behaved" and doesn't execute any of these fallbacks anyway. The closest way to test this is filesystem-based streams, but Wasmtime's default behavior is to lie and just do blocking operations on these streams regardless. Through changes in Wasmtime, and flags to Wasmtime, however, it's possible to Wasmtime in a mode where filesystem streams are "ill behaved" where zero-length reads don't signal readiness, triggering these fallback paths. With those changes the added test here, and minor additionally support for nonblocking reads/writes, exercise these paths and triggers the previous bug.


alexcrichton:
The companion Wasmtime PR for this is https://github.com/bytecodealliance/wasmtime/pull/13163 which exercises these tests more thoroughly.

libc-bottom-half/sources/file.c
```diff
@@ -123,7 +123,7 @@ static int file_get_read_stream(void *data, wasi_read_t *read) {
 #endif
   read->offset = &file->offset;
   read->timeout = 0;
-  read->blocking = true;
+  read->blocking = (file->oflag & O_NONBLOCK) == 0;
```
badeend, line 126: Does this mean `read`/`write` can start returning `EWOULDBLOCK` for regular files?

alexcrichton: It does, yeah. I'm not sure how useful that is in practice, but it's the only way I could think of to test this PR

libc-bottom-half/sources/file_utils.c
```diff
@@ -121,6 +121,48 @@ static size_t wasip3_io_update_event(wasip3_io_state_t *state,
...
+        filesystem_stream_u8_write(state->stream, state->buf + state->buf_start,
+                                   state->buf_end - state->buf_start);
+    state->flags |= WASIP3_IO_INPROGRESS;
+    if (status == WASIP3_WAITABLE_STATUS_BLOCKED)
+      return false;
+    state->buf_start += wasip3_io_update_code(state, status);
+  }
+
+  // We either broke out of the `while` loop normally, or we hit the `break`
+  // in the loop which wants to continue to the top of this outer loop. Test
```
dicej: This comment appears to have been copied without being updated for its new context; the `break`, outer loop, and test to which it refers no longer exist.

dicej (approved):

libc-bottom-half/sources/file.c
badeend: I suppose if an application requested O_NONBLOCK, then getting EWOULDBLOCK is exactly what it asked for. Nonetheless, I wonder how many programs are actually prepared to handle that for regular files, given ~40 years of assumptions that the flag is ignored for those.

alexcrichton: Er sorry meant to think about this a bit more before landing. Do you feel that this should get backed out? I don't know how to test anything related to "ill behaving streams" otherwise since files are basically the only host-defined stream that doesn't handle zero-length reads/writes like TCP does.

badeend: The Linux man page mentions:

> Note that this flag has no effect for regular files and block devices; that is, I/O operations will (briefly) block when device activity is required, regardless of whether `O_NONBLOCK` is set.

And _also_:

> Since `O_NONBLOCK` semantics might eventually be implemented, applications should not depend upon blocking behavior when specifying this flag for regular files and block devices.

So yeah,. there's that.. :)

---

I feel this leaks testing concerns into user-visible behavior so I'm leaning toward keeping the fix but backing out the nonblocking behavior for regular files. I don't have a super strong well-funded opinion for it though. 

That would indeed leave some of the new code untested. As long as the test suite uses off-the-shelf, unmodified Wasm engines, we're unlikely to hit all code paths anyway. A longer-term solution might be some kind of host-side "chaos mode" that deterministically generates all kinds of odd-but-technically-correct stream behaviors. That could also be more generally useful for other languages and toolchains targeting the component model directly. I realize that is completely out of scope for this PR.

alexcrichton: Hm yeah ok I agree and it seems prudent. I've posted https://github.com/WebAssembly/wasi-libc/pull/796 to disable this, and for further testing -- in addition to a theoretical chaos mode this'd be a good use of https://github.com/WebAssembly/wasi-libc/issues/766 I believe.

badeend: ah right, this could indeed be a nice use case for component composition

PR #798, opened 2026-06-11
Update WASIp3 WITs to 0.3.0

alexcrichton:
This commit is intended to couple the WASI subgroup's decision today to stamp 0.3.0 and ship it. This updates the vendored WITs here to their 0.3.0 versions which means that the WASIp3 version of libc will now be built against this version. I've tested this against a build of bytecodealliance/wasmtime#13612 to verify these changes.

This has a few minor changes as well such as:

* The implementation of `exit` in libc now uses `exit_with_code` instead of `exit` where possible. This is only done if the `status` being exited with fits within the `u8` that the WASI interface allows.

* Before cancelling/deleting the timeout task in `poll` it's removed from the `waitable-set` to accommodate upstream changes in Wasmtime and avoid a trap.


alexcrichton:
I'll note that this'll fail until there's a merged build of bytecodealliance/wasmtime#13612 to run tests with as CI will be, expectedly, failing.

wasi/p3/wit/deps/wasi-http-0.3.0/package.wit
```diff
@@ -1,13 +1,13 @@
-package wasi:http@0.3.0-rc-2026-03-15;
+package wasi:http@0.3.0;
```
pchickey, line 1: Like in the corresponding Wasmtime PR, this is also missing https://github.com/WebAssembly/WASI/pull/920/changes, so did that diff just not make it into the released wits? Its not the end of the world since it can be included in 0.3.1 but I just wanted to check

alexcrichton: Looks like this is due to https://github.com/bytecodealliance/wasm-tools/pull/2540, e.g. publication processes in wasi and latent issues in wasm-tools. This should get fixed by 0.3.1 and in the meantime shouldn't cause any undue issues as the documentation here doesn't actually go anywhere through libc.

pchickey (approved):
Modulo whatever minor stuff is going on with the failing tests

PR #799, opened 2026-06-12
WASIp3 Cooperative Threading

TartanLlama:
Currently a draft while the LLVM patches merge and I get all of this ready for review


libc-top-half/musl/src/thread/coop-threads/sem_timedwait.c
```diff
@@ -0,0 +1,34 @@
...
+		sem->__count--;
+		return 0;
+	}
+	
+	/* Would need to block, but timeouts not supported in cooperative threading.
+	 * There's no way to interrupt a suspended thread after a timeout. */
+	if (at) {
+		errno = ENOSYS;
+		return -1;
+	}
```
alexcrichton: Would it make sense to thread `at` into `__waitlist_wait_on` and have the error here generated by that function? That way there's sort of a centralized "this isn't supported" as opposed to having to make sure to write this down in all functions with a timeout parameter

libc-top-half/musl/src/thread/coop-threads/pthread_cond_timedwait.c
```diff
@@ -0,0 +1,15 @@
+#include "pthread_impl.h"
+
+int __pthread_cond_timedwait(pthread_cond_t *restrict c, pthread_mutex_t *restrict m, const struct timespec *restrict ts)
+{
+	// TODO(wasip3) timed waits
+	if (ts) {
+		return ETIMEDOUT;
+	}
```
alexcrichton: This one might be difficult to leave unsupported into the future as AFAIK waiting with timeouts on condvars is a relatively common operation. Leaving this as a TODO for now though is fine by me

libc-top-half/musl/src/thread/coop-threads/__wasi_coop_thread_start.s
```diff
@@ -0,0 +1,33 @@
...
+	# Set up the minimum C environment.
+	local.get   0  # start_arg
+	i32.load    0  # stack
+	call __wasm_set_stack_pointer
+
+	local.get   0  # start_arg
+	i32.load    4  # tls_base
+	call __wasm_set_tls_base
+	# Make the C function do the rest of work.
+    call __wasm_component_model_builtin_thread_index
```
alexcrichton: Could this be deferred to the C entrypoint to avoid needing to write details like `$root` in asm files? That way it should be possible to use the generated bindings to get the current thread index on the C side of things

libc-top-half/musl/src/exit/at_quick_exit.c
```diff
@@ -7,29 +7,26 @@
 
 static void (*funcs[COUNT])(void);
 static int count;
-#if defined(__wasilibc_unmodified_upstream) || defined(_REENTRANT)
-static volatile int lock[1];
-volatile int *const __at_quick_exit_lockptr = lock;
-#endif
+DECLARE_STRONG_LOCK(lock, static);
```
alexcrichton, line 10: is it intentional this is "strong" bu locked with "weak" below?

libc-top-half/musl/src/thread/common/tss_create copy.c
alexcrichton, line 1: Stray file name? (unusual to see a space)

libc-top-half/musl/src/thread/common/tss_delete copy.c
alexcrichton, line 1: stray file?

libc-top-half/musl/src/thread/common/tss_set copy.c
alexcrichton, line 1: stray file?

libc-bottom-half/sources/__libcall_thread_context.S
alexcrichton, line 1: This I think is ok for an initial implementation, but I think we're going to want to eventually ensure that calls to `__wasm_get_stack_pointer` go directly to the builtin intrinsic without needing a wrapper in the middle. Otherwise there's basically no way to create a module that directly performs the call because even with LTO these won't get inlined.

libc-top-half/musl/src/thread/coop-threads/mtx_trylock.c
```diff
@@ -0,0 +1,21 @@
...
+	if (m->_m_type == PTHREAD_MUTEX_NORMAL) {
+		if (m->_m_lock == 0) {
+			m->_m_lock = EBUSY;
+			return thrd_success;
+		} 
+		else {
+			return thrd_busy;
+		}
+	}
+	int ret = __pthread_mutex_trylock((pthread_mutex_t *)m);
```
alexcrichton: Would it make sense to have some sort of internal macro/header to do this cast to avoid needing to open-code it having the same representation?

libc-bottom-half/sources/__wasm_init_task.s
```diff
@@ -0,0 +1,96 @@
...
+    # to reference this constant without linking failing for shared libraries,
+    # i.e. a position-independent data reference. 
+    i32.const 131072 
+    call malloc
+    call __wasm_set_stack_pointer
+
+    # Allocate a new TLS area
+    call __allocate_aligned_tls
+    call __copy_tls
+    call __wasm_set_tls_base
```
alexcrichton: Would it be possible to delegate all of this to C instead of assembly? Or basically delegate as much of this function as possible to C instead of assembly here. I suspect that the initial instructions are required to be in assembly, but these look compatible (or at least a subset) to move over there

alexcrichton (commented):
An initial pass over things.

One major piece remaining in terms of synchronization is locks around the `descriptor_table.h` and derivative subsystems. None of that is locking-aware right now and would probably cause lots of problems if two threads tried to print to stdout for example. That's fine for an initial PR but we'll want to avoid broadcasting that threads are generally available until those issues are sorted out

alexcrichton:
Another high level thing I'm just now remembering -- we'll need to afford a transition period from today-without-coop-threads to tomorrow-with-coop-threads-stable. Basically I think we'll want some sort of CMake flag for "ok turn on threads" which is default off for wasip3. That way the default build of wasi-sdk, for example, will have a libc.a which has pthread symbols but the symbols don't actually do anything (return errors as they do today). Once the `component-model-threading` feature is finished/stable in Wasmtime and we've proven out the ecosystem then we can remove the option and unconditionally turn it on.

I'd like to ideally still only have one target, `wasm32-wasip3`, and from-source builds would have one extra option of doing pthreads-or-not. That keeps the target triples down and still easy to think about. Source-wise I think that'd involve adjusting some of the `#ifdef` here to use `__wasm_libcall_thread_context__` plus some CMake-set variable and that's what uses all the threading intrinsics here. With the CMake-set thing disabled it'd basically be the p3-target as-is now, plus some extra stuff w.r.t. the stack pointer and TLS.

Does that sound ok to you @TartanLlama? If so I was thinking [`wasi/version.h.in`](https://github.com/WebAssembly/wasi-libc/blob/main/libc-bottom-half/headers/public/wasi/version.h.in) is probably where we could invent some new `#define` name for "ok p3 actually has threads turned on"

TartanLlama:
Yeah, I think the additional CMake option makes sense

alexcrichton:
Ok I've extracted out a number of issues as follow-up from this PR:

* https://github.com/WebAssembly/wasi-libc/issues/804
* https://github.com/WebAssembly/wasi-libc/issues/805
* https://github.com/WebAssembly/wasi-libc/issues/806
* https://github.com/WebAssembly/wasi-libc/issues/807
* https://github.com/WebAssembly/wasi-libc/issues/808
* https://github.com/WebAssembly/wasi-libc/issues/809
* https://github.com/WebAssembly/wasi-libc/issues/810
* https://github.com/WebAssembly/wasi-libc/issues/811
* https://github.com/WebAssembly/wasi-libc/issues/819

Otherwise this is now [green in wasi-sdk](https://github.com/WebAssembly/wasi-sdk/pull/636) so I'm going to merge this PR here. More work is going to be required for coop threads fully working, and even wasip3 without coop threads isn't quite production ready just yet. That being said I think everything's in a reasonable enough state to at least start testing externally in the ecosystem.

alexcrichton:
Ok I've gone through a few iterations of llm-assisted review now which discovered a number of bugs. Everything should be fixed/tested now. Minor things still remain, but within the realm of "not really that consequential". I'll note that I've filed a few more issues as I went along (the list above expanded) and @TartanLlama you should do a once-over of my commits when you get back. Finally, I also disabled a few more open-posix-testsuite tests due to them exercising UB. I'm a bit suspicious that a nontrivial number of tests in this open test suite seem to do UB (e.g. join a detached thread, detach a joined thread, etc), but alas. Better than nothing!

In the meantime I'm still going to merge this to unblock a wasi-sdk release which will then, circularly, unblock this repository updating to test all of this in CI.

